PR #2, opened 2019-02-08
Bucketmanagement

Kidswiss:


service/restore/restoreRunner.go
```diff
@@ -6,28 +6,32 @@ import (
...
 	service.CommonObjects
-	config config
+	config   config
+	observer *observe.Observer
 }
 
 // NewRestoreRunner returns a new restore runner
-func NewRestoreRunner(restore *backupv1alpha1.Restore, common service.CommonObjects) *RestoreRunner {
-	return &RestoreRunner{
+func NewRestoreRunner(restore *backupv1alpha1.Restore, common service.CommonObjects, observer *observe.Observer) *restoreRunner {
```
mhutter: Why unexport `restoreRunner` when it is returned by exported `NewRestoreRunner`?

Kidswiss: Good catch, `NewRestoreRunner` needs to be unexported of course.

PR #6, opened 2019-03-19
Test travis PR build

ccremer:
just for testing
* it should build the docker image, but not push it
* should building after merging it to master, but not push


ccremer:
test was successful, built, but not pushed

PR #10, opened 2019-04-26
Spin up pods for backup commands

Kidswiss:
This functionality enables to define prebackuppods per backup, that get
spinned up prior to the wrestic run and ended after wrestic has
finished.


service/backup/backupRunner.go
```diff
@@ -189,3 +196,56 @@ func (b *backupRunner) updateBackupStatus() {
...
+
+func (b *backupRunner) stopPodTemplates() {
+	for _, set := range b.runningSets {
+		b.Logger.Infof("removing command pod %v/%v", set.GetNamespace(), set.GetName())
+		option := metav1.DeletePropagationForeground
+		err := b.K8sCli.Extensions().ReplicaSets(b.backup.GetNamespace()).Delete(set.GetName(), &metav1.DeleteOptions{
+			PropagationPolicy: &option,
+		})
+		if err != nil {
+			b.Logger.Errorf("could not remove replicaSet, maybe already deleted? %v", err)
```
srueg: If it's already deleted it would be an error of type `IsNotFound`:
```
if errors.IsNotFound(err) {
...
}
```

service/backup/utils.go
```diff
@@ -116,6 +131,53 @@ func newServiceAccountDefinition(backup *backupv1alpha1.Backup, config config) s
...
+	} else {
+		for _, template := range templates.Items {
+
+			template.Spec.Pod.PodTemplateSpec.ObjectMeta.Annotations = map[string]string{
+				b.config.backupCommandAnnotation: template.Spec.BackupCommand,
+				b.config.fileExtensionAnnotation: template.Spec.FileExtension,
+			}
+
+			podLabels := map[string]string{
+				"backupCommandPod": "true",
```
srueg: There should be a label which uniquely defines a pod, if multiple PreBackupPods are defined.

service/backup/backupRunner.go
```diff
@@ -189,3 +196,56 @@ func (b *backupRunner) updateBackupStatus() {
...
+
+// startPodTemplates will start and wait all pods defined in the template and
+// wait for at least one replication to be available for each pod.
+func (b *backupRunner) startPodTemplates() {
+	b.runningSets = b.getReplicaSets()
+	for _, set := range b.runningSets {
+		b.Logger.Infof("Creating command pod %v/%v\n", b.backup.GetNamespace(), set.GetName())
+		runningSet, err := b.K8sCli.Extensions().ReplicaSets(b.backup.GetNamespace()).Create(&set)
+		if err != nil {
+			b.Logger.Errorf("error creating command pod %v: %v\n", set.GetName(), err)
```
srueg: I'm not really happy with the whole error handling. Can we at least expose failures as Prom metrics? Otherwise they will just silently fail (as no one looks at the logs).

manifest/examples/backup/prebackuppod.yaml
```diff
@@ -0,0 +1,31 @@
...
+          ports:
+            - containerPort: 3306
+              protocol: TCP
+          volumeMounts:
+            - mountPath: /var/lib/mysql
+              name: mariadb
+      volumes:
+        - name: mariadb
+          persistentVolumeClaim:
+            claimName: mariadb
```
srueg: Can't we just run `sleep 1000` or similar in the PreBackupPod? Then we don't have to set any env vars or use a PVC...

PR #60, opened 2019-07-09
Monitoring

Kidswiss:


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Simplyfied merging logic for backend configs using mergo
```
corvus-ch: The changes referred by this line, does not seem to be related to monitoring. Unless I missed something, this is a scope creep and have to go into a PR of its own.

monitoring/httpendpoint.go
```diff
@@ -68,6 +70,7 @@ func new() *MetricExporter {
 			panic(err)
 		}
 	}()
+	tmp.Metrics = metrics.NewPrometheus(prometheus.DefaultRegisterer)
```
corvus-ch: Bad variable naming `tmp`. Go best practices suggest to be more expressive. The variable holds a metric exporter object. Suggested variables names would be `metricExporter`, `exporter`, `me` (no, not `e` as that is reserved for errors).

service/observe/subscription.go
```diff
@@ -40,6 +65,12 @@ type Subscriber struct {
 	TopicName string // contains the name that this subscriber is registered with
 }
 
+type promMetrics struct {
```
corvus-ch: This does not provide any value and the code can work without. Please remove and use package level variables without any wrapper.

service/observe/subscription.go
```diff
@@ -155,6 +187,10 @@ func (s *Subscriber) WatchLoop(watch WatchObjects) {
 
 	defer GetInstance().GetBroker().Unsubscribe(s.TopicName, s)
 
+	// initialise the metrics, as there's no rate change if it just pops up with
+	// the value of 1.
+	s.initMetrics(watch.Job.GetNamespace(), string(watch.JobName))
```
corvus-ch: This gets executed over and over again but needs to happen only once after the app starts. This better be done within the Subscribers constructor function or a package init function.

service/observe/subscription.go
```diff
@@ -180,3 +218,18 @@ func (s *Subscriber) WatchLoop(watch WatchObjects) {
 		}
 	}
 }
+
+func (s *Subscriber) incrFailure(namespace, jobType string) {
```
corvus-ch: This function is self contained. Make it a package level function instead of a receiver to Subscriber. Same goes for the `incSuccess` and `initMetrics`.

service/schedule/scheduleRunner.go
```diff
@@ -270,8 +264,20 @@ func ScheduledLabelFilter() string {
...
 
-func (s *scheduleRunner) waitForLock(name string, backend string, jobs []observe.JobName) {
+func (s *scheduleRunner) waitForLock(name string, backend string, jobs []observe.JobName, jobType string) {
+	queue := queueCount.WithLabelValues(s.schedule.GetNamespace(), jobType)
+
 	s.Logger.Infof("%v for repo %v is queued waiting for jobs %v to finish", name, backend, jobs)
+	queue.Inc()
 	s.observer.GetLocker().WaitForRun(backend, jobs)
 	s.Logger.Infof("All blocking jobs on %v for %v are now finished", backend, name)
+	queue.Dec()
```
corvus-ch: This is a good place to make use of `defer`.

manifest/examples/prometheus/rules.yml
```diff
@@ -0,0 +1,19 @@
...
+      summary: "Job in {{ $labels.namespace }} of type {{ $labels.jobType }} failed"
+- name: no jobs finished within 24 hours
+  rules:
+  - alert: NoBackupRunning
+    expr: rate(k8up_jobs_total[1d]) == 0 and on(namespace) k8up_schedules_gauge > 0
+    for: 25h
+    labels:
+      severity: critical
+    annotations:
+      summary: "Jobs in {{ $labels.namespace }} not running, there might be an operator deadlock"
```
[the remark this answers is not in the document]
Kidswiss: Changed it to: "No K8up jobs were run in {{ $labels.namespace }} within the last 24 hours. Check the operator, there might be a deadlock" as the check looks if any jobs at all run during the last 24h.

service/schedule/scheduleRunner.go
Kidswiss: Nice catch! That should lower the risk of deadlocks even further!

CHANGELOG.md
Kidswiss: Point taken. This was an effort to make the abomination called `scheduleRunner.Start()` in the schedule package a bit more manageable. I since created a separate ticket to clean that monster a bit. I'd like to keep this in here as I'm afraid to break stuff (which is very much a red flag that shows how bad that functions is currently.)

manifest/examples/prometheus/prometheus.yml
```diff
@@ -0,0 +1,31 @@
...
+  static_configs:
+  - targets:
+    - localhost:9090
+- job_name: operator
+  honor_timestamps: true
+  scrape_interval: 15s
+  scrape_timeout: 10s
+  metrics_path: /metrics
+  scheme: http
+  static_configs:
```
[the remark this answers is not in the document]
Kidswiss: Changed the existing `job_name` to `operator-debug` and added a new one with service discovery named `operator-kubernetes` as defined on some prod clusters.

PR #67, opened 2019-09-13
Implement the multistage Docker build

anothertobi:


Dockerfile
```diff
@@ -1,9 +1,19 @@
...
 
 RUN apk add --no-cache tzdata
 
 WORKDIR /go/src/github.com/vshn/k8up
 COPY . .
 RUN go install -v ./...
 
-ENTRYPOINT [ "operator" ]
+# runtime image
+FROM docker.io/alpine:3
```
tobru: Why not `FROM scratch`?

Dockerfile
```diff
@@ -1,9 +1,19 @@
...
+WORKDIR /app
+
+RUN apk --no-cache add ca-certificates tzdata
```
tobru: When using `scratch` these files could simply be copied from the build container.

Dockerfile
```diff
@@ -1,9 +1,19 @@
-FROM golang:1.11-alpine
+FROM golang:1.11-alpine as build
```
tobru: Maybe try Go 1.13?

anothertobi: Upgrading the Go version should be done in another PR as the current one should just move the Docker build to a multi stage build.

Dockerfile
anothertobi: The 6MB from the alpine image won't hurt and the overhead of copying things around and hacking them into scratch is not worth it IMO.

tobru: It somehow "hurts" because it adds unnecessary additional stuff which could increase the attack surface and adds a "base" image maintenance burden. I'm in favor of using `FROM scratch` whenever possible.

anothertobi: Will change it to scratch 👍  

@Kidswiss do you know where tzdata should be located (as expected by k8up)?

anothertobi: @Kidswiss and I looked at this together and it is too complicated to get it into scratch.

Issue #71, opened 2019-12-12
Explain promURL


dns2utf8:
Dear vshn team

In some part of the docs it says it points at a Prometheus endpoint.
However, in the examples it is the same URL as the minio server.
So my question, what does it do now?

Cheers

dns2utf8:
Fixed by my PR #72

PR #72, opened 2019-12-12
Fix promURL

dns2utf8:
Dear vshn team

I hope I understand the example correctly.

Cheers,
Stefan

Addresses #71


srueg:
Good catch! Thanks for the contribution :+1:

PR #83, opened 2020-01-07
Documentation: How to pass environment variables to backup command

akosma:


docs/modules/ROOT/pages/advanced-config.adoc
```diff
@@ -71,6 +71,14 @@ spec:
 
 ----
 
+[TIP]
+.Passing environment variables
+====
+If you want to pass environment variables to the `backupCommand` you'll have to wrap them in a shell. Example:
+
+`appuio.ch/backupcommand=/bin/bash -c 'mysqldump -uroot -p\"\${MARIADB_ROOT_PASSWORD}\" --all-databases'" --overwrite`
```
dns2utf8:
```suggestion
`kubectl -n wordpress-demo annotate pods ${YOUR_POD_NAME} "appuio.ch/backupcommand=/bin/bash -c 'mysqldump -uroot -p\"\${MARIADB_ROOT_PASSWORD}\" --all-databases'" --overwrite`
```
without the kubectl it would be very confusing and have an open `"` and an unexplained `--overwrite`

Kidswiss:
LGTM

dns2utf8:
I would either remove the `" --overwrite` or add the `kubectl ...` to explain the annotation.

docs/modules/ROOT/pages/advanced-config.adoc
dns2utf8:
```suggestion
`"appuio.ch/backupcommand=/bin/bash -c 'mysqldump -uroot -p\"\${MARIADB_ROOT_PASSWORD}\" --all-databases'"`
```
As an alternative document this.

akosma:
I will add @dns2utf8 comments directly in master, sorry we clicked merge too fast!! :)

Kidswiss:
Okay let's remove `" --overwrite` so we don't have to do the escaping stuff:
```
appuio.ch/backupcommand=/bin/bash -c 'mysqldump -uroot -p "${MARIADB_ROOT_PASSWORD}" --all-databases'
```

dns2utf8:
@Kidswiss all the escaping is needed so the shell you are executing the command in does not evaluate the variables. In the end the goal is to let the backup job expand the variables with the credentials inside the DB container. So one can use the same command on many DB pods without having the clear text password in the annotation.

Kidswiss:
@dns2utf8 That's true, but if you have the annotation in the yaml file you don't have to do the escaping as you're not running `kubectl annotate...`.

Like:
```
      annotations:
        appuio.ch/backupcommand: /bin/bash -c 'mysqldump -uroot -p "${MARIADB_ROOT_PASSWORD}" --all-databases'
```

@akosma I feel like we should document both ways to add annotations properly as this is a source for confusion as we're seeing here :)

dns2utf8:
That would make things very clear :+1:

PR #87, opened 2020-01-14
Relaxing the image pull policy

dns2utf8:
One of our k8s clusters overloaded when too many backup jobs pulled `docker.io/vshn/wrestic:v0.0.10` for every job.

I was not able to test the modified operator yet. However, it compiles and I (coming from rust) believe this is a good sign.

Cheers,
Stefan

### DIFF
--- a/service/common.go
+++ b/service/common.go
@@ -89,6 +89,12 @@ func GetBasicJob(kind string, config config.Global, object metav1.Object) *batch
 		config.Label:      "true",
 		config.Identifier: PseudoUUID(),
 	}
+	
+	pullPolicy := corev1.PullIfNotPresent
+	if strings.HasSuffix(config.Image, ":latest") {
+		pullPolicy = corev1.PullAlways
+	}
+	
 	return &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      nameJob,
@@ -118,7 +124,7 @@ func GetBasicJob(kind string, config config.Global, object metav1.Object) *batch
 									Value: object.GetNamespace(),
 								},
 							},
-							ImagePullPolicy: corev1.PullAlways,
+							ImagePullPolicy: pullPolicy,
 							TTY:             true,
 							Stdin:           true,
 						},


Kidswiss:
Hi @dns2utf8 

Thanks for the request! I see the issue you may have if you're backing up a lot of namespaces at once.

The new config makes sense, but how about making it configurable? Sometimes I had to debug stuff on a customer cluster and used `vxxx-debug` tags, then it's very helpful to AlwaysPull.

Regards
Simon

dns2utf8:
Hi Simon

I see the function already has access to `config.Global`. Do you mean there should be an additional field there?
Eg. `BackupJobImagePullPolicy`?
Where would one configure said config?

Best,
Stefan

Kidswiss:
If we're following the pattern for other such configs I'd say:

* a global config via environment variable
* ability to overwrite it in the various CRDs (backup, schedule, etc.)

That would give the most flexibility, as for debugging the flag can be set on single backups or schedules.

But there should be a better way to handle the merging and precedences of the configs as it's currently just a if/else cascade...

dns2utf8:
My current client has tasks on a higher priority currently.
So I can not spend much more time on this (unless somebody else hires me).

What I like about the current implementation is, it behaves like k8s does when omitting the pull policy.
I could add a new member to the global config object. However, I don't know how to add the CRD overwrites.

Cheers,
Stefan

Kidswiss:
I'm currently in a similar situation, we're working on https://github.com/projectsyn/ currently and has a higher prio than k8up (although k8up will be part of the whole thing), as it is currently rough around the edges but works.

I can carve out some time for fixing really critical bugs but new features aren't a prio right now...

I'm keeping this open until I get some more time for k8up again.

ccremer:
Hi @dns2utf8 
With #114 we have rewritten the code quite a bit. As a consequence, your PR is outdated. And also not necessary anymore.

With the latest code (currently unreleased though), we do not further specify the image pull policy. That results in the behaviour described in https://kubernetes.io/docs/concepts/configuration/overview/#container-images. To quote:
> imagePullPolicy is omitted and either the image tag is :latest or it is omitted: Always is applied.
> imagePullPolicy is omitted and the image tag is present but not :latest: IfNotPresent is applied.

The same is true for OpenShift 3.11: https://docs.openshift.com/container-platform/3.11/dev_guide/managing_images.html#image-pull-policy

I think at this point you could either await the next K8up release, or try it out on your own risk with our `master` image tag. `latest` does not point to a latest master branch anymore, but to the latest released git tag.

In the future though, we plan to rework the CRDs (new API version) to address various issues that we have now experienced so far.

I hope it's ok to close this PR

PR #91, opened 2020-02-06
Rename annotations

corvus-ch:
This change renames the annotations. The will now use `k8up.syn.tools/…` as their prefix instead of `appuio.ch/…`.


corvus-ch:
@Kidswiss The documentation for `FileExtensionAnnotation` is missing. This is the perfect opportunity to add it. As I do not have enough understanding, I leave that to you.

Kidswiss:
I added a note to the CHANGELOG that makes the impact of the change clear. Also added the file-extension annotation to the docs.

Rest LGTM.

PR #94, opened 2020-04-02
Add the ability to define tags for backups and restores

Kidswiss:
With this commit it is possible to pass a list of tags to the backups
and restores. Thus leverating Restic's internal tag management.


service/common.go
```diff
@@ -157,3 +157,13 @@ func DefaultEnvs(backend *backupv1alpha1.Backend, config config.Global) []corev1
 	}...)
 	return vars
 }
+
+// BuildTagArgs will prepend "--tag " to every element in the given []string
+// TODO: this is wrong, needs fixing!!
```
srueg, line 162: What's wrong about it?

Kidswiss: That can be removed, I fixed the underlying issue, but forgot to remove the `TODO:`

PR #99, opened 2020-05-06
Split up tutorial in docs into separate chapters

bittner:
@Kidswiss and I made a significant overhaul of the K8up documentation.

- Splits up tutorial in docs into separate chapters
- Adds documentation on `Archive` object

These changes should it make easier to remove related usage instructions from cloud services, such as [APPUiO](https://appuio.ch) ([related docs](https://github.com/appuio/docs/pull/74)).

PR #100, opened 2020-05-06
Rename baas -> k8up in manifests

bittner:
Replaces the mentions of `baas` by `k8up` in the manifest files.

**Relates to:** #99

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ oc login -u system:admin # default developer account doesn't have the rights to 
 #The operator has the be run at least once before to create the CRD
 go run cmd/operator/*.go -development
 #Add a demo backupworker (adjust the variables to your liking first)
-kubectl apply -f manifest-examples/baas.yaml
+kubectl apply -f manifest-examples/k8up.yaml
 #Add a demo PVC if necessary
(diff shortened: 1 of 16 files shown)


manifest/examples/prometheus/rules.yml
```diff
@@ -1,8 +1,8 @@
 groups:
 - name: K8up
   rules:
-  - alert: baas_last_errors
-    expr: baas_backup_restic_last_errors > 0
+  - alert: k8up_last_errors
```
Kidswiss, line 4: This may break the example configuration, as the prometheus metrics are still using baas.

bittner: Does this relate to the entire change set or only to these two lines?

What would you suggest? Should we close the entire PR, or only undo this specific change?

Kidswiss: It's only those two lines.

bittner: I undid the changes on those two lines.

bittner:
@Kidswiss Shall we merge these changes now? Can you take another look?

PR #102, opened 2020-05-27
Adding mutex to the subscriber

Kidswiss:
There was a race condition, which could case the subscriber channel to
be written even though it was already closed.


Kidswiss:
Yes, also had to add the mutex to the unit tests. They should now work again.

Kidswiss:
It's now working, forgot to save the last changes of the file before committing :D

PR #112, opened 2020-09-24
Explicitly define ApiVersions for K8s APIs

mhutter:
Unversioned APIs are deprecated and will be removed.

### DIFF
--- a/service/archive/archiveRunner.go
+++ b/service/archive/archiveRunner.go
@@ -40,7 +40,7 @@ func (a *archiveRunner) Start() error {
 
 	go a.watchState(archiveJob)
 
-	_, err := a.K8sCli.Batch().Jobs(a.archiver.Namespace).Create(archiveJob)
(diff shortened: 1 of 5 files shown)


vendor/github.com/modern-go/concurrent/log.go
```diff
@@ -10,4 +10,4 @@ import (
 var ErrorLogger = log.New(os.Stderr, "", 0)
 
 // InfoLogger is used to print informational message, default to off
-var InfoLogger = log.New(ioutil.Discard, "", 0)
```
Kidswiss, line 13: Did you run `go fmt` on the whole project or something?

Kidswiss (commented):
Should I also check for some leftovers?

mhutter:
> Should I also check for some leftovers?

I think I got them all, courtesy of `golangci-lint`

ccremer:
I think with the rewrite in #114 , this PR is outdated and doesn't apply anymore.
Right? Can we close?

mhutter:
Yes, PR is obsolete.

PR #114, opened 2020-10-26
WIP: Rewrite

Kidswiss:
Rewriting a lot of the old code base and concepts to get rid of the POC code and technical dept.

Very WIP currently. This PR is to make the process more visible.

Some changes from the existing design and code:
* use operator-sdk
* get rid of POC code
* ~redo configuration (get rid of the myriad of envVars)~ will be done seperately, let's focus on a foundation
* simplyfy observation and locking

Resolves: #115


Kidswiss:
The rewrite for the architecture is complete.

Currently only the job type backup is implemented with all features and there are some other things missing as well.

I'll open tickets for each ting that still needs to be re-implemented. Although the remaining job types should be much easier as the backup is by far the most complex one.

queue/execution.go
```diff
@@ -0,0 +1,89 @@
...
+
+// GetRepositories returns a list of all repositories that are currently
+// handled by the queue.
+func (eq *ExecutionQueue) GetRepositories() []string {
+	eq.mutex.Lock()
+	defer eq.mutex.Unlock()
+	repositories := make([]string, len(execution.queues))
+	for repository := range execution.queues {
+		repositories = append(repositories, repository)
+	}
```
mweibel, line 87: That's a bug: https://play.golang.org/p/aX2TKO8LGTF

You either want to do:

```golang
	repositories := make([]string, 0, len(execution.queues))
	for repository := range execution.queues {
		repositories = append(repositories, repository)
	}
```

or:

```golang
	var repositories []string  // `repositories := make([]string, 0)` would be also possible but this declaration is slightly preferrable in case `len(execution.queues) == 0`
	for repository := range execution.queues {
		repositories = append(repositories, repository)
	}
```

Kidswiss: Good catch!

Went with method 1 and pushed it. Although I don't expect that there will be that many repositories to make a measurable difference.

Issue #118, opened 2020-10-27
Deduplicate prune jobs


Kidswiss:
If a prune job is already running for a given repository and another one is triggered (either manually or some schedule), it should get skipped.

This ensures that multiple prune jobs don't clog up the operator and starve out the time for actual backups.

The same applies to check jobs, too.

Kidswiss:
One idea here:

Add a `skipped` field to the `K8upstatus` in the CRD.

If the operator detect already a running prune set the skipped true. Check for that case in the execution worker.

Schnitzel:
In the case of Lagoon we have the case that multiple namespaces share the same s3 bucket:
- `drupal-example-main`
- `drupal-example-develop`
- `drupal-example-staging`

they all use the bucket `baas-drupal-example`
There is no specific `drupal-example` namespace. Therefore each of the namespaces create a backup schedule that includes a prune schedule, therefore having 3 prunes per week. We made sure with random generator that these prunes are not at the same time.
But still pruning 3 times the same bucket is not really useful.

Would this new system also be able to detect this? Like it realizes that there are 3 prunes for the same bucket and it would only run one? Maybe that's also connected to #116 where we would just for all 3 prunes say `@weekly` and then the operator runs only one prune at a good time per week?

Kidswiss:
As discussed yesterday in the internal call I'm closing this issue and merge it with #116.

Because we realised that these two features are very closely related and make sense to be implemented together.

tobru:
This has now moved into #214

PR #130, opened 2020-11-05
feat(prune): rewrite prune controller

### LINKED ISSUE #125: Add Prune jobs to the rewrite
Add the check job logic to the new implementation: https://github.com/vshn/k8up/tree/master/service/prune

mweibel:
fixes #125


mweibel:
@Kidswiss could you have a look at this and test it?

I tested it locally and it seems to work, however as I'm quite new in this project and there are no tests I'd like if you could take a look at it and test it. 

Testing locally with a 1 minute schedule makes #118 definitively necessary. I'll think about it.

Kidswiss (commented):
Skimming through the code it looks like what I had in mind. I'll test it ASAP.

Let's discuss your points from the chat on Monday.

executor/prune.go
```diff
@@ -0,0 +1,169 @@
...
+
+	jobObj, err := job.GetGenericJob(pruneObject, p.Config)
+	jobObj.GetLabels()[job.K8upExclusive] = "true"
+	if err != nil {
+		return err
+	}
+
+	// FIXME(mw): I just copied this over. I'm unsure if that's really necessary.
+	go func() {
+		p.startPrune(jobObj)
```
mweibel, line 53: having this in a separate goroutine: I'm not sure if this is needed and not sure if it's a good idea.

Possible issue: `p.prune = pruneObject` -- this could be a race condition. I copied parts from the backup executor but I'm not quite sure about why some things are done as they are done :)

Kidswiss: Yeah that part of the backup is a very special case because we have to wait for the pre-backup pods to be ready and don't want to slow down the queue execution.

It's not really needed here.

Kidswiss: As the go routine here isn't really require you can change `startPrune()` to return an error.

Kidswiss:
Tested it and it looks like the prune is triggered successfully.

PR #132, opened 2020-11-05
feat: rewrite check executor

### LINKED ISSUE #123: Add Check jobs to the rewrite
Add the check job logic to the new implementation: https://github.com/vshn/k8up/tree/master/service/check

With #114 merged the new implementation lives on the `development` branch.

susana-garcia:
fixes #123

### DIFF
--- a/executor/check.go
+++ b/executor/check.go
@@ -1,10 +1,18 @@
 package executor
 
-import "github.com/vshn/k8up/job"
+import (
+	stderrors "errors"
+
+	k8upv1alpha1 "github.com/vshn/k8up/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


susana-garcia:
@Kidswiss please, have a look and test it. I test it locally by applying the samples (check and schedule with the check spec) and I didn't see errors, but I'm not really sure if it's doing what it's suppose to do. If you have better ideas on how to test it more efficiently, please let me know. Thank you in advance.

Kidswiss:
At the first glance it looks about like what I had in mind. Testing it now.

Kidswiss:
The check is triggered successfully.

PR #135, opened 2020-11-09
feat: rewrite check executor

### LINKED ISSUE #123: Add Check jobs to the rewrite

susana-garcia:
Reopen PR:

fixes #123 

related: #132

### DIFF
--- a/executor/check.go
+++ b/executor/check.go
@@ -1,10 +1,18 @@
 package executor
 
-import "github.com/vshn/k8up/job"
+import (
+	stderrors "errors"
+
+	k8upv1alpha1 "github.com/vshn/k8up/api/v1alpha1"
+	"github.com/vshn/k8up/constants"
+	"github.com/vshn/k8up/job"
+	corev1 "k8s.io/api/core/v1"
+)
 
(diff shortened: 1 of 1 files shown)


Kidswiss:
The check loses the `RESTIC_PASSWORD` env var after `vars.Merge()`. Haven't digged in too deep to analyse why. But that part is basically identical to the other jobs...

@mweibel we want to change the whole env handling anyways. Do you think we should invest much time here to debug?

mweibel:
maybe we should add a (failiing) test similar to what I did in #140 and merge it then? making it pass when we rewrite the env stuff?

Kidswiss:
Sounds good to me. In that case I approve the PR.

Kidswiss (approved):

susana-garcia:
@Kidswiss I've read the comments in #140 and I assumed we'll merge first the open PRs, fix the envVar merging and add better tests later on as discussed initially (as the ones in #140 are not solving the problem of losing values).

PR #139, opened 2020-11-10
feat: rewrite archive

### LINKED ISSUE #124: Add Archive jobs to the rewrite
Add the check job logic to the new implementation: https://github.com/vshn/k8up/tree/master/service/archive

susana-garcia:
fixes #124

### DIFF
--- a/api/v1alpha1/backend.go
+++ b/api/v1alpha1/backend.go
@@ -125,6 +125,24 @@ func (s *S3Spec) EnvVars(vars map[string]*corev1.EnvVarSource) map[string]*corev
 	return vars
 }
 
+func (s *S3Spec) RestoreEnvVars() map[string]*corev1.EnvVarSource {
+	vars := make(map[string]*corev1.EnvVarSource)
+
+	if s.AccessKeyIDSecretRef != nil {
+		vars[constants.RestoreS3AccessKeyIDEnvName] = &corev1.EnvVarSource{
+			SecretKeyRef: s.AccessKeyIDSecretRef,
+		}
+	}
+
+	if s.SecretAccessKeySecretRef != nil {
+		vars[constants.RestoreS3SecretAccessKeyEnvName] = &corev1.EnvVarSource{
+			SecretKeyRef: s.SecretAccessKeySecretRef,
+		}
+	}
+
+	return vars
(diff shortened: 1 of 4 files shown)


susana-garcia:
@Kidswiss can you have a look at this PR when you a minute, please? I'm not entirely happy with it, so I'd really appreciate your feedback

Kidswiss:
Haven't tested it yet, but looks more or less like the other job implementations. I think the parts you're not happy about are the same that Michael already mentioned, correct?

susana-garcia:
@Kidswiss not really, the parts I'm not that happy with are more the parts I added from the other branch that I'm not sure if that was needed. For example: `BuildTagArgs` or the `setupEnvVars`

Kidswiss:
I actually assume that the tag handling got lost while I implemented the `backup` case. So it should be there actually. I'd even say it should be already handled in the `GetGenericJob()` method as all the jobs can accept a list of tags.

As for the envVars: It looks good, but looking through the code I found an issue.  Archive (and restore for that matter) may need two sets of S3 credentials:
1. for the backup location
2. for the restore location

You only call `RestoreMethod.S3.EnvVars()` so that will only create the one set of the S3 credentials. But the bigger issue I found is that `RestoreMethod.S3.EnvVars()` will produced the same set of envVars as `Backend.S3.EnvVars()` would. Which would overwrite the other set. `RestoreMethod.S3.EnvVars()` should create envVars named like:

```
RESTORE_S3ENDPOINT
RESTORE_ACCESSKEYID
RESTORE_SECRETACCESSKEY
RESTORE_DIR # this one would be used for the restore/archive restoreMethod folder
```

See https://github.com/vshn/wrestic/blob/master/restic/restic.go#L37
and for the restore method: https://github.com/vshn/k8up/blob/master/apis/backup/v1alpha1/restore.go#L36

So it needs to check if the restore method is folder or s3. If it's S3 set above envVars.

I hope it's somewhat clear what I'm trying to say :) if not we can do a zoom call.

Kidswiss (approved):
Looks good, the ugly things will be looked at once we have a complete picture of the jobs and their ugliness.

Tested it and the produced jobs looks fine.

PR #140, opened 2020-11-10
feat(restore): rewrite restore

### LINKED ISSUE #126: Add Restore jobs to the rewrite
Add the check job logic to the new implementation: https://github.com/vshn/k8up/tree/master/service/restore

mweibel:
fixes #126

### DIFF
--- a/api/v1alpha1/restore_types.go
+++ b/api/v1alpha1/restore_types.go
@@ -35,6 +35,16 @@ type RestoreSpec struct {
 	Tags []string `json:"tags,omitempty"`
 }
 
+func (r RestoreSpec) CreateObject(name, namespace string) runtime.Object {
+	return &Restore{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Spec: r,
+	}
+}
+
 // RestoreMethod contains how and where the restore should happen
 // all the settings are mutual exclusive.
 type RestoreMethod struct {
@@ -63,6 +73,22 @@ type Restore struct {
 	Status RestoreStatus `json:"status,omitempty"`
 }
 
+func (r *Restore) GetRuntimeObject() runtime.Object {
+	return r
+}
+
+func (r *Restore) GetMetaObject() metav1.Object {
+	return r
+}
+
+func (r *Restore) GetType() string {
+	return "restore"
+}
+
+func (r *Restore) GetK8upStatus() *K8upStatus {
+	return &r.Status.K8upStatus
+}
+
 // +kubebuilder:object:root=true
 
 // RestoreList contains a list of Restore
@@ -74,14 +100,4 @@ type RestoreList struct {
 
 func init() {
 	SchemeBuilder.Register(&Restore{}, &RestoreList{})
-}
-
-func (r RestoreSpec) CreateObject(name, namespace string) runtime.Object {
-	return &Restore{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      name,
-			Namespace: namespace,
-		},
-		Spec: r,
-	}
 }
--- a/api/v1alpha1/backend.go
+++ b/api/v1alpha1/backend.go
@@ -125,19 +125,43 @@ func (s *S3Spec) EnvVars(vars map[string]*corev1.EnvVarSource) map[string]*corev
 	return vars
 }
 
-func (s *S3Spec) RestoreEnvVars() map[string]*corev1.EnvVarSource {
-	vars := make(map[string]*corev1.EnvVarSource)
-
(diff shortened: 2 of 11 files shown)


api/v1alpha1/restore_types.go
```diff
@@ -26,7 +26,9 @@ import (
 // one you need to define an existing PVC.
 type RestoreSpec struct {
 	// Backend contains the backend information
-	Backend       *Backend       `json:"backend,omitempty"`
+	Backend *Backend `json:"backend,omitempty"`
+	// FIXME(mw): does not work
+	// +kubebuilder:validation:required
```
ccremer, line 31: In https://book.kubebuilder.io/reference/markers/crd-validation.html it is uppercased, try this?
```suggestion
    // +kubebuilder:validation:Required
```
or `// +kubebuilder:validation:Required=true`

mweibel: doesn't work unfortunately. The uppercase version I've tried already and providing `=true` yields an error:
```
» make manifests
/Users/michael/code/go-workspace/bin/controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/Users/michael/code/vshn/k8up/api/v1alpha1/restore_types.go:31:2: extra arguments provided: "true" (at <input>:1:1)
/Users/michael/code/vshn/k8up/api/v1alpha1/restore_types.go:31:2: extra arguments provided: "true" (at <input>:1:1)
/Users/michael/code/vshn/k8up/api/v1alpha1/restore_types.go:31:2: extra arguments provided: "true" (at <input>:1:1)
/Users/michael/code/vshn/k8up/api/v1alpha1/restore_types.go:31:2: extra arguments provided: "true" (at <input>:1:1)
```

ccremer: However, I was just quickly looking into the code and looked into the structs. So far I cannot see a reason why they have to be pointers at all. If it were
```go
type RestoreSpec struct {
	// Backend contains the backend information
	Backend       Backend       `json:"backend,omitempty"`...
}
```
then the Go deserializer will just fill the Backend field with an empty Backend struct. Now I don't want to question the API, but it would be nice if there's almost no pointers.

Kidswiss: The backend shouldn't be required for every CRD there can be defaults provided. That's also the main reasoning behind using pointers everywhere. We can explicitly check if a value was set or not. But on the other hand we HAVE to check if a value was set or not.

Examples in https://book.kubebuilder.io/cronjob-tutorial/api-design.html#designing-an-api use pointers for almost everything due to this.

mweibel: yeah pointer vs non-pointer .. classic Go discussion ;) The optional part is quite significant though, as for empty structs it's rather ugly to check whether they've been set or not. 

@Kidswiss what's your opinion on the validation part? i.e. marking everything optional and specifics required (or vice-versa)?

mweibel:
@Kidswiss the implementation is more or less complete I believe. I have a known issue with the folder restore method (job doesn't get setup properly, need to debut this). But if you have some time, feel free to do a review already. 

Code is similar to others, I'm not very happy - I think we should start merging the other PRs in order to start refactoring, unifying and simplifying. 

I added a unit test for the job spec which tests some cases. The `globals` can't be tested well because the assignment is already on loading the Go code (i.e. I'd need to set env vars on `init()` probably, which is not really that great for testing individual cases).

api/v1alpha1/restore_types.go
ccremer: The site you linked is just using int32 pointers for API compatibility reasons, but not the structs (e.g. `JobTemplate batchv1beta1.JobTemplateSpec 'json:"jobTemplate"'))`. I don't see any mentions regarding struct pointers...

But I don't want to involve myself too much here atm, I was just helping out Michael for the generation problem :) If you have reasons to do it that way, I'm perfectly fine with it :)

Kidswiss: @mweibel sounds reasonable, that should help with the endless nil checking for certain things.

@ccremer the pointers for the structs were used for the same reason as for the basic types: checking if it was explicitly set or not. Example is the logic to determine what backend is actually used.

Kidswiss:
I'm currently going through the open PRs.

I'm more or less just checking if the generated jobs look how they are supposed to look. The ugly things we discussed can be looked at once they are merged and we have a complete picture.

Kidswiss:
I'm observing the same issue here as in #135 where the merging drops some values...

As such I can't do the functional test of this job as it's invalid. But I'd suggest we merge, fix the envVar merging and then test through all the jobs with references.

tobru:
@mweibel This needs a conflict cleanup.

PR #144, opened 2020-11-17
fix(archive): s3 restore env vars

mweibel:
pr #140 refactored the RestoreEnvVars() method slightly and due to fast merging the issue wasnt discovered


ccremer (approved):
Though I can't judge if it makes sense

tobru (approved):

PR #146, opened 2020-11-17
Re-add prometheus metrics after Operator SDK migration

### LINKED ISSUE #127: Add prometheus metrics to the rewrite
The rewrite currently doesn't contain any k8up metrics.

The operator-sdk provides a default metrics endpoint though. So use that one to add the custom metrics, if possible.

Most metrics are defined in the observer or the scheduling in the old version:
https://github.com/vshn/k8up/blob/master/service/schedule/scheduleRunner.go#L30
https://github.com/vshn/k8up/blob/master/service/observe/subscription.go#L23

susana-garcia:
Fixes #127


mweibel:
I think you need to rebase this one :)

main.go
```diff
@@ -63,7 +62,7 @@ func main() {
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:             scheme,
-		MetricsBindAddress: metricsAddr,
+		MetricsBindAddress: constants.GetMetricBind(),
```
mweibel, line 65: what's the reason for this change?

susana-garcia:
I'm not sure about the last commit with the metrics `metricsFailureCounter, metricsSuccessCounter and metricsTotalCounter`. I wasn't sure about the job type as I think we don't have this info available in that bit of code (or that's what I think). All feedback is welcome!

queue/execution.go
```diff
@@ -25,6 +38,10 @@ type Executor interface {
 	// Logger returns the logger in the job's context so we can
 	// Associate the logs with the actual job.
 	Logger() logr.Logger
+	// GetType() string
+	GetType() string
+	// GetNamespace() string
+	GetNamespace() string
 	// GetName() string
```
ccremer, line 45: Please check these comments, they don't add value like this

queue/execution.go
```diff
@@ -87,3 +112,11 @@ func (eq *ExecutionQueue) GetRepositories() []string {
...
 	return repositories
 }
+
+func (eq *ExecutionQueue) incQueueCount(namespace, jobType string) {
+	queueCount.WithLabelValues(namespace, jobType).Inc()
+}
+
+func (eq *ExecutionQueue) decQueueCount(namespace, jobType string) {
+	queueCount.WithLabelValues(namespace, jobType).Dec()
+}
```
ccremer, line 122: Prometheus makes clear distinction between counters (not decrementable) and gauges. Let's not mix them in the naming :)
```suggestion
func (eq *ExecutionQueue) incQueueGauge(namespace, jobType string) {
	queueGauge.WithLabelValues(namespace, jobType).Inc()
}

func (eq *ExecutionQueue) decQueueGauge(namespace, jobType string) {
	queueGauge.WithLabelValues(namespace, jobType).Dec()
}
```

scheduler/scheduler.go
```diff
@@ -25,6 +28,13 @@ const (
 
 var (
 	scheduler *Scheduler
+
+	scheduleCount = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "k8up_schedules_gauge",
+		Help: "How many schedules this k8up manages",
+	}, []string{
+		"namespace",
+	})
```
ccremer, line 37: Same here, Gauges aren't Counters

constants/constants.go
```diff
@@ -39,6 +39,7 @@ var (
 	globalS3Endpoint               = getEnvVar("BACKUP_GLOBALS3ENDPOINT", "")
 	globalSecretAccessKey          = getEnvVar("BACKUP_GLOBALSECRETACCESSKEY", "")
 	globalStatsURL                 = getEnvVar("BACKUP_GLOBALSTATSURL", "")
+	metricBind                     = getEnvVar("BACKUP_METRICBIND", ":8080")
```
ccremer, line 42: `bind` and `address` or `addr` go together
```suggestion
	metricBindAddr                     = getEnvVar("BACKUP_METRIC_BINDADDRESS", ":8080")
```

constants/constants.go
```diff
@@ -138,6 +139,10 @@ func GetBackupAnnotation() string {
 	return backupAnnotation
 }
 
+func GetMetricBind() string {
+	return metricBind
+}
```
ccremer, line 144:
```suggestion
func GetMetricBindAddress() string {
	return metricBindAddr
}
```

executor/generic.go
```diff
@@ -108,6 +108,14 @@ func (g *generic) GetRepository() string {
 	return g.Repository
 }
 
+func (g *generic) GetNamespace() string {
+	return g.Obj.GetMetaObject().GetNamespace()
+}
+
+func (g *generic) GetType() string {
+	return g.Obj.GetType()
+}
```
ccremer, line 117: I don't exactly know what this "GetType" means. What type of type are we talking about? Go types?

main.go
ccremer: I guess it's to make K8up "configurable with EnvVars only"? Best option IMO would be to use Viper or even better Koanf to make K8up configurable with both flags and Envvars, but that introduces a new library.

observer/observer.go
```diff
@@ -185,3 +214,13 @@ func (o *Observer) RegisterCallback(name string, callback func()) {
...
+
+func incFailure(namespace string) {
+	metricsFailureCounter.WithLabelValues(namespace).Inc()
+	metricsTotalCounter.WithLabelValues(namespace).Inc()
+}
+
+func incSuccess(namespace string) {
+	metricsSuccessCounter.WithLabelValues(namespace).Inc()
+	metricsTotalCounter.WithLabelValues(namespace).Inc()
+}
```
ccremer, line 226: Let's make the method names a little more descriptive
```suggestion
func incFailureCounters(namespace string) {
	metricsFailureCounter.WithLabelValues(namespace).Inc()
	metricsTotalCounter.WithLabelValues(namespace).Inc()
}

func incSuccessCounters(namespace string) {
	metricsSuccessCounter.WithLabelValues(namespace).Inc()
	metricsTotalCounter.WithLabelValues(namespace).Inc()
}
```

ccremer (changes requested):

executor/generic.go
susana-garcia: it's the type of the CDRs and can have these string values: backup, check, archive, prune and restore.

Kidswiss: Maybe `GetJobType()` would be a clearer name?

susana-garcia: that sounds better to me, thanks

susana-garcia: I'll also update the GetNamespace to GetJobNamespace, if that's ok

main.go
susana-garcia: In master it was made setting a default value using viper. Should I use Koanf package instead and take the value from the EnvVars or set a default (if the value is not present in EnvVars)? With flags you meant to have a default, right?

Kidswiss: It makes sense to have this configurable via EnvVars. I don't know Koanf, but jusdging from their Github page it looks like a lighter Viper alternative. So that's fine by me.

susana-garcia: ok, I'll use Koanf, thanks

ccremer: You can have a look into Seiso where we use Koanf too: https://github.com/appuio/seiso/blob/master/cmd/root.go#L49

susana-garcia: @ccremer should I also follow the same guidelines for the env vars loading or it’s not something generic you use in all projects? https://github.com/appuio/seiso/blob/fddf46ce9081c232cb98084f22abe254fd889165/cmd/root.go#L101

ccremer:
I took the liberty and fixed the env var loading with koanf, should work now.

susana-garcia:
@ccremer thanks, so then it's ready for another review :)

susana-garcia:
@ccremer well, just the refactoring of the constants.go is missing, I'll do that on Monday

ccremer:
Hi Susana
We discussed this today and think that the internal config refactoring should be separate from this PR. Let's get this PR (that only addresses prometheus) first and open a new PR where we overhaul the internal config for all the constants.

I'm allowing myself to move the latest Koanf commit to a new PR

ccremer (approved):

susana-garcia:
@ccremer sure, no problem

ccremer:
I was able to curl `localhost:8080/metrics`, but the k8up metric only show up when a value has been added. I made an uncommitted increment from main.go to verify that, and it shows up.

@Kidswiss from my PoV, this can now be merged

tobru (approved):

Kidswiss (approved):
Reviewed the code, the location where the metrics are incremented and decremented look sensible.

Also the koanf stuff looks to be separated now?

PR #147, opened 2020-11-18
fix: only add ResticPassword if set

mweibel:
see #135 for context - @Kidswiss mentioned that the RESTIC_PASSWORD setting is lost with the new implementation. 

I believe this was responsible for the Restic password not being merged correct. As it's been set by default even if the backend didn't specify, the global default wasn't being used instead. 
I added a test to clarify the way Mergo merges.


mweibel:
also I think: Mergo is not really necessary if the idea is just to add keys into the map if they haven't been defined yet

srueg (approved):

Kidswiss:
Just tested this, seems to fix the dropped `RESTIC_PASSWORD` thanks!

PR #148, opened 2020-11-19
Deduplicate prune 

mweibel:
This PR is based on #147 (that's why those commits are in there as well. Will need to rebase after merging that PR). 

This change ensures that the restic repository is always available and set on jobs which is required for the queue to be able to detect which jobs are running against which repository. It adds a test related to this case to document and ensure what the execution queue does works. 

Additionally it refactors to have a type for the possible job types. 

I'm not entirely sure if this solves #118 already. There's code (worker.go) which ensures a job isn't run if there's another exclusive job running or an exclusive job isn't running if there is any other job running. 
However that same code also is probably misbehaving (I'd still have to check, or rather: one should write a test for this):

1. job A (exclusive) is running
2. meanwhile job B is added to the queue
3. queue pops the job off the queue and checks if `shouldRun` -> false

At this point job B is never run, as it's gone from the queue and doesn't get re-added. 
Correct me if I misinterpret the code please :)

### DIFF
--- a/api/v1alpha1/job_types.go
+++ b/api/v1alpha1/job_types.go
@@ -1,0 +1,13 @@
+package v1alpha1
+
+// Type defines what job type this is.
+type Type string
+
+const (
+	BackupType   Type = "backup"
+	CheckType    Type = "check"
+	ArchiveType  Type = "archive"
+	RestoreType  Type = "restore"
+	PruneType    Type = "prune"
+	ScheduleType Type = "schedule"
+)
--- a/api/v1alpha1/archive_types.go
+++ b/api/v1alpha1/archive_types.go
@@ -74,8 +74,8 @@ func (a *Archive) GetMetaObject() metav1.Object {
 	return a
 }
 
-func (*Archive) GetType() string {
-	return "archive"
+func (*Archive) GetType() Type {
+	return ArchiveType
 }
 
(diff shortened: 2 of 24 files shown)


api/v1alpha1/job_types.go
```diff
@@ -0,0 +1,13 @@
+package v1alpha1
+
+// Type defines what job type this is.
+type Type string
```
ccremer, line 4: I think this is highly confusing with the built-in Go's "type"
```suggestion
// JobType defines what kind a certain job is.
type JobType string
```

ccremer (changes requested):

Kidswiss:
With the additional information from the customer in #118 unfortunately just skipping if another one of these jobs is already running isn't the desired result here.

The main problem they want to solve is, that they have multiple backups pointing to the same repository which all have a prune job defined that runs weekly at random days. But they don't want the prunes to run for each definition, instead the operator should see that there are multiple prune definitions that all run during the same period (like weekly) and then only run it once.

Kidswiss:
@ccremer deleting the development branch killed this PR :D

We agreed in the call yesterday that it would still make sense to merge it...

@mweibel Can you open a new one this time to master? thx

mweibel:
ah right that's the reason ;)
yeah, on it already :+1:

PR #149, opened 2020-11-23
Rework Workflows, build

ccremer:
* Aligns workflows with https://github.com/vshn/espejo
* Introduces Goreleaser for building the binary, docker images and changelogs in workflows
  * I'm not sure what to do yet with CHANGELOG.md, it's already outdated... should we remove this?
  * After a release has been created by Goreleaser, we can still copy the release changelog from CHANGELOG.md if we want to keep this. At least Goreleaser prepares a nice changelog format.
* The CRD of K8up is attached to K8up releases for easy download or apply to clusters
* Dockerfile and Go updated to 1.15
* Added Renovate
* Adds more nice badges


ccremer:
wrong source branch

PR #150, opened 2020-11-23
Rework Workflows and Build

ccremer:
* Aligns workflows with https://github.com/vshn/espejo
  * The `latest` image tag now points to the latest git release instead of master. Do we want to keep pointing `latest` to master or only point latest to a stable release?
  * The floating `master` image tag points to the latest commit in master branch
* Introduces Goreleaser for building the binary, docker images and changelogs in workflows
  * After a release has been created by Goreleaser, we can still adjust the release changelog from CHANGELOG.md if needed.
* The CRD of K8up is attached to K8up releases for easy download or apply to clusters
* Dockerfile and Go updated to 1.15
* Adds K8up mirror image on quay.io
* Added Renovate
* Adds more nice badges


tobru:
> The latest image tag now points to the latest git release instead of master. Do we want to keep pointing latest to master?

The idea of pointing :latest to the latest release sounds interesting. What about having a :master tag which points to the master branch?

> I'm not sure what to do yet with CHANGELOG.md, it's already outdated... should we remove this?

Yes, let's remove it and let Goreleaser to it's job.

tobru (approved):

ccremer:
> The idea of pointing :latest to the latest release sounds interesting. What about having a :master tag which points to the master branch?

We could that as well. Also interesting is to have a floating Major-SemVer tag. At the moment, K8up is a `v0` release, so it would be `v0`, but once mature and we release 1.0, there's an additional `v1` tag. Similar to Espejo: https://hub.docker.com/r/vshn/espejo/tags?page=1&ordering=last_updated

tobru (approved):
Looks good to me, thanks

PR #151, opened 2020-11-25
Setup KIND integration

ccremer:
* Integrates KIND into makefile, just run `make run_kind` and it will
  * Install the CRDs
  * Run the Operator with this kubeconfig
* Makes the CRD compatible with K8s v1.18 again (the CRD generated was invalid):
```
The CustomResourceDefinition "prebackuppods.backup.appuio.ch" is invalid: 
* spec.validation.openAPIV3Schema.properties[spec].properties[pod].properties[spec].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[pod].properties[spec].properties[containers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
make: *** [Makefile:163: setup_e2e_test] Error 1
```


ccremer:
wrong branch again

PR #152, opened 2020-11-25
Integrate KIND into makefile

ccremer:
* Integrates KIND into makefile, just run `make run_kind` and it will
  * Install the CRDs
  * Run the Operator with this kubeconfig
* Makes the CRD compatible with K8s v1.18 again (the CRD generated was invalid):
```
The CustomResourceDefinition "prebackuppods.backup.appuio.ch" is invalid: 
* spec.validation.openAPIV3Schema.properties[spec].properties[pod].properties[spec].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[pod].properties[spec].properties[containers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
make: *** [Makefile:163: setup_e2e_test] Error 1
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,5 +34,6 @@ bin/
 dist/
 k8up-crd.yaml
 k8up
+testbin/
 
 __debug_bin
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,12 @@ CRD_OPTIONS ?= "crd:trivialVersions=true"
 
 CRD_FILE ?= k8up-crd.yaml
 
+TESTBIN_DIR ?= ./testbin/bin
+KIND_BIN ?= $(TESTBIN_DIR)/kind
+KIND_VERSION ?= 0.9.0
+KIND_KUBECONFIG ?= ./testbin/kind-kubeconfig
+KIND_CLUSTER ?= k8up
+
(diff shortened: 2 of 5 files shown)


tobru (approved):
`hack/fix-crd.go` might need some explanation what it does and why it is needed, either with a comment, better filename or hint in the README. I'm also not sure if the file is not well formatted?

ccremer:
> `hack/fix-crd.go` might need some explanation what it does and why it is needed, either with a comment, better filename or hint in the README. I'm also not sure if the file is not well formatted?

I left comments in the `Makefile` as for the reasons...
But yes, it's not properly formatted. `go fmt ./...` doesnt catch this file, since there's the `// +build fix-crd` and that excludes it from fmt...

tobru:
> I left comments in the Makefile as for the reasons...

Yes, but I'd rather want to see these comments in the go file to not have them disconnected. When I see such a file in a repo I would have no idea where to look for a description what it does, also with such an non-descriptive filename.

ccremer:
I renamed the file and moved the comment

Issue #153, opened 2020-11-25
CRD compatibility support discussion


ccremer:
### Issue

The current CRD generated is of version `apiextensions.k8s.io/v1beta1`, but some generated properties aren't valid for that version in K8s 1.18+. The generator currently used could do `apiextensions.k8s.io/v1` if we upgrade it, but `v1` is not available in older Kubernetes server versions (e.g. OpenShift 3.11), that means we would not be able to install K8up.

### Proposal

In order to stay compatible with both Openshift 3.11 and also Kubernetes 1.18+ (Rancher, K3s etc) I propose:

1. Generating the CRD with controller-gen 0.4+ but for spec v1beta1, make the necessary post-generate-patches to make it work. The YAML files are saved in `config/crd/apiextensions.k8s.io/v1beta1/` (see #152 )
2. Additionally generate the CRD with controller-gen 0.4+, save the YAML files in `config/crd/apiextensions.k8s.io/v1/`
3. Make both CRD versions available in GitHub releases for download (attachment)
4. For development we use `v1` to stay up-to-date with Operator SDK (which will eventually also migrate to controller-gen 0.4+) and K8s API.
    * Also because we are developing the Operator against K8s 1.18 API and beyond
    * Use GitHub actions to run e2e tests (with KIND) for a pre-1.18 cluster and post-1.18 cluster version (matrix jobs)

### Background

* In #114 we have migrated K8up to Operator SDK 1.0+. Operator SDK uses controller-gen version 0.3.0
* controller-gen 0.4.x uses `v1` by default
* Reason why it can't be installed:
```
The CustomResourceDefinition "prebackuppods.backup.appuio.ch" is invalid: 
* spec.validation.openAPIV3Schema.properties[spec].properties[pod].properties[spec].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[pod].properties[spec].properties[containers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
make: *** [Makefile:163: setup_e2e_test] Error 1
```

* v1beta1 is deprecated and will be removed in 1.22:
```
[...]

PR #154, opened 2020-11-26
Ensure legacy and current CRD spec version support

ccremer:
* Makefile now generates CRD specs for v1 and v1beta1 (see #153 )
* Structural changes in the CRD directories
* Scaffolds e2e tests for local and CI/CD runs (actual framework is left open for later)


tobru (approved):

Makefile
```diff
@@ -121,7 +132,7 @@ ifeq (, $(shell which controller-gen))
 	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
 	cd $$CONTROLLER_GEN_TMP_DIR ;\
 	go mod init tmp ;\
-	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0 ;\
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1 ;\
```
Kidswiss, line 135: Idea: we could use a `generate.go` like we did in the lieutenant and trigger the generation through that. https://github.com/projectsyn/lieutenant-operator/blob/master/generate.go

Pros: the version of controller-gen can be managed at one place via the `go.mod` file. Also it would automatically go get the right version if not available.

ccremer: That's interesting, I'll try it out!

Makefile
```diff
@@ -100,8 +109,8 @@ lint: fmt vet
 	git diff --exit-code
 
 # Generate code
-generate: controller-gen
-	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+generate: vet
+	go run sigs.k8s.io/controller-tools/cmd/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
```
Kidswiss, line 113: I feel like we can use the `go generate` approach here, too. What do you think?

Kidswiss:
Make generate currently creates the `v1` and `v1beta1` folders in the project root. Is that really intended?

ccremer:
> `make manifests` currently creates the `v1` and `v1beta1` folders in the project root. Is that really intended?

No it's not. Can you try with latest changes? Are you on Mac or Linux?

Makefile
ccremer: I have unified all generated stuff into one generate.go file. It takes several seconds now, thus I removed the generate target dependency from Makefile's run and test targets, so it'll start faster without generating. otherwise it's possible run just `make generate run` with the same result

Kidswiss: Now it's creating the CRDs in the correct folder 👍

Kidswiss:
Now it's creating the CRDs in the correct folder 👍

Kidswiss (approved):

Issue #121, opened 2020-10-27
New configuration concept


Kidswiss:
Currently defaults for backups can be defined by setting environment variables. While this servers its purpose it's not very flexible and rather cumbersome to configure.

So I propose a new way to configure and schedule backups. There should be two high level objects for that:
* jobPlan
* schedule

The jobPlan will what jobs and their defaults. For example the repository, credentials and encryption password. These plans can then be referenced by schedules. The schedules can overwrite or append the defaults defined in the plans. For example the plan defines the S3 endpoint and the schedule defines the bucket. The namespace where the schedule is created will also be the namespace where the backups will run. The plans on the other hand can live in arbitrary namespaces.

Some use cases:
* provide customers a sensible default for a managed cluster
* manage uniform backups for all namespaces in a non-shared cluster

This makes the default handling much more flexible.

PR #155, opened 2020-11-26
Migrate constants to Koanf

ccremer:
New internal configuration library [Koanf](https://github.com/knadh/koanf) (pronouned "konf") adds more flexibility into configuring the K8up operator itself.

* Introduces a new Configuration struct, that allows easier access than the tons of public Getters
* Koanf can be configured to also respect CLI flags (if there are any), config files etc with flexible merging hierarchy. It's not used yet, but it allows something similar to Viper, but it's more flexible and lightweight.


ccremer:
@susana-garcia Do you have permissions to push to this branch?

susana-garcia:
@ccremer no, I don't

Kidswiss:
Makes sense to use some small framework to manage the configuration, although a ton of the current envVars will die with #121.

cfg/config.go
```diff
@@ -0,0 +1,60 @@
...
+	PodFilter                      string `koanf:"podfilter"`
+	PromURL                        string `koanf:"promurl"`
+	RestartPolicy                  string `koanf:"restartpolicy"`
+}
+
+var (
+	Config = NewDefaultConfig()
+)
+
+const (
```
susana-garcia, line 34: I'd add the const block at the beginning

susana-garcia (commented):
Looks good to me, I just add a small comment

PR #160, opened 2020-11-30
Recover documentation

ccremer:
https://github.com/vshn/k8up/commit/758ca6f1427c7b9e0fb92fbf74fdff5b3b0882cf removed all documentation, this PR re-adds it

Please check if all is still up-to-date and needed.

I think there are some leftover files in `manifest/install` -> should we migrate to the Kustomize setup in `config/` instead? Or delete it?


tobru:
> Please check if all is still up-to-date and needed.

I'll do a docs review later this week and open separate PRs to update it. But it would be good to have the current docs in place again.

> I think there are some leftover files in manifest/install -> should we migrate to the Kustomize setup in config/ instead? Or delete it?

A Kustomize setup would be great.

tobru (approved):

Kidswiss (approved):

ccremer:
Do we still want Vale?

tobru:
> Do we still want Vale?

Nope, kill it with :fire:

ccremer:
@Kidswiss I've also taken the liberty to fix https://github.com/vshn/k8up/issues/141, which is just a documentation issue

Kidswiss (approved):

PR #161, opened 2020-11-30
Print version number at K8up startup

ccremer:
This should be easy and aligns with https://github.com/vshn/espejo/blob/master/main.go

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,6 +21,11 @@ import (
 )
 
 var (
+	// These will be populated by Goreleaser
+	version string
+	commit  string
+	date    string
+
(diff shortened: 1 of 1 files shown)


tobru (approved):

Kidswiss:
@ccremer will the vars be set during build time like it's done with the lieutenant-operator? https://github.com/projectsyn/lieutenant-operator/blob/master/Makefile#L45

ccremer:
Correct, it's defined in https://goreleaser.com/customization/build/ (somewhere a comment in the yaml)

PR #164, opened 2020-12-01
feat: limit max concurrent jobs by type (prune, backup, check, etc.)

### LINKED ISSUE #120: Limit max concurrent jobs
The operator should have the ability to limit how many jobs should get run concurrently.

This limit should be configurable by job type (prune, backup, check, etc.)

susana-garcia:
Fixes #120

### DIFF
--- a/observer/observer.go
+++ b/observer/observer.go
@@ -9,6 +9,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/vshn/k8up/cfg"
 	batchv1 "k8s.io/api/batch/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
@@ -56,6 +57,7 @@ type Observer struct {
 // ObservableJob defines a batchv1.job that is being observed by the Observer.
 type ObservableJob struct {
 	Job        *batchv1.Job
+	JobType    string
 	Event      EventType
 	Exclusive  bool
 	Repository string
@@ -200,6 +202,58 @@ func (o *Observer) IsAnyJobRunning(repository string) bool {
 	}
 
 	return false
+}
+
+// IsLimitConcurrentJobsReached checks if the limit of concurrent jobs by type (backup, check, etc)
+// has been reached
+func (o *Observer) IsLimitConcurrentJobsReached(jobType, repository string) bool {
+	listOfJobs := o.GetJobsByRepository(repository)
+	fmt.Println("type: ", jobType)
+	switch jobType {
+	case "backup":
+		l := cfg.Config.GlobalConcurrentArchiveJobsLimit
+		return filterByJobType(l, jobType, listOfJobs)
+	case "check":
+		l := cfg.Config.GlobalConcurrentCheckJobsLimit
+		return filterByJobType(l, jobType, listOfJobs)
+	case "archive":
+		l := cfg.Config.GlobalConcurrentArchiveJobsLimit
+		return filterByJobType(l, jobType, listOfJobs)
+	case "prune":
+		l := cfg.Config.GlobalConcurrentPruneJobsLimit
+		return filterByJobType(l, jobType, listOfJobs)
+	case "restore":
+		l := cfg.Config.GlobalConcurrentRestoreJobsLimit
+		return filterByJobType(l, jobType, listOfJobs)
+	default:
+		// In case of type "" -> Job
+		return false
+	}
+}
+
+func filterByJobType(limit int, jobType string, listOfJobs []ObservableJob) bool {
+	if limit > 0 {
+		if len(listOfJobs) == 0 {
+			return false
+		}
+
+		result := filter(listOfJobs, func(val ObservableJob) bool {
+			return val.JobType == jobType
+		})
+
+		return len(result) > limit
+	}
+	return false
+}
+
+func filter(arr []ObservableJob, cond func(ObservableJob) bool) []ObservableJob {
+	result := []ObservableJob{}
+	for i := range arr {
+		if cond(arr[i]) {
+			result = append(result, arr[i])
+		}
+	}
+	return result
 }
 
 // RegisterCallback will register a function to the given observed job.
--- a/observer/observer.go
+++ b/observer/observer.go
@@ -15,13 +15,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
-var (
+const (
 	Update   EventType = "update"
 	Delete   EventType = "delete"
 	Create   EventType = "create"
 	Failed   EventType = "failed"
 	Suceeded EventType = "suceeded"
 	Running  EventType = "running"
+)
+
+var (
 	observer *Observer
 
 	promLabels = []string{
--- a/observer/observer.go
+++ b/observer/observer.go
@@ -9,7 +9,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	"github.com/vshn/k8up/cfg"
+	"github.com/vshn/k8up/api/v1alpha1"
 	batchv1 "k8s.io/api/batch/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
@@ -60,7 +60,7 @@ type Observer struct {
 // ObservableJob defines a batchv1.job that is being observed by the Observer.
 type ObservableJob struct {
 	Job        *batchv1.Job
-	JobType    cfg.JobType
+	JobType    v1alpha1.JobType
 	Event      EventType
 	Exclusive  bool
 	Repository string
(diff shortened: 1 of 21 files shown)


susana-garcia:
I still need to refactor the constants bit with the new koanf configuration part, after that PR is merged

ccremer:
PR #155 is now merged, you could do a rebase so that the diff is more clear :)

susana-garcia:
@ccremer already done :)

observer/observer.go
```diff
@@ -202,6 +204,58 @@ func (o *Observer) IsAnyJobRunning(repository string) bool {
...
+	case "prune":
+		l := cfg.Config.GlobalConcurrentPruneJobsLimit
+		return filterByJobType(l, jobType, listOfJobs)
+	case "restore":
+		l := cfg.Config.GlobalConcurrentRestoreJobsLimit
+		return filterByJobType(l, jobType, listOfJobs)
+	default:
+		// In case of type "" -> Job
+		return false
+	}
```
ccremer, line 231: All job implement the Executor interface. Does it make sense to add a Getter there for accessing the limit? This would eliminate this switch-case with very similar code.
@Kidswiss ?

observer/observer.go
```diff
@@ -202,6 +204,58 @@ func (o *Observer) IsAnyJobRunning(repository string) bool {
 	return false
 }
 
+// HasLimitConcurrentJobsReached checks if the limit of concurrent jobs by type (backup, check, etc)
+// has been reached
+func (o *Observer) HasLimitConcurrentJobsReached(jobType, repository string) bool {
+	listOfJobs := o.GetJobsByRepository(repository)
+	fmt.Println("type: ", jobType)
+	switch jobType {
```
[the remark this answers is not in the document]
susana-garcia: Ok, but I guess I'll add the new type under `executor/generic.go`, what do you think?

observer/observer.go
susana-garcia: @ccremer in the Getter, I should pass the job type to return the proper limit value, right?

ccremer: I thought something about this:
```go
$ cat queue/execution.go
type Executor interface {
	...
	GetConcurrencyLimit() int
}

$ cat executor/restore.go
func (r *RestoreExecutor) GetConcurrencyLimit() int {
	return cfg.Config.GlobalConcurrentRestoreJobsLimit
}
```
And return 0 or -1 if no limit set or supported. Or the function returns an error, that would make it explicit (I regard returning 0 or -1 sometimes as lazyness when something doesn't feel right regarding limits)

susana-garcia: you're completely right, thanks

ccremer:
Can you add some unit tests as well? We should really start with those :)

susana-garcia:
@ccremer I'm adding some unit tests to test the observer methods, is that ok?

observer/observer.go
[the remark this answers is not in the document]
Kidswiss: Isn't there a JobType type somewhere already? I thought I saw it implemented somewhere...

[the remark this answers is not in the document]
susana-garcia: @Kidswiss as far as I know, before we only had the job type as string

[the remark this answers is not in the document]
mweibel: see #148 ;)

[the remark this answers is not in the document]
Kidswiss: Right, thanks @mweibel! Knew I saw it somewhere but wasn't able to find it :D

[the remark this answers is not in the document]
susana-garcia: fantastic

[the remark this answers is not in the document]
susana-garcia: I'll wait until the other PR is on master to redo that part then

susana-garcia:
@ccremer  @Kidswiss also I forgot to ask: for the tests we want to have BDD testing? I saw that we're already using `ginkgo` or are you thinking on something else?

observer/observer.go
```diff
@@ -202,6 +207,34 @@ func (o *Observer) IsAnyJobRunning(repository string) bool {
...
+
+func filter(arr []ObservableJob, cond func(ObservableJob) bool) []ObservableJob {
+	result := []ObservableJob{}
+	for i := range arr {
+		if cond(arr[i]) {
+			result = append(result, arr[i])
+		}
+	}
+	return result
+}
```
ccremer, line 236: I think this can be simplified with an early exit:
```suggestion
func (o *Observer) IsLimitConcurrentJobsReached(jobType cfg.JobType, limit int, repository string) bool {
	if limit <= 0 {
		return false
	}
	listOfJobs := o.GetJobsByRepository(repository)
	if len(listOfJobs) == 0 {
		return false
	}

	count := 0
	for _, job := range listOfJobs {
		if job.JobType == jobType {
			count++
			if count >= limit {
				return true
			}
		}
	}
	return false
}
```

queue/execution.go
```diff
@@ -72,7 +75,7 @@ func (eq *ExecutionQueue) Add(exec Executor) {
 		eq.queues[repository] = newPriorityQueue()
 	}
 	eq.queues[repository].add(exec)
-	eq.incQueueGauge(exec.GetJobNamespace(), exec.GetJobType())
+	eq.incQueueGauge(exec.GetJobNamespace(), string(exec.GetJobType()))
```
ccremer, line 78: I think conversion to string should happen in the `incQueueGauge` itself and not in the caller

ccremer:
> @ccremer I'm adding some unit tests to test the observer's methods, is that ok?

Yes, you can limit tests for the relevant change in this PR only

observer/observer.go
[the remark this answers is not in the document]
ccremer: Right, I already forgot that :)

ccremer:
> @ccremer @Kidswiss also I forgot to ask: for the tests we want to have BDD testing? I saw that we're already using `ginkgo` or are you thinking on something else?

I think for this change it's sufficient to create classic unit tests with `testing.T`. I don't think you need the in-memory K8s API server to test the limits, unless it's very hard to mock/stub the dependent functions

observer/observer.go
```diff
@@ -202,6 +207,34 @@ func (o *Observer) IsAnyJobRunning(repository string) bool {
 	return false
 }
 
+// IsLimitConcurrentJobsReached checks if the limit of concurrent jobs by type (backup, check, etc)
+// has been reached
+func (o *Observer) IsLimitConcurrentJobsReached(jobType cfg.JobType, limit int, repository string) bool {
+	listOfJobs := o.GetJobsByRepository(repository)
```
Kidswiss, line 213: This will only return all the jobs running for this repository. So this function currently checks the limit per type per repository. That's not quite what this feature is about. For example, if we set the backup limit to 10 and for repository a there are 5 running backups and for repository b there are 6 running backups it will still trigger a job for repository c resulting in 12 running backups which would be over the defined limit of 10 backups.

The issue isn't about the amount of jobs that run against one repository but rather how many jobs are running on the cluster and may overload it (IO, memory, network, amount of running pods, etc.)

So the observer needs to be extended so it can return the amount of running jobs by type for all repositories.

Kidswiss (changes requested):

observer/observer.go
susana-garcia: @Kidswiss I see, I'll change that part then, thanks

susana-garcia:
@ccremer @Kidswiss I think it's ready now, but please review it again.

observer/observer.go
```diff
@@ -202,6 +207,38 @@ func (o *Observer) IsAnyJobRunning(repository string) bool {
...
+
+func filter(arr []ObservableJob, cond func(ObservableJob) bool) []ObservableJob {
+	result := []ObservableJob{}
+	for i := range arr {
+		if cond(arr[i]) {
+			result = append(result, arr[i])
+		}
+	}
+	return result
+}
```
ccremer, line 240: I think you can remove this function?

observer/observer.go
```diff
@@ -202,6 +207,38 @@ func (o *Observer) IsAnyJobRunning(repository string) bool {
 	return false
 }
 
+// IsLimitConcurrentJobsReached checks if the limit of concurrent jobs by type (backup, check, etc)
+// has been reached
+func (o *Observer) IsLimitConcurrentJobsReached(jobType v1alpha1.JobType, limit int) bool {
```
ccremer, line 212: It should be either `IsLimitOfConcurrentJobsReached` or this one:
```suggestion
func (o *Observer) IsConcurrentJobsLimitReached(jobType v1alpha1.JobType, limit int) bool {
```

ccremer (changes requested):

susana-garcia:
@ccremer @Kidswiss please review it again when you have a moment.

observer/observer_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	ojs := make(map[string]ObservableJob)
+	ojs["some-name"] = oj
+	fmt.Println(len(ojs))
+
+	o.observedJobs = ojs
+
+	// 1 observedJob and limit = 1
+	isLimitReached = o.IsConcurrentJobsLimitReached(v1alpha1.BackupType, 1)
+	assert.Equal(t, true, isLimitReached)
```
ccremer, line 49: I think there are `assert.True()` and `assert.False()` shortcuts

susana-garcia:
@ccremer I think it's ready

ccremer (approved):
I cannot really judge whether the business logic does what is needed to. I'll let @Kidswiss judge that, but code LGTM

susana-garcia:
@ccremer ok, thank you

observer/observer.go
```diff
@@ -202,6 +207,28 @@ func (o *Observer) IsAnyJobRunning(repository string) bool {
 	return false
 }
 
+// IsConcurrentJobsLimitReached checks if the limit of concurrent jobs by type (backup, check, etc)
+// has been reached
+func (o *Observer) IsConcurrentJobsLimitReached(jobType v1alpha1.JobType, limit int) bool {
+	if limit <= 0 {
+		return false
+	}
+	listOfJobs := o.observedJobs
```
Kidswiss, line 216: I'd mutex lock this assignment:

```suggestion
        o.mutex.Lock()
	listOfJobs := o.observedJobs
	o.mutex.Unlock()
```

Because writing to the map happens in a separate go routine.

The rest LGTM.

Kidswiss (changes requested):

susana-garcia:
@Kidswiss I think the PR is ready

Kidswiss (approved):

PR #169, opened 2020-12-01
feat: ensure repository is always set

mweibel:
This is a reopening of #148 

This change ensures that the restic repository is always available and set on jobs which is required for the queue to be able to detect which jobs are running against which repository. It adds a test related to this case to document and ensure what the execution queue does works.

Additionally it refactors to have a type for the possible job types.

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -1,5 +1,7 @@
 package cfg
 
+import "fmt"
+
 const (
 	RestoreS3EndpointEnvName        = "RESTORE_S3ENDPOINT"
 	RestoreS3AccessKeyIDEnvName     = "RESTORE_ACCESSKEYID"
@@ -62,4 +64,9 @@ func NewDefaultConfig() *Configuration {
 		MetricsBindAddress:      ":8080",
 		PodFilter:               "backupPod=true",
 	}
+}
+
+// GetGlobalRepository is a shortcut for building an S3 string "s3:<endpoint>/<bucket>"
+func GetGlobalRepository() string {
+	return fmt.Sprintf("s3:%s/%s", Config.GlobalS3Endpoint, Config.GlobalS3Bucket)
 }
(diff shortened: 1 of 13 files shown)


cfg/config.go
```diff
@@ -58,3 +60,7 @@ func NewDefaultConfig() *Configuration {
 		PodFilter:               "backupPod=true",
 	}
 }
+
+func GetGlobalRepository() string {
+	return fmt.Sprintf("s3:%s/%s", Config.GlobalS3Endpoint, Config.GlobalS3Bucket)
+}
```
mweibel, line 66: it doesn't feel right to put this in this package. But I'm not sure where to better put it. Before rebasing it's been in `constants.go`. 
Any idea where it'd fit better?

controllers/archive_controller.go
```diff
@@ -39,7 +40,11 @@ func (r *ArchiveReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, err
 	}
 
-	config := job.NewConfig(ctx, r.Client, log, archive, r.Scheme)
+	repository := cfg.GetGlobalRepository()
```
Kidswiss, line 43: I'm not sure how much sense these snippets make. Because the `Backend.String()` basically already does what `GetGlobalRepository()` does but will also consider if there are any overrides within the `Backend` instance. See https://github.com/vshn/k8up/blob/master/api/v1alpha1/backend.go#L87

Kidswiss (changes requested):

cfg/config.go
Kidswiss: I feel like this function is redundant anyways, see my other comment about the `Backend.String()` function. If there's still need for this function though, the `Backend` type would sound like a good home for this I guess.

controllers/archive_controller.go
mweibel: the point is: if Backend is nil, that won't work.

Kidswiss: Ah right, yes then it makes sense.

Kidswiss (approved):

ccremer:
I have rebased the branch and fixed some conflicts. Waiting on checks to complete

PR #170, opened 2020-12-02
Add e2e test with bats-detik

ccremer:
It took while with try-and-error, but I got it running

This PR adds e2e testing libary and 1 e2e test. 
The setup process is as following:

1. Install BATS (requires local node and npm, which is a prerequisite)
2. Start a local docker registry in docker
3. Install KIND and set up a KIND cluster just for K8up
4. Connect the KIND cluster with the docker registry

Now the initial setup is done (only to be done once). For each e2e test run, we

1. Push the local docker image to the local docker registry, so that KIND can pull the image from local registry
2. Execute BATS
    * which installs the Operator with Kustomize including CRDs
    * it means we are actually testing our container image and deployment
3. Assert outcome with the Syntax that DETIK provides

The current single test case verifies if the Operator can be started from a Deployment (Pod status "running"), but does not do anything else yet. But from this point on it should be trivial to add more assertions that specifically check for logs, status fields in objects etc.

The bash files in `e2e/lib` are vendored (with exception of `e2e/lib/k8up.bash`) and coming from https://github.com/bats-core/bats-detik/tree/master/lib, but I have not created a make target that updates a newer download of it.

This should be executed for both K8s 1.18+ and Openshift-compatible 1.13 (GH action takes care of that)

One thing to note is that by now this e2e test in GH takes between 3 and 5 minutes to run.

Links:
- https://kind.sigs.k8s.io/docs/user/local-registry/
- https://github.com/bats-core/bats-detik
- https://stackoverflow.com/questions/1789594/how-do-i-write-the-cd-command-in-a-makefile

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,7 +53,9 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-go-
     - name: Run e2e tests
-      run: make crd e2e_test -e CRD_SPEC_VERSION=${{ matrix.crd-spec-version }} -e KIND_NODE_VERSION=${{ matrix.kind-node-version }} -e KIND_KUBECTL_ARGS=--validate=false
+      run: make crd install_bats setup_e2e_test e2e_test -e CRD_SPEC_VERSION=${{ matrix.crd-spec-version }} -e KIND_NODE_VERSION=${{ matrix.kind-node-version }} -e KIND_KUBECTL_ARGS=--validate=false
+    - name: Show e2e debug logs
+      run: cat e2e/debug/detik/*
   image:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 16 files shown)


ccremer:
This is now ready for review.

Do you guys think this is worth to using or is it becoming too complicated?

tobru:
> Do you guys think this is worth using or is it becoming too complicated?

I feel this could absolutely work and we should give it a try to make use of it.

> Does it require more documentation now or should this rather separately with documentation rewrite?

I'd do it separately. It's important to link to the upstream tool documentation and this has been done, so fine by me.

tobru (approved):

Kidswiss (approved):
Like Freddy Mercury said: "It's a kind of magic" but looks nice how the actual test cases can be defined.

LGTM

ccremer:
> how the actual test cases can be defined.

You don't need to use the built-in syntax from DETIK. Any command that fails with non-zero exit code will fail the test. So you can certainly do `kubectl get backup my-backup -o yaml | grep my-string` if that works easier.

PR #175, opened 2020-12-03
Allow default resources to be set on all job types

### LINKED ISSUE #143: Make the backup jobs request and limits configurable
## Summary

**As** K8up admin\
**I want** to override the default resource request and limits of Pods generated by K8up\
**So that** I can optimize resource usage or comply with cluster or namespace resource policies.

## Context

On clusters with a default pod cpu/memory limit, backups jobs are currently limited to these default values, because there is no possibility to override or remove them from k8up.
```
Resource Limits
 Type                      Resource                 Min  Max  Default Request  Default Limit  Max Limit/Request Ratio
 ----                      --------                 ---  ---  ---------------  -------------  -----------------------
 Container                 cpu                      1m   -    100m             250m           -
 Container                 memory                   1Mi  -    128Mi            256Mi          -
```
This limits the use cases on such clusters heavily. 

## Out of Scope

* Actually define global default resource limits or requests. Defining defaults in a non-major upgrade of K8up could potentially stop existing backups to work (crash with OOM, etc)

## Further links

* URLs of relevant Git repositories, PRs, Issues, etc.

## Acceptance criteria

*Given* a K8up `Schedule` object with per-schedule-specified resources\
*When* K8up schedules Jobs\
*Then* the containers in Pods are scheduled with configured resource request and limits

*Given* a K8up `Schedule` object outside of cluster admin's responsibility\
*When* K8up schedules Jobs\
*Then* the containers in Pods are scheduled with configured global default resource request and limits
(in a multi-tenant cluster, customers can create schedules, while cluster-admins can provide global defaults in case customer doesn't define those)

## Implementation Ideas

* Global environment variables for defaults
* The actual resource usage is heavily dependent on the amount of files that are backed up. So we should have the ability to set the limits by job and schedule, where values can be overwritten like for the S3 endpoints
  * The order of precedence could be: `global defaults < schedule defaults < job type specifics` (right overrides left)

susana-garcia:
Fixes #143


job/job.go
```diff
@@ -86,3 +89,20 @@ func GetGenericJob(obj Object, config Config) (*batchv1.Job, error) {
 
 	return job, err
 }
+
+func getK8upResourceLimits(r *corev1.ResourceRequirements) corev1.ResourceRequirements {
+	if r == nil {
```
Kidswiss, line 94: Maybe we can do this check more granularly?

Given:
* global memory limit :1Gb
* global cpu limit: 1000m

Now if the user creates a job or schedule with cpu limit 1500m an no memory limit, no memory limit will be set at all. In that case it would make more sense to apply the global default to these nil values.

job/job.go
```diff
@@ -73,8 +75,9 @@ func GetGenericJob(obj Object, config Config) (*batchv1.Job, error) {
 					RestartPolicy: corev1.RestartPolicyOnFailure,
 					Containers: []corev1.Container{
 						{
-							Name:  obj.GetMetaObject().GetName(),
-							Image: cfg.Config.BackupImage,
+							Name:      obj.GetMetaObject().GetName(),
+							Image:     cfg.Config.BackupImage,
+							Resources: getK8upResourceLimits(obj.GetResourceLimits()),
```
Kidswiss, line 80: This works perfectly for all the jobs. But the schedule will never reach this point. If the resource limits are defined in the schedule itself it should populate the resource limits of the job definitions it contains, or they won't be applied at all.

Just had an idea: the job definitions in the schedule are just instances of the various job spec types. So it would be possible to set a resource limit on the schedule itself but then also on the various job definitions, too. Example:

```
kind: Schedule
metadata:
  name: schedule-test
spec:
  k8upResourceLimits:
    requests:
        memory: "64Mi"
        cpu: "250m"
    limits:
      memory: "128Mi"
      cpu: "500m"
  backup:
    k8upResourceLimits:
      limits:
        memory: "256Mi"
        cpu: "1000m"
    schedule: '* * * * *'
    keepJobs: 4
    promURL: http://10.144.1.224:9000
  check:
    schedule: '*/5 * * * *'
    promURL: http://10.144.1.224:9000
```

In this sample the scheduled backup should receive a merge between the limits defined in the schedule and the limits in the backup definition.

The check should simply get the value defined on the schedule itself.

Kidswiss (changes requested):

job/job.go
ccremer: Hm, I'm a little bit unhappy with the name `k8upResourceLimits`. what about:
```yaml
kind: Schedule
metadata:
  name: schedule-test
spec:
  resourceRequirementsTemplate:
    requests:
        memory: "64Mi"
        cpu: "250m"
    limits:
      memory: "128Mi"
      cpu: "500m"
  backup:
    resources:
      limits:
        memory: "256Mi"
        cpu: "1000m"
    schedule: '* * * * *'
    keepJobs: 4
    promURL: http://10.144.1.224:9000
  check:
    schedule: '*/5 * * * *'
    promURL: http://10.144.1.224:9000
```

susana-garcia: @Kidswiss I have a couple of questions. 
- In case of the backup schedule, how should we handle the scenario when `resourceRequirementsTemplate` is provided only in the `backup`? Do we get the default values for `schedule` and add them as well to the ones provided in the `backup`?
- Also, if `backup` and `schedule`, don't have `resourceRequirementsTemplate`, then we just add the defaults to both, right?

ccremer: My idea and understanding is:
* The `resourceRequirementsTemplate` is only in the `.spec` of `Schedule`. This serves as a template to be overridden by `.spec.backup.resources`. So, if there is no template specified, but in `.spec.backup.resources` there are values, well, there is nothing to override, just apply directly.
In other words, `.spec.backup` doesn't have the field `resourceRequirementsTemplate`.
* I think you mean: If `.spec.backup` and `.spec.check` don't have a `resource` field, then apply the values from the `.spec.resourceRequirementsTemplate` property, right?

The point of this story is to provide defaults in the template, but make it possible to override, e.g. if a certain big database backup takes 3 gigs of memory, but all the other jobs take only 100 MB.

cfg/config.go
```diff
@@ -49,17 +53,21 @@ var (
...
+		GlobalKeepJobs:              6,
+		BackupImage:                 "172.30.1.1:5000/myproject/restic",
+		PodExecRoleName:             "pod-executor",
+		RestartPolicy:               "OnFailure",
+		MetricsBindAddress:          ":8080",
+		PodFilter:                   "backupPod=true",
+		GlobalCPUResourceRequest:    "100m",
+		GlobalCPUResourceLimit:      "250m",
+		GlobalMemoryResourceRequest: "128Mi",
+		GlobalMemoryResourceLimit:   "256Mi",
```
ccremer, line 71: @Kidswiss does it make sense to provide global defaults? What are usually wrestic's requirements? or should we leave that empty? (I think there shouldn't be a CPU limit by default for example)

job/job.go
Kidswiss: > .spec.check.resources would be left unspecified.

Only if the global default is not specified. If there's a global default it should be applied if nothing is defined.

So the flow should go like:

global defaults are overwritten by `resourceRequirementsTemplate` are overwritten by `resources`.

config/samples/k8up_v1alpha1_check.yaml
```diff
@@ -3,6 +3,10 @@ kind: Check
 metadata:
   name: check-test
 spec:
+  resourceRequirementsTemplate:
+    requests:
+        memory: "64Mi"
+        cpu: "250m"
```
ccremer, line 9:
```suggestion
  resources:
    requests:
        memory: "64Mi"
        cpu: "250m"
```

api/v1alpha1/restore_types.go
```diff
@@ -17,6 +17,8 @@ type RestoreSpec struct {
 	KeepJobs      *int           `json:"keepJobs,omitempty"`
 	// Tags is a list of arbitrary tags that get added to the backup via Restic's tagging system
 	Tags []string `json:"tags,omitempty"`
+	// ResourceRequirementsTemplate describes the compute resource requirements (cpu, memory, etc.)
+	ResourceRequirementsTemplate *corev1.ResourceRequirements `json:"resourceRequirementsTemplate,omitempty"`
```
ccremer, line 21: Can you rename this property also in the other Job types (except Schedule)?
```suggestion
	// Resources describes the compute resource requirements of the containers in the scheduled Pod
	Resources *corev1.ResourceRequirements `json:"resourceRequirementsTemplate,omitempty"`
```

api/v1alpha1/restore_types.go
```diff
@@ -73,6 +75,10 @@ func (r *Restore) GetK8upStatus() *K8upStatus {
 	return &r.Status.K8upStatus
 }
 
+func (r *Restore) GetResourceLimits() *corev1.ResourceRequirements {
+	return r.Spec.ResourceRequirementsTemplate
+}
+
```
ccremer, line 81:
```suggestion
func (r *Restore) GetResources() *corev1.ResourceRequirements {
	return r.Spec.Resources
}

```

job/job.go
susana-garcia: Ok, I understand what you mean. But in the second question I meant what happens when none are provided, but i guess I just add the default ones to the, let's call them children jobs (all but schedule). Also I realized that I was thinking on adding the `resourceRequirementsTemplate` as limits to the schedule job itself, what it's probably not needed.

susana-garcia: @Kidswiss ok, I see. But the schedule doesn't need to have limits.

job/job.go
```diff
@@ -44,6 +45,7 @@ type Object interface {
 	GetRuntimeObject() runtime.Object
 	GetK8upStatus() *k8upv1alpha1.K8upStatus
 	GetType() v1alpha1.JobType
+	GetResourceLimits() *corev1.ResourceRequirements
```
ccremer, line 48:
```suggestion
	GetResources() *corev1.ResourceRequirements
```

job/job.go
```diff
@@ -86,3 +89,35 @@ func GetGenericJob(obj Object, config Config) (*batchv1.Job, error) {
...
+	li[corev1.ResourceCPU] = resource.MustParse(cfg.Config.GlobalCPUResourceLimit)
+	li[corev1.ResourceMemory] = resource.MustParse(cfg.Config.GlobalMemoryResourceLimit)
+
+	return li
+}
+
+func getDefaultResourceRequests() map[corev1.ResourceName]resource.Quantity {
+	re := make(map[corev1.ResourceName]resource.Quantity)
+	re[corev1.ResourceCPU] = resource.MustParse(cfg.Config.GlobalCPUResourceRequest)
+	re[corev1.ResourceMemory] = resource.MustParse(cfg.Config.GlobalMemoryResourceRequest)
```
ccremer, line 120: Shouldn't there be some error handling? Users could decide to not specify CPU limits (leave empty), [as is recommended by upstream Kubernetes for production workload](https://learnk8s.io/production-best-practices). `MustParse` suggests there will be a panic on errors

job/job.go
```diff
@@ -86,3 +89,35 @@ func GetGenericJob(obj Object, config Config) (*batchv1.Job, error) {
...
+	if r.Requests == nil {
+		r.Requests = getDefaultResourceRequests()
+	}
+	return *r
+}
+
+func getDefaultResourceLimits() map[corev1.ResourceName]resource.Quantity {
+	li := make(map[corev1.ResourceName]resource.Quantity)
+	li[corev1.ResourceCPU] = resource.MustParse(cfg.Config.GlobalCPUResourceLimit)
+	li[corev1.ResourceMemory] = resource.MustParse(cfg.Config.GlobalMemoryResourceLimit)
```
ccremer, line 112: I see no advantage in abbreviating `limits` to `li`, it's less readable IMO (clean code).

ccremer (changes requested):

job/job.go
susana-garcia: as far as I understood, we'll always have default: https://github.com/vshn/k8up/issues/143

ccremer: Hm, I wouldn't say so, but that's not entirely clear either :)
If we start defining default limits, this could break many existing backups out there. I would argue that "make it possible to define limits" doesn't mean we have to define defaults.
So I really think we should leave defaults empty in K8up v1alpha1, but we can revise those in K8up v1alpha2 (= v2 of K8up).
I'll clear up the story

ccremer: Done, is that more clear?

config/samples/k8up_v1alpha1_check.yaml
```diff
@@ -3,6 +3,10 @@ kind: Check
 metadata:
   name: check-test
 spec:
+  resources:
+    requests:
+        memory: "64Mi"
+        cpu: "250m"
   backend:
```
ccremer, line 10: Can you also provide an example in the Schedule yaml with the template?

handler/schedule.go
```diff
@@ -71,34 +73,39 @@ func (s *ScheduleHandler) createJobList() scheduler.JobList {
...
 	if s.schedule.Spec.Archive != nil {
+		s.schedule.Spec.Archive.Resources = s.mergeResourcesWithDefaults(s.schedule.Spec.Archive.Resources)
 		jobList.Jobs = append(jobList.Jobs, scheduler.Job{
 			Type:     scheduler.ArchiveType,
 			Schedule: s.schedule.Spec.Archive.Schedule,
 			Object:   s.schedule.Spec.Archive.ArchiveSpec,
 		})
 	}
 	if s.schedule.Spec.Backup != nil {
+		s.schedule.Spec.Backup.Resources = s.mergeResourcesWithDefaults(s.schedule.Spec.Backup.Resources)
```
ccremer, line 84: these probably won't be applied correctly, need to find a different structure

ccremer:
Thanks @susana-garcia for the work!

susana-garcia:
@ccremer my pleasure 👍

PR #180, opened 2020-12-03
Docs: fix typo readme

susana-garcia:


Kidswiss (approved):

ccremer (approved):
Could you reword your commit with a `Docs: ` prefix? That way it won't be visible in the auto-generated changelog. I don't think this is worth mentioning ;)

Kidswiss:
@ccremer Are there some docs about the automatic changelog generator? Like, are there other prefixes that control its behaviour?

ccremer:
@Kidswiss our filters are here: https://github.com/vshn/k8up/blob/master/.goreleaser.yml#L27
In general, it's documented here: https://goreleaser.com/customization/release/#customize-the-changelog

Kidswiss:
Ah so that's also goreleaser magic, nice! Really need to read into that thing. Thanks!

susana-garcia:
@Kidswiss  that makes two of us :)

PR #186, opened 2020-12-07
Add new smart schedules

### LINKED ISSUE #116: Support "auto" schedules
## Summary

**As** K8up user\
**I want** to specify smart schedules\
**So that** I can let K8up figure out optimal schedules for optimal resource usage

## Context

The scheduler should be able accept "auto" or "smart" schedules.

These can be like for crons:
* daily
* weekly
* monthly
* etc.

They should behave in such a way that the jobs will be run at least once during that defined time. The old cron syntax should still be supported for use cases where specific time is necessary.

The idea for the auto schedules would be to be triggered at any time in the given frame. For example daily should mean that the job should run at least once every day sometime between 00:00 and 23:59. When exactly should be determined by the operator.

This feature is intended mostly for jobs that need exclusive access to the backup repository like prune and check. They don't have any impact on the applications and can thus run whenever no backups are running.

For prune and check jobs the operator has to figure out the best time between backups to a repository when the jobs can run. One idea could be that the prune could be triggered right after all backups have finished. That would eliminate the need for a separate prune schedule completely.

Additional from #118:

Also the cron library used by k8up has some predefined schedules like `@daily` or `@weekly` so for the auto schedules we'd have to define them without an @ as not to break those pre-defined schedules. https://pkg.go.dev/github.com/robfig/cron#hdr-Predefined_schedules

And finally, the cron library supports intervals which could come in very handy for this feature.

## Out of Scope

* Implement a smart scheduling strategy with evenly distributed schedules (as discussed in this issue, this poses more problems than it may solve)

## Further links

* [Cron Standard](https://en.wikipedia.org/wiki/Cron)
* #118 
* [Intervals](https://pkg.go.dev/github.com/robfig/cron#hdr-Intervals)

## Acceptance criteria

*Given* a schedulable K8up object\
*When* a [standardized cron syntax](https://en.wikipedia.org/wiki/Cron) is specified\
*Then* schedule the resource at the specified times
(this keeps existing functionality)

*Given* a schedulable K8up object\
*When* a non-standardized predefined cron syntax is specified\
*Then* schedule the resource at randomized times within the given timeframe, with a stable randomization seed
(e.g. when `@hourly-randomized` this could result in `<random number between 0 and 59> * * * *` schedule, contrast to predefined `0 * * * *`, which would defeat the purpose of smart scheduling)

*Given* a schedulable K8up object\
*When* a non-standardized predefined cron syntax with `@every` is specified\
*Then* schedule the resource at randomized times with the given interval
(e.g. when `@every 2h` this could result in a schedule that runs every 2h with a random start time)

*Given* a schedulable K8up object\
*When* any sort of non-standardized predefined cron syntax is specified\
*Then* store the resulting schedule in the status field of the object
(to make the actual schedule transparent for the user)

## Implementation Ideas

* Examples of non-standard cron syntax:
  * `@hourly` -> same as https://pkg.go.dev/github.com/robfig/cron#hdr-Predefined_schedules
  * `@hourly-randomized` -> this will trigger randomization of the schedule and then generate the actual, standard cron syntax
* stable random seed could be generated from the Name + Namespace combination, create a unique number out of it (checksum?). For the minute, calculate this seed modulo 60, for hours modulo 24 (in case of `@daily-randomized`) and fill that until we get a stable, random schedule like `23 5 * * *`

ccremer:
Implements new schedules:
* `@hourly-random`
* `@daily-random`
* `@weekly-random`
* `@monthly-random`
* `@yearly-random` or `@annually-random`

Special case: Monthly. The `@monthly-random` will produce a schedule with a day-of-month between 1 and 27, even though some months have 30 and 31 days. This is to prevent schedules being skipped if the current month has fewer days than specified in the day-of-month field in the schedule.

Implemented status field for random schedules, e.g.
```
metadata:
  ...
status:
  effectiveSchedules:
    archive: 0 * * * *
    check: 44 * * * *
```
Once the status field contains a schedule for a specific job type, it will use that and leave the `.spec.[type].schedule` as-is, so that the schedule like `@hourly-random` is not lost/overwritten in the spec with the generated schedule. If the generated schedule is missing in the status, it will be created and status updated. That means users can potentially adjust the generated schedule if that is not a good randomized schedule for some reason. It skips non-random schedules (e.g. `*/2 * * * *`) from putting them into the status. However, if the user provides another non-random schedule in the status still, that one will be used.

Closes #116


ccremer:
Who wants/can review?

Missing currently is e2e test. Should we create one or add one later? I have added unit tests for the generator, and done some manual tests in the kind-cluster regarding the status fields. Is that sufficient?

handler/schedule_test.go
```diff
@@ -150,3 +154,131 @@ func TestScheduleHandler_mergeResourcesWithDefaults(t *testing.T) {
...
+	}
+}
+
+func TestScheduleHandler_randomizeSchedule_VerifyCronSyntax(t *testing.T) {
+	s := ScheduleHandler{
+		Config: job.Config{Log: zap.New(zap.UseDevMode(true))},
+	}
+	formatString := "%d is not between %d and %d"
+	arr := []int{0, 0, 0, 0, 0}
+	for i := 0; i < 200; i++ {
```
corvus-ch, line 259: Place the `@yearly` and `@weekly` loop into two sub tests functions. It makes test failures easier to read and helps to narrow down what failed.

corvus-ch (changes requested):

corvus-ch:
> Missing currently is e2e test. Should we create one or add one later?

Giving the unit test coverage, I do not see the need for e2e tests.

config/samples/k8up_v1alpha1_schedule.yaml
```diff
@@ -52,7 +52,7 @@ spec:
       requests:
         memory: "64Mi"
         cpu: "250m"
-    schedule: '*/5 * * * *'
+    schedule: '@hourly-random'
```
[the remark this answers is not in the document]
srueg: The problem is, that `@hourly` is an allowed syntax which results in `0 * * * *`
https://crontab.guru/#@hourly

handler/schedule.go
```diff
@@ -56,16 +65,19 @@ func (s *ScheduleHandler) Handle() error {
...
-		if err != nil {
-			return fmt.Errorf("error while adding finalizer: %w", err)
-		}
+	if !controllerutil.ContainsFinalizer(s.schedule, scheduleFinalizerName) {
+		controllerutil.AddFinalizer(s.schedule, scheduleFinalizerName)
+		s.requireSpecUpdate = true
 	}
 
+	if s.requireSpecUpdate {
+		return s.updateSchedule()
```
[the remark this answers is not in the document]
srueg: As far as I remember, the `.Update()` call will update the local object with the latest version from the API server. This results in overwriting our local `.Status`, so it won't be the one we set anymore.
To solve this we don't update the status and wait for the next reconcile run, in which no update to spec should be necessary and we update the status.

handler/schedule.go
```diff
@@ -125,3 +137,65 @@ func (s *ScheduleHandler) mergeResourcesWithDefaults(resources corev1.ResourceRe
...
+	newSchedule := s.randomizeSchedule(s.schedule.Namespace+"/"+s.schedule.Name+"@"+string(jobType), originalSchedule)
+	s.schedule.Status.GeneratedSchedules[jobType] = newSchedule
+	s.requireStatusUpdate = true
+	return newSchedule
+}
+
+// randomizeSchedule randomizes the given originalSchedule with a seed. The originalSchedule has to be one of the supported
+// '@x-random' predefined schedules, otherwise it returns originalSchedule unmodified.
+func (s *ScheduleHandler) randomizeSchedule(seed, originalSchedule string) (schedule string) {
+	hash := sha1.New()
```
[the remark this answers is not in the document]
srueg: Using a seed helps to make sure jobs of different namespaces are distributed across time:
```
s.randomizeSchedule(s.schedule.Namespace+"/"+s.schedule.Name+"@"+string(jobType), originalSchedule)
```

Using `math.Rand` with time as a seed would mean different go routines which are running at the same time would generate the same schedule.

So I guess we could go with `math.Rand` and still provide a seed based on namespace/job name/job type 👍

handler/schedule.go
```diff
@@ -125,3 +137,65 @@ func (s *ScheduleHandler) mergeResourcesWithDefaults(resources corev1.ResourceRe
...
+	hash.Write([]byte(seed))
+	sum := hex.EncodeToString(hash.Sum(nil))
+	sumBase16, _ := new(big.Int).SetString(sum, 16)
+	defer s.Log.V(1).Info("Randomized schedule", "seed", seed, "sum", sum, "from_schedule", originalSchedule, "new_schedule", &schedule)
```
srueg, line 178: IMHO it would make more sense to store `schedule` in a variable and log & return it at the end of the function instead of using `defer` and named returns...

srueg (approved):

handler/schedule.go
corvus-ch: Just as a note: using defer for logging, allows to log at the latest possible point in time while not having to take into consideration all error cases in the rest of the function. Here, this feat is not required though.

handler/schedule.go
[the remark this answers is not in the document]
corvus-ch: Ok, I see the need for the fixed seed. As this function is attache do `ScheduleHandler`, we can get away with just passing `jobTyp`. The namespace and name are fields already available to the function. This makes it simpler for the caller.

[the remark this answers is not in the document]
ccremer: There are reasons for this, please read the discussions here: https://github.com/vshn/k8up/issues/116

config/samples/k8up_v1alpha1_schedule.yaml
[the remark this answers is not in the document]
ccremer: 1. As discussed in https://github.com/vshn/k8up/issues/116 and also with @Kidswiss, we want to keep the `-random` suffixes. Should we later implement another method to generate schedules, this would allow us to add more different kind of implementations, e.g. `@hourly-smart`, `@daily-nightly` etc.
2. As @srueg mentioned correctly, we want to keep the possibility to set the pre-defined schedules

handler/schedule_test.go
```diff
@@ -150,3 +154,131 @@ func TestScheduleHandler_mergeResourcesWithDefaults(t *testing.T) {
 		})
 	}
 }
+
+func TestScheduleHandler_randomizeSchedule(t *testing.T) {
+	name := "k8up-system/my-scheduled-backup@backup"
```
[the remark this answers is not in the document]
ccremer: This is what Goland generates. Do you have an example of table-driven tests that closer matches your idea?

handler/schedule.go
cimnine, line 178: I was wondering what benefit the information `seed` and `sum` provide, at least on the `Info` log-level. To me they seem to be purely an implementational detail and should not concern any outside observer.

handler/schedule.go
```diff
@@ -125,3 +137,65 @@ func (s *ScheduleHandler) mergeResourcesWithDefaults(resources corev1.ResourceRe
...
+		return fmt.Sprintf("%d %d * * *", minute, hour)
+	case ScheduleMonthlyRandom:
+		return fmt.Sprintf("%d %d %d * *", minute, hour, dayOfMonth.Add(dayOfMonth, big.NewInt(1)))
+	case ScheduleAnnuallyRandom:
+	case ScheduleYearlyRandom:
+		return fmt.Sprintf("%d %d %d %d *", minute, hour, dayOfMonth.Add(dayOfMonth, big.NewInt(1)), month.Add(month, big.NewInt(1)))
+	case ScheduleWeeklyRandom:
+		return fmt.Sprintf("%d %d * * %d", minute, hour, weekday)
+	}
+	return originalSchedule
```
cimnine, line 200: A switch without a `default` triggers my internal alert-bells. I therefore suggest:

```suggestion
	case ScheduleWeeklyRandom:
		return fmt.Sprintf("%d %d * * %d", minute, hour, weekday)
	default:
		return originalSchedule
	}
```

handler/schedule.go
```diff
@@ -125,3 +137,65 @@ func (s *ScheduleHandler) mergeResourcesWithDefaults(resources corev1.ResourceRe
...
+}
+
+func (s *ScheduleHandler) getOrGenerateSchedule(jobType k8upv1alpha1.JobType, originalSchedule string) string {
+	if s.schedule.Status.GeneratedSchedules == nil {
+		s.schedule.Status.GeneratedSchedules = make(map[k8upv1alpha1.JobType]string)
+	}
+	if schedule, found := s.schedule.Status.GeneratedSchedules[jobType]; found {
+		return schedule
+	}
+	newSchedule := s.randomizeSchedule(s.schedule.Namespace+"/"+s.schedule.Name+"@"+string(jobType), originalSchedule)
```
cimnine, line 165: I believe that `randomizeSchedule` should only be called when there is actually a respective schedule present. Or it should be the first step of `randomizeSchedule` to check if the schedule starts with an `@` and return otherwise.

If the latter of the two suggestions would be implemented – and in accordance with the previous suggestion of renaming `generatedSchdule` to `effectiveSchedule` –, then I believe a better way to name this function would be `calculateEffectiveSchedule`.

handler/schedule.go
cimnine: Also, if a check would be implemented that prevents the whole `randomizeSchedule` function to be executed when there is no need for it (i.e. the originalSchedule does not start with an `@`), then the default case should actually log and return an error. Something along the lines of `fmt.Errorf("'%s' is not a recognised schedule", originalSchedule)`.

handler/schedule.go
cimnine, line 178: I suggest to refactor this part into it's own function, e.g. `initializeSeed` or similar. This makes is clear what purpose that part of the code has.

cimnine (approved):
While reviewing this PR I was missing updates to the documentation that explain these new schedules and their purpose/impact and the restrictions/limitations (e.g. that `@monthly-random` only schedules between the days 1 to 27).

handler/schedule.go
ccremer: `.V(1).Info` is a debug message (verbosity level 1 is debug, 0 is informational). so, it should only be printed on debug level. Granted, currently debug is default and not configurable yet

handler/schedule.go
ccremer: The compiler complained about missing return statement after the switch clause even with a return statement in the default clause. So, with or without the default clause, the effect is the same so I left it out on purpose since it's redundant.

handler/schedule.go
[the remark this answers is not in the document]
ccremer: `rand.NewSource(seed int64)` expects an integer as seed. Any alternatives to generating a number from a string other than Checksums? `big.Int` was chosen because checksum functions return number in hexadecimal strings. Besides, there won't be much calculations efforts: It's only calculated when a new schedule object was created, or when existing schedules have the schedules updated with a randomizer schedule but the status has not yet been updated. Once the status field is written, there will be no random calculation anymore. IMO this should be acceptable (in single cases it's not even reliably measurable)

handler/schedule.go
ccremer: the only reason why that funtion needs the ScheduleHandler was because of a debug message... Now I've somewhat restructured this so that logging happens in the caller (business logic)

ccremer: I have restructured the methods a bit. now the randomizer doesn't need to log anything while the caller can still log the seed (which might be somewhat useful for debugging)

handler/schedule.go
ccremer: I combined your and Simon's feedback. Better?

handler/randomizer.go
```diff
@@ -0,0 +1,58 @@
...
+		effectiveSchedule = fmt.Sprintf("%d %d * * *", minute, hour)
+	case ScheduleMonthlyRandom:
+		effectiveSchedule = fmt.Sprintf("%d %d %d * *", minute, hour, dayOfMonth.Add(dayOfMonth, big.NewInt(1)))
+	case ScheduleAnnuallyRandom:
+	case ScheduleYearlyRandom:
+		effectiveSchedule = fmt.Sprintf("%d %d %d %d *", minute, hour, dayOfMonth.Add(dayOfMonth, big.NewInt(1)), month.Add(month, big.NewInt(1)))
+	case ScheduleWeeklyRandom:
+		effectiveSchedule = fmt.Sprintf("%d %d * * %d", minute, hour, weekday)
+	default:
+		return originalSchedule, errors.New("unrecognized random schedule: " + originalSchedule)
```
cimnine, line 47:
```suggestion
		return originalSchedule, fmt.Errorf("unrecognized random schedule: '%v'", originalSchedule)
```

I've learned in the past that it is good habit to enclose string values with ticks or quotes. It makes it easy to identify problems with misplaced whitespaces. I'm not certain if that even could apply here, but it doesn't hurt either IMO.

handler/randomizer.go
```diff
@@ -0,0 +1,58 @@
+package handler
+
+import (
+	"crypto/sha1"
+	"encoding/hex"
+	"errors"
```
cimnine, line 6:
```suggestion
```

handler/schedule.go
```diff
@@ -8,7 +8,10 @@ import (
 	"github.com/vshn/k8up/job"
 	"github.com/vshn/k8up/scheduler"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"strings"
```
cimnine, line 14: Not sure about the code conventions for this projects, but I usually ensure imports from the standard lib are at the top and that they are sorted alphabetically within their respective block.

handler/schedule.go
```diff
@@ -125,3 +129,43 @@ func (s *ScheduleHandler) mergeResourcesWithDefaults(resources corev1.ResourceRe
...
+	if s.schedule.Status.EffectiveSchedules == nil {
+		s.schedule.Status.EffectiveSchedules = make(map[k8upv1alpha1.JobType]string)
+	}
+	if existingSchedule, found := s.schedule.Status.EffectiveSchedules[jobType]; found {
+		return existingSchedule
+	}
+	if !strings.HasSuffix(originalSchedule, "-random") {
+		return originalSchedule
+	}
+	schedule := originalSchedule
```
cimnine, line 160: Isn't this redundant? `schedule` will be overwritten by the call to `randomizeSchedule` anyway without the variable ever being read before.

handler/schedule.go
```diff
@@ -125,3 +129,43 @@ func (s *ScheduleHandler) mergeResourcesWithDefaults(resources corev1.ResourceRe
...
+	seed := s.schedule.Namespace + "/" + s.schedule.Name + "@" + string(jobType)
+	schedule, err := randomizeSchedule(seed, originalSchedule)
```
cimnine, line 162: I argue that it's not the responsibility of the 'businessLogic' to come up with an effective seed. Therefore I suggest to create a function `createSeed(s)` that returns a new seed and place it in `randomizer.go`.

handler/schedule.go
```diff
@@ -125,3 +129,43 @@ func (s *ScheduleHandler) mergeResourcesWithDefaults(resources corev1.ResourceRe
...
+		s.schedule.Status.EffectiveSchedules = make(map[k8upv1alpha1.JobType]string)
+	}
+	if existingSchedule, found := s.schedule.Status.EffectiveSchedules[jobType]; found {
+		return existingSchedule
+	}
+	if !strings.HasSuffix(originalSchedule, "-random") {
+		return originalSchedule
+	}
+	schedule := originalSchedule
+	seed := s.schedule.Namespace + "/" + s.schedule.Name + "@" + string(jobType)
```
cimnine, line 161: PS: Is it intentional that the seed is deterministic? I.e. for the same namespace, schedule name and jobType it will end up with the same 'random' schedule time over and over again – or did I get this wrong?

handler/schedule.go
```diff
@@ -125,3 +129,43 @@ func (s *ScheduleHandler) mergeResourcesWithDefaults(resources corev1.ResourceRe
...
+	schedule, err := randomizeSchedule(seed, originalSchedule)
+	if err != nil {
+		s.Log.Info("Could not randomize schedule, ignoring and try original schedule", "seed", seed, "error", err.Error())
+	} else {
+		s.Log.V(1).Info("Randomized schedule", "seed", seed, "from_schedule", originalSchedule, "effective_schedule", schedule)
```
cimnine, line 166: I'm wondering what value the `seed` information has to an outside observer. I would say it's more confusing that helping when debugging, because it's an internal information, and therefore not add it to the log statements (here and the one above).

cimnine (approved):
I like the refactored version. It's a great improvement. Please read my comments as suggestions to improve it even further.

handler/schedule.go
ccremer: yes, that's intentional, requested in #116

handler/schedule.go
ccremer: I would argue that I need internal information when debugging, otherwise debugging would be fruitless, right?
Given that the stable randomization with seed is a requested feature, and the seed is relevant for the randomization, I see no harm when logging this. Maybe someone finds an issue with randomized schedules ("they are all the same" etc), then the seed becomes very much relevant.
For the Info log above, I agree though

handler/schedule.go
ccremer: there's no convention in this project. Whatever the IDE creates (in my case Goland), which would be accepted by go fmt, is fine.
would that be configurable in editorconfig somehow?

handler/randomizer.go
```diff
@@ -0,0 +1,61 @@
...
+func (s *ScheduleHandler) createSeed(schedule *k8upv1alpha1.Schedule, jobType k8upv1alpha1.JobType) string {
+	return schedule.Namespace + "/" + schedule.Name + "@" + string(jobType)
+}
+
+// randomizeSchedule randomizes the given originalSchedule with a seed. The originalSchedule has to be one of the supported
+// '@x-random' predefined schedules, otherwise it returns an error with the original schedule unmodified.
+func randomizeSchedule(seed, originalSchedule string) (string, error) {
+	checksum := calculateChecksumFromSeed(seed)
+
+	minute := new(big.Int).Set(checksum).Mod(checksum, big.NewInt(60))
```
corvus-ch, line 29:
```suggestion
	minute := new(big.Int).Mod(checksum, big.NewInt(60))
```
The call to `Set` does effect the end result. Applies to all other 

It might be worth to extract that into a function.

handler/randomizer.go
```diff
@@ -0,0 +1,61 @@
...
+	hour := new(big.Int).Set(checksum).Mod(checksum, big.NewInt(24))
```
corvus-ch, line 30: Do not calculate values not needed later on. Depending on the branch the switch statement takes, this and the following variables are not needed in all cases.

handler/randomizer_test.go
```diff
@@ -0,0 +1,80 @@
...
+			schedule, err := randomizeSchedule(seed, "@yearly-random")
+			assert.NoError(t, err)
+			fields := strings.Split(schedule, " ")
+			for j, f := range fields {
+				number, _ := strconv.Atoi(f)
+				arr[j] = number
+			}
+			assert.InDelta(t, 0, arr[0], 59.0, formatString, arr[0], 0, 59)
+			assert.InDelta(t, 0, arr[1], 59.0, formatString, arr[1], 0, 59)
+			assert.InDelta(t, 1, arr[2], 26.0, formatString, arr[2], 1, 27)
```
corvus-ch, line 25: Confusion of including or not. The above tests handles the upper limit including while this does not. The lover limit is including in all assertions.

handler/schedule.go
corvus-ch: I prefer also to have the suggested order. Goland does this if the standard imports and the other imports are separated by an empty line. But in the end, what ever `go fmt` accepts.

corvus-ch (changes requested):
Much better. I had a deeper look into the usage of `big.Int`. The call to `Set` does not contribute to the end result. `Set` sets the value of big.Int. `Mod` also sets the value of big.Int after calculating the modulo of the given numbers. So calling `Mod` after `Set` will override what ever happen in `Set`.

The following is equal:

```go
bigInt := big.NewInt(1234567890)
fmt.Println(new(big.Int).Set(bigInt).Mod(bigInt, big.NewInt(60)))
fmt.Println(new(big.Int).Mod(bigInt, big.NewInt(60)))
```

handler/schedule.go
ccremer: I rarely have to tinker manually in the imports, I let the IDE sort and do stuff there. There are now no empty lines and the IDE still doesn't do any sort of ordering. I would like to avoid having to sort manually, I don't think it adds much value and makes maintenance cumbersome.

handler/randomizer.go
ccremer: minute and hour is used in all cases, and dayOfMonth in 2 out of 5. Considering there's a explanation comment I am fine with having dayOfMonth also calculated in all cases in order to avoid duplication.

handler/randomizer_test.go
ccremer: This has to do with the cron syntax and the date.
For the minutes, we have a range between 0 and 59 of possible values (think max value = 0 + 59)
For the hours, a range between 0 and 23
However, since dates start at 1, we have different delta:
For dayOfMonth:  start at 1 + delta of 26 (max value = 1 + 26 = 27)

Or do you mean something else?

ccremer:
@tobru I have added some documentation. Mind have a look?

tobru (approved):
Documentation approved, thank you :+1:

handler/randomizer_test.go
corvus-ch: Ah, this is what I was missing.

PR #189, opened 2020-12-08
Self-documenting Makefile

cimnine:
This PR applies a neat trick as described in https://www.freecodecamp.org/news/self-documenting-makefile/.
A new `make` target called `help` is introduced that lists all documented targets and their documentation comment.

The syntax is simple:

```Makefile
# Makefile

target: ## I'm the documentation for the target
    @echo I'm the target code
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,8 @@ KIND_KUBECTL_ARGS ?= --validate=true
 KIND_REGISTRY_NAME ?= kind-registry
 KIND_REGISTRY_PORT ?= 5000
 
+SETUP_E2E_TEST := testbin/.setup_e2e_test
+
 # Image URL to use all building/pushing image targets
 DOCKER_IMG ?= docker.io/vshn/k8up:$(IMG_TAG)
 QUAY_IMG ?= quay.io/vshn/k8up:$(IMG_TAG)
@@ -50,11 +52,11 @@ endif
 SHELL := /bin/bash
 
 KUSTOMIZE ?= go run sigs.k8s.io/kustomize/kustomize/v3
+KUSTOMIZE_BUILD_CRD ?= $(KUSTOMIZE) build $(CRD_ROOT_DIR)/$(CRD_SPEC_VERSION)
 
-all: build
+all: build ## Invokes the build target
 
-# Run tests
-test: fmt vet
+test: fmt vet ## Run tests
 	go test ./... -coverprofile cover.out
 
 # Run tests (see https://sdk.operatorframework.io/docs/building-operators/golang/references/envtest-setup)
@@ -63,57 +65,47 @@ ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
 $(TESTBIN_DIR):
 	mkdir -p $(TESTBIN_DIR)
 
-# Run integration tests with envtest
 # See https://storage.googleapis.com/kubebuilder-tools/ for list of supported K8s versions
 # No, there's no 1.18 support, so we're going for 1.19
-integration_test: export ENVTEST_K8S_VERSION = 1.19.2
+integration_test: export ENVTEST_K8S_VERSION = 1.19.2 ## Run integration tests with envtest
 integration_test: generate fmt vet $(TESTBIN_DIR)
 	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/master/hack/setup-envtest.sh
 	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); go test -tags=integration -v ./... -coverprofile cover.out
 
-# Build manager binary
-build: generate fmt vet
+build: generate fmt vet ## Build manager binary
 	$(build_cmd)
 
-dist: generate fmt vet
+dist: generate fmt vet ## Generates a release
 	goreleaser release --snapshot --rm-dist --skip-sign
 
-# Run against the configured Kubernetes cluster in ~/.kube/config
-run: fmt vet
+run: fmt vet ## Run against the configured Kubernetes cluster in ~/.kube/config
 	go run ./main.go
 
-# Install CRDs into a cluster
-install: generate
+install: generate ## Install CRDs into a cluster
 	$(KUSTOMIZE) build $(CRD_ROOT_DIR)/v1 | kubectl apply -f -
 
-# Uninstall CRDs from a cluster
-uninstall: generate
+uninstall: generate ## Uninstall CRDs from a cluster
 	$(KUSTOMIZE) build $(CRD_ROOT_DIR)/v1 | kubectl delete -f -
 
-# Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-deploy: generate
+deploy: generate ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
-# Generate manifests e.g. CRD, RBAC etc.
-generate:
+generate: ## Generate manifests e.g. CRD, RBAC etc.
 	@CRD_ROOT_DIR="$(CRD_ROOT_DIR)" go generate -tags=generate generate.go
 	@rm config/*.yaml
 
-# Generate CRD to file
-crd: generate
+crd: generate ## Generate CRD to file
 	$(KUSTOMIZE) build $(CRD_ROOT_DIR)/v1 > $(CRD_FILE)
 	$(KUSTOMIZE) build $(CRD_ROOT_DIR)/v1beta1 > $(CRD_FILE_LEGACY)
 
-# Run go fmt against code
-fmt:
+fmt: ## Run go fmt against code
 	go fmt ./...
 
-# Run go vet against code
-vet:
+vet: ## Run go vet against code
 	go vet ./...
 
-lint: fmt vet
+lint: fmt vet ## Invokes the fmt and vet targets
 	@echo 'Check for uncommitted changes ...'
 	git diff --exit-code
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -63,76 +62,62 @@ ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
...
+vet: ## Run go vet against code
 	go vet ./...
 
 lint: fmt vet
 	@echo 'Check for uncommitted changes ...'
 	git diff --exit-code
 
-# Build the binary without running generators
-$(BIN_FILENAME):
+$(BIN_FILENAME): ## Build the binary without running generators
```
ccremer, line 109: This is an "internal" target. Unfortunately `make help` does not expand this variable:
```
$ make help
$(BIN_FILENAME)      Build the binary without running generators
build                Build manager binary
bundle               Generate bundle manifests and metadata, t...
```
```suggestion
$(BIN_FILENAME):
```

ccremer (commented):
Neat!
* Could you propose some comments for the remaining make targets as well? clean, e2e etc
* Would be cool if it's also mentioned in the README

Makefile
```diff
@@ -168,10 +154,14 @@ $(KIND_BIN): $(TESTBIN_DIR)
...
 	$(KIND_BIN) delete cluster --name $(KIND_CLUSTER) || true
 	docker stop "$(KIND_REGISTRY_NAME)" || true
 	docker rm "$(KIND_REGISTRY_NAME)" || true
 	docker rmi "$(E2E_IMG)" || true
 	rm -r testbin/ dist/ bin/ cover.out $(BIN_FILENAME) || true
 	$(MAKE) -C e2e clean
+
+.PHONY: help
+help: ## Show this help
+	@egrep -h '\s##\s' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
```
ccremer, line 167: `egrep` is actually deprecated. Can you modify the command to work with `grep -E` instead? see https://man7.org/linux/man-pages/man1/egrep.1.html

Makefile
```diff
@@ -63,93 +62,80 @@ ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
...
-docs-serve:
+docs-serve: ## Locally run the docs server
 	$(antora_preview_cmd)
 
-install_bats:
+install_bats: ## Installs the bats util via NPM
 	$(MAKE) -C e2e install_bats
 
-e2e_test: docker-build
+e2e_test: docker-build ## Runs the e2e integration test
```
ccremer, line 133:
```suggestion
e2e_test: docker-build ## Runs the e2e test suite
```

Makefile
```diff
@@ -63,93 +62,80 @@ ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
...
 	docker push $(E2E_IMG)
 	$(MAKE) -C e2e run_bats -e KUBECONFIG=../$(KIND_KUBECONFIG)
 
 setup_e2e_test: export KUBECONFIG = $(KIND_KUBECONFIG)
-setup_e2e_test: $(KIND_BIN)
+setup_e2e_test: $(KIND_BIN) ## Installs kind and configures the kind-cluster
```
ccremer, line 138: A comment for `run_kind` seems to be missing

ccremer (changes requested):

Makefile
```diff
@@ -63,98 +62,85 @@ ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
...
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
 	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
 	operator-sdk bundle validate ./bundle
 
-docs-serve:
+docs-serve: ## Locally run the docs server
 	$(antora_preview_cmd)
 
-install_bats:
+install_bats: ## Installs the bats util via NPM
```
ccremer, line 130:
```suggestion
install_bats: ## Installs the bats util via NPM (required for e2e tests)
```

Makefile
```diff
@@ -63,98 +62,85 @@ ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
...
 	$(MAKE) -C e2e install_bats
 
-e2e_test: docker-build
+e2e_test: docker-build ## Runs the e2e test suite
```
ccremer, line 133:
```suggestion
e2e_test: docker-build ## Runs the e2e test suite (requires bats, see install_bats)
```

ccremer (approved):
last ones, then lgtm

I suggest to squash your commits to have a cleaner commit history

Makefile
cimnine: IMO sollte das Makefile ja selber merken, dass `install_bats` noch nicht gelaufen ist. Das habe ich nun so umgesetzt.

cimnine:
I've rebased onto master and squashed my commits.

Makefile
ccremer: I generally agree, but then it should also depend on `setup_e2e_test`, but both these targets take a bit of time. So to make local development easier (when running over and over again) I decided to separate those

ccremer: The most correct approach would be to make Make skip when kind and bats are set up. But I couldn't come up with an easy solution without overengineering, so I gave up and accepted a little imperfection :)

cimnine: A tactic to work around that in Makefiles is to create marker files. I.e. one would `touch testfiles/.setup_e2e_test` at the end of the respective target, which avoids the target from being run all over again when it ran already. Shall I add this?

ccremer: Let's try, but I would recommend `testbin/` dir instead, to reuse an existing one.

cimnine: That's what I meant 🙈

Makefile
```diff
@@ -63,100 +64,101 @@ ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
...
-	@kubectl config use-context kind-$(KIND_CLUSTER)
-	@$(KUSTOMIZE) build $(CRD_ROOT_DIR)/$(CRD_SPEC_VERSION) | kubectl apply $(KIND_KUBECTL_ARGS) -f -
-
 run_kind: export KUBECONFIG = $(KIND_KUBECONFIG)
-run_kind: setup_e2e_test
+run_kind: $(SETUP_E2E_TEST) ## Runs the operator in kind
 	go run ./main.go
 
+.PHONY: setup_e2e_test
+setup_e2e_test: force_setup_e2e_test $(SETUP_E2E_TEST) ## (Re-)run the e2e setup
```
ccremer, line 144: I have a alternative proposal:
Introduce a `clean_e2e_setup` target instead, that also removes the `k8up-system` namespace and CRD, but leaves the kind cluster intact? To re-run the test with a clean slate, run `make clean_e2e_setup e2e_test`, One can still run `make clean e2e_test` to rebuild the cluster from scratch.
```suggestion
setup_e2e_test: $(SETUP_E2E_TEST) ## Run the e2e setup
```

Makefile
```diff
@@ -167,11 +169,10 @@ $(KIND_BIN): $(TESTBIN_DIR)
...
-	docker rmi "$(E2E_IMG)" || true
-	rm -r testbin/ dist/ bin/ cover.out $(BIN_FILENAME) || true
-	$(MAKE) -C e2e clean
+$(KIND_KUBECONFIG): $(KIND_BIN)
+
+$(SETUP_E2E_TEST): export KUBECONFIG = $(KIND_KUBECONFIG)
+$(SETUP_E2E_TEST): $(KIND_BIN)
+	@kubectl config use-context kind-$(KIND_CLUSTER)
+	@$(KUSTOMIZE) build $(CRD_ROOT_DIR)/$(CRD_SPEC_VERSION) | kubectl apply $(KIND_KUBECTL_ARGS) -f -
+	touch $(SETUP_E2E_TEST)
```
ccremer, line 178:
```suggestion
	@touch $(SETUP_E2E_TEST)
```

Makefile
```diff
@@ -63,100 +64,101 @@ ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
...
 run_kind: export KUBECONFIG = $(KIND_KUBECONFIG)
-run_kind: setup_e2e_test
+run_kind: $(SETUP_E2E_TEST) ## Runs the operator in kind
 	go run ./main.go
 
+.PHONY: setup_e2e_test
+setup_e2e_test: force_setup_e2e_test $(SETUP_E2E_TEST) ## (Re-)run the e2e setup
+
+force_setup_e2e_test:
+	@rm $(SETUP_E2E_TEST) || true
```
ccremer, line 147:
```suggestion
clean_e2e_setup:
	@kubectl delete ns k8up-system --wait
	@$(KUSTOMIZE) build $(CRD_ROOT_DIR)/$(CRD_SPEC_VERSION) | kubectl delete $(KIND_KUBECTL_ARGS) -f -
	@rm $(SETUP_E2E_TEST) || true
```

Makefile
```diff
@@ -63,100 +65,105 @@ ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
...
+run_kind: $(SETUP_E2E_TEST) ## Runs the operator in kind
 	go run ./main.go
 
+.PHONY: setup_e2e_test
+setup_e2e_test: $(SETUP_E2E_TEST) ## Run the e2e setup
+
+.PHONY: clean_e2e_setup
+clean_e2e_setup: export KUBECONFIG = $(KIND_KUBECONFIG)
+clean_e2e_setup: ## Clean the e2e setup (e.g. to rerun the setup_e2e_test)
+	kubectl delete ns k8up-system --wait || true
```
ccremer, line 150: A better command, but `|| true` is still required, in case the kubeconfig is inexistent.
```suggestion
	kubectl delete ns k8up-system --ignore-not-found --force --grace-period=0 || true
```

ccremer (approved):
Though I'm starting to think whether it would make more sense to move the kind and e2e stuff into the Makefile in the `e2e` dir. I would still keep the high-level make targets as a the "interface" but similar to install_bats execute them in a different makefile.

I'm not sure what the best approach is to pass/expand parameters into another makefile. 

WDYT?

cimnine:
> Though I'm starting to think whether it would make more sense to move the kind and e2e stuff into the Makefile in the e2e dir.

I think this is a good idea, though I would like to complete the work on this PR for now. Would you be ok with me creating a new issue to reorganise the `Makefile`?

ccremer (approved):
fine by me

cimnine:
I've created the issue https://github.com/vshn/k8up/issues/198 to track the idea of extracting e2e related make targets to the relevant Makefile.

PR #191, opened 2020-12-08
Flatten K8upStatus spec

ccremer:
It's probably history, but it only adds unnecessary complexity today


job/job.go
```diff
@@ -42,7 +42,7 @@ type Config struct {
 type Object interface {
 	GetMetaObject() metav1.Object
 	GetRuntimeObject() runtime.Object
-	GetK8upStatus() *k8upv1alpha1.K8upStatus
+	GetK8upStatus() *k8upv1alpha1.Status
```
cimnine, line 45: I'm not sure about this, but shouldn't it be `GetStatus()` now?

```suggestion
	GetStatus() *k8upv1alpha1.Status
```

api/v1alpha1/status.go
```diff
@@ -0,0 +1,9 @@
+package v1alpha1
+
+// Status defines the observed state of a generic K8up job. It is used for the
+// operator to determine what to do.
+type Status struct {
```
cimnine, line 5: I'm just curious: Is there a reason why you've renamed `K8upStatus` to `Status`?

cimnine (approved):
No functionality change can be observed, this is purely a semantic change. Therefore I believe there are no implications on compatibility etc. I've added to minor remarks, which are rooted in my curiosity rather than a request for change.

api/v1alpha1/status.go
ccremer: Lazyness :) Kubebuilder takes the Type description when the Property is not documented for the CRD generation
```yaml
            description: Status defines the observed state of a generic K8up job. It is used for the operator to determine what to do.
```
If I left it K8upStatus, then it would take also the comment ("description: K8upStatus defines the observed s..."). Now we could leave the Comment mentioning "Status" only, but then any decent IDE/linting system complains about mismatching type and comment, since type comments should begin with the name of the Type AFAIK.

We could override the description when documenting the field like previously:
```go
	// Status this comments overrides the comment of the struct now.
	Status Status     `json:"status,omitempty"`
```
But then I'd have to document every status field again.

do you feel we should change that?

cimnine: No, it's fine by me :)

cimnine (approved):

PR #192, opened 2020-12-08
Add .dockerignore

cimnine:
Since I was running the e2e_test quite a bit now I suspected that the context which Docker sends to it's daemon could be reduced by declaring a `.dockerignore` file and the build speed thereby improved. This didn't work out (at least with [buildkit](https://github.com/moby/buildkit) enabled). Since I've done the work anyway, I may as well offer it as a PR.


tobru (approved):

ccremer:
The only file that the Dockerbuild needs, is the pre-built binary `k8up`. Would it be possible to reverse the dockerignore and make a whitelist instead? e.g. ignore everything except the binary?

PR #200, opened 2020-12-10
Disable LEADER_ELECTION by default for `make run run_kind`

cimnine:
Without further configuration, k8up expects a k8s cluster to have a leader election namespace.

Just executing `make run` oder `make run_kind` therefore leads to the following error message:

    0000-00-00T00:00:00.000+0100	ERROR	setup	unable to start K8up operator	{"error": "unable to find leader election namespace: not running in-cluster, please specify LeaderElectionNamespace"}

This PR adjusts the Makefile in a way such that when invoking the mentioned make targets no leader election namespace is required.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,8 @@ KIND_REGISTRY_PORT ?= 5000
 
 SETUP_E2E_TEST := testbin/.setup_e2e_test
 
+ENABLE_LEADER_ELECTION ?= false
+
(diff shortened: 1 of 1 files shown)


ccremer (approved):
I suggest to add "for local development" in your commit message. Otherwise it shows up in the changelog and could lead to confusion if it's for production as well

PR #202, opened 2020-12-11
Fix Null Pointer Dereference

cimnine:
There is an error in the code which leads to a null pointer dereference panic during runtime.
`s.schedule.Spec.Archive` is referenced, whereas `s.schedule.Spec.Check` is tested for non-nullness. This seems wrong and will be rectified in this PR.


tobru (approved):
Absolutely makes sense, thanks for fixing

PR #204, opened 2020-12-14
Add Conditions to the Status information

cimnine:
The goal of this PR is to provide more information about various error and success-conditions to the average K8up user. Such users are mainly interacting with K8up through the relevant custom resources. Therefore, when querying the status of these resources (e.g. `kubectl describe backups/X`), Conditions shall provide additional information about the resource's individual status.

Closes https://github.com/vshn/k8up/issues/138


executor/archive.go
```diff
@@ -67,16 +73,12 @@ func (a *ArchiveExecutor) startArchive(job *batchv1.Job, archive *k8upv1alpha1.A
 	if err != nil {
 		if !apierrors.IsAlreadyExists(err) {
 			a.Log.Error(err, "could not create job")
+			a.setConditionFalse(ConditionJobCreated, "could not create job")
```
corvus-ch, line 76: Please include the error into the condition message.

Having an error log followed by a condition is repeating pattern. It is probably worth the create a function to simplify this. This would also address the issue of having the message string twice. Maybe move the log into `setConditionFalse()`.

If you omit the abstraction, at least eliminate the duplication of the message strings (local variable/file level variable/constant).

executor/backup.go
```diff
@@ -63,18 +75,26 @@ func (b *BackupExecutor) Execute() error {
...
+		ConditionCleanupSucceeded,
+	)
+
 	err := b.createServiceAccountAndBinding()
 	if err != nil {
 		return err
 	}
 
-	job, err := job.GetGenericJob(b.Obj, b.Config)
+	genericJob, err := job.GetGenericJob(b.Obj, b.Config)
```
corvus-ch, line 91: Try to prevent changes irrelevant to the task at hand. I is a good habit that have saved me a lot of potential merge conflicts.

executor/restore.go
```diff
@@ -58,27 +62,19 @@ func (r *RestoreExecutor) startRestore(restore *k8upv1alpha1.Restore) {
...
-	}
-}
-
-func (r *RestoreExecutor) setJobStatusStarted() error {
-	original := r.Obj
-	new := r.Obj.GetRuntimeObject().DeepCopyObject().(*k8upv1alpha1.Restore)
-	new.GetStatus().Started = true
-
-	return r.Client.Status().Patch(r.CTX, new, client.MergeFrom(original.GetRuntimeObject()))
+	r.setStartedTrue(ConditionJobCreated)
```
corvus-ch, line 77: It would be nice to indicate the id of the created resource. This makes easier to look at that job (no need to do `kubectl get job` and find the matching one from the list).

Apply to other locations where this idea applies.

executor/restore.go
```diff
@@ -137,15 +137,13 @@ func (r *RestoreExecutor) buildRestoreObject(restore *k8upv1alpha1.Restore) (*ba
...
 		args = append(args, "-restoreType", "folder")
-		methodDefined = true
-	}
-
-	if !methodDefined && restore.Spec.RestoreMethod.S3 != nil {
+	case restore.Spec.RestoreMethod.S3 != nil:
 		args = append(args, "-restoreType", "s3")
-		methodDefined = true
+	default:
+		r.Log.Error(err, "undefined restore method (-restoreType)", "namespace", r.Obj.GetMetaObject().GetNamespace())
```
corvus-ch, line 146: Should probably also be a condition.

executor/worker.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/vshn/k8up/observer"
 	"github.com/vshn/k8up/queue"
+
```
corvus-ch, line 10: File changed only by white spaces. Try not to do so. It keeps the risk of merge conflicts low. If you do want to fix formatting in general, do so in a PR on its own where only formatting fixes are done.

Comment by a burnt developer.

corvus-ch (changes requested):
I miss a condition for when the job finished. This might not be that easy to achieve though as this would require to what jobs and have a reconciliation loop for when they change. Maybe create a feature request for this.

executor/archive.go
cimnine: I've changed the signature of setConditonFalse() slightly to allow passing the actual errors. _Logging_ and _updating the Conditions_ are two different concerns in my opinion and that's why I would not want to mix them in one method.

executor/backup.go
cimnine: Thanks for the advice. I try to adhere to [the Boyscout Rule](https://deviq.com/boy-scout-rule/): _Leave your code better than you found it._ Here, `job` was colliding with an import named `job`. Since I've discovered it, I've changed it.

executor/worker.go
cimnine: I was changing other code in this file and therefore also performed minor reformatting. Later, I refactored the code and my changes were not relevant any longer and were removed. The formatting stayed. If they were to become a merge conflict, then I can still decide to replace them with the incoming change.

executor/restore.go
cimnine: What would be the condition in your opinion? Since 

https://github.com/vshn/k8up/pull/204/files/3ac6b29a55917977ea2a16f3b6c2c0d741e52747#diff-7d65ba23b900330e9af0767e831c1a0ce44d9ceef8ec848084ac39c6d390e044R65

captures an error in here as a Condition anyway, I rather suggest to treat this as an error, i.e. to `return nil, err` (where err is a new error created with `fmt.Errorf` which replaces `r.Log.Error`.)

executor/restore.go
cimnine: Good idea. Implemented ✔️

cimnine:
> I miss a condition for when the job finished. This might not be that easy to achieve though as this would require to what jobs and have a reconciliation loop for when they change. Maybe create a feature request for this.

I'll quickly look into it tomorrow, but it'll probably lead to a feature request.

executor/restore.go
corvus-ch: Ah, I see. Yes, returning an error is the better option then.

ccremer:
> I miss a condition for when the job finished. This might not be that easy to achieve though as this would require to what jobs and have a reconciliation loop for when they change. Maybe create a feature request for this.

Not sure if I understand it correctly. We already have a reconciler for batch.jobs: https://github.com/vshn/k8up/blob/master/controllers/job_controller.go#L50, so we are watching those.

prebackup/prebackup.go
```diff
@@ -34,14 +36,38 @@ func NewPrebackup(config job.Config) *PreBackup {
...
+	ConditionPreBackupPodsRemoved status.ConditionType = "PreBackupPodsRemoved"
+)
+
 // Start will start the defined pods as deployments.
 func (p *PreBackup) Start() error {
 
+	p.SetConditionsUnknown(
+		ConditionPreBackupPodsAvailable,
+		ConditionPreBackupPodsReady,
+	)
```
ccremer, line 56: (applies to other job types as well)
I would advise to only set conditions once they're actually needed for the first time. Setting all possible conditions to unknown right at the start may confuse the users reading them (it's so easy to read a condition and assume they're true by overlooking the "unknown" property).

prebackup/prebackup.go
```diff
@@ -61,10 +87,6 @@ func (p *PreBackup) getPodTemplates() (*k8upv1alpha1.PreBackupPodList, error) {
 func (p *PreBackup) generateDeployments(templates *k8upv1alpha1.PreBackupPodList) []appsv1.Deployment {
 	deployments := make([]appsv1.Deployment, 0)
 
-	if len(templates.Items) == 0 {
-		return deployments
-	}
```
ccremer, line 66: By removing this, empty list of pre backup pods leads to the condition `ConditionPreBackupPodsReady`. I doubt that's intended?

prebackup/prebackup.go
```diff
@@ -175,39 +205,42 @@ func (p *PreBackup) waitForReady(watcher watch.Interface) error {
...
 		default:
-			p.Log.Info("unexpected event during deployment wathc ", "name", deployment.GetName(), "event type", event.Type, "namespace", deployment.GetNamespace())
+			p.Log.Info("unexpected event during deployment watch ", "name", deployment.Name, "event type", event.Type, "namespace", deployment.Namespace)
 		}
 	}
 
 	return nil
 }
 
+func deadlineExceededCondition(last *appsv1.DeploymentCondition) bool {
```
ccremer, line 235: to not confuse booleans with conditions
```suggestion
func isDeadlineExceeded(last *appsv1.DeploymentCondition) bool {
```

ccremer (changes requested):

corvus-ch:
> We already have a reconciler for batch.jobs: https://github.com/vshn/k8up/blob/master/controllers/job_controller.go#L50, so we are watching those.

I was not aware of that one. So then this should be simple, I guess.

executor/prune.go
```diff
@@ -61,31 +67,22 @@ func (p *PruneExecutor) Exclusive() bool {
...
 
-	if err := p.setJobStatusStarted(); err != nil {
-		p.Log.Error(err, "could not update prune status")
-	}
-}
-
-func (p *PruneExecutor) setJobStatusStarted() error {
-	original := p.Obj
-	new := p.Obj.GetRuntimeObject().DeepCopyObject().(*k8upv1alpha1.Prune)
-	new.GetStatus().Started = true
```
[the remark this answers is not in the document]
cimnine: This is included in `p.SetStartet()`, [see line 85](https://github.com/vshn/k8up/pull/204/files#diff-254ecda41174d201e3eb2e3039f1936b1e03b24be0a4ca21633d759cadcaabc3R85).

`status.Started` is now uses a `Client.Status().Patch()` and [is unified across all `job.Config` types](https://github.com/vshn/k8up/pull/204/files#diff-4dd53f1e4705ab643eee58d7f40474c44a9e71f766698e80f55926974cec2946R142).

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/knadh/koanf v0.14.0
 	github.com/onsi/ginkgo v1.14.2
 	github.com/onsi/gomega v1.10.4
+	github.com/operator-framework/operator-lib v0.1.0
```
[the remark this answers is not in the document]
cimnine: I'm using the operator-lib here because the relevant Kubernetes implementation [is only available in 1.19+](https://github.com/kubernetes/apimachinery/blob/d4f471b82f0a17cda946aeba446770563f92114d/pkg/apis/meta/v1/types.go#L1368). I asked during one of the daily meetings what the minimal requirements for k8up are and I was told it's currently k8s 1.18. So I opted for the implementation provided by the operator-lib.

Since these types were remove in operator-framework 0.2.0 in favour of the upstream implementation this will surface eventually.

prebackup/prebackup.go
cimnine: The check has simply [moved up the call-stack](https://github.com/vshn/k8up/pull/204/files/baafb8008fcbcbaab26eae22ca47a0a31c34a777#diff-b539ff69a01fcdd227c6d02d84533cbc0efdfed55761119bd9c125fb21337bd2R64) so that the method is not even called.

cimnine:
> > We already have a reconciler for batch.jobs: https://github.com/vshn/k8up/blob/master/controllers/job_controller.go#L50, so we are watching those.
> 
> I was not aware of that one. So then this should be simple, I guess.

I don't see a straight-forward way to get back to the original Backup object, though. This is what I've come up with so far:

```go
// Handle extracts some information from the batchv1.job that make observations
// easier.
func (j *JobHandler) Handle() error {

	jobEvent := observer.Create

	if _, exists := j.job.GetLabels()[job.K8uplabel]; !exists {
		return nil
	}

	if j.job.Status.CompletionTime != nil {
		references := j.job.GetObjectMeta().GetOwnerReferences()
		if len(references) < 1 {
			return nil
		}

		reference := references[0]

		ownerBackupObj := &v1alpha1.Backup{}
		err := j.Client.Get(j.CTX, client.ObjectKey{Name: reference.Name, Namespace: j.job.Namespace}, ownerBackupObj)
		if err != nil {
			return err
		}
	}
```

- Is that how you'd solve it?
- Do I understand correctly that the OwnerReference could either be a Backup, Restore, Check, Prune, Snapshot, etc.? So I would have to account for all possible types that start jobs? (That seems error-prone to me.)

go.mod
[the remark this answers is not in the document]
ccremer: Yes, I pointed that out already. However with K8s 1.19 we won't be needing operator-lib anymore just for the status, K8s comes with its own helper methods, as mentioned.
That's why i asked for a comment stating that it's temporary until we managed to upgrade to 1.19

job/job.go
```diff
@@ -89,3 +91,62 @@ func GetGenericJob(obj Object, config Config) (*batchv1.Job, error) {
...
+
+	c.Obj.GetStatus().Started = true
+
+	c.Obj.GetStatus().Conditions.SetCondition(status.Condition{
+		Type:    trueCondition,
+		Status:  corev1.ConditionTrue,
+		Message: fmt.Sprintf(message, args...),
+	})
+
+	err := c.Client.Status().Patch(c.CTX, c.Obj.GetRuntimeObject(), patch)
```
ccremer, line 148: In another Operator we use `r.Client.Status().Update(..., ...)`. What is the difference to Patch()?

prebackup/prebackup.go
```diff
@@ -219,28 +248,33 @@ func (p *PreBackup) getLastDeploymentCondition(deployment *appsv1.Deployment) *a
 
 // Stop will remove the deployments.
 func (p *PreBackup) Stop() {
-
 	templates, err := p.getPodTemplates()
 	if err != nil {
 		p.Log.Error(err, "could not fetch podtemplates", "name", p.Obj.GetMetaObject().GetName(), "namespace", p.Obj.GetMetaObject().GetNamespace())
+		p.SetConditionFalse(ConditionPreBackupPodsRemoved, "could not fetch podtemplates: %v", err)
```
ccremer, line 254:
```suggestion
		p.SetConditionFalse(ConditionPreBackupPodsRemoved, "could not fetch pod templates: %v", err)
```

prebackup/prebackup.go
```diff
@@ -175,39 +201,42 @@ func (p *PreBackup) waitForReady(watcher watch.Interface) error {
...
+			p.Log.Info("waiting for command pod to get ready", "name", deployment.Name, "namespace", deployment.Namespace)
 
 		case watch.Error:
 
 			last := p.getLastDeploymentCondition(deployment)
 
 			if last != nil {
-				return fmt.Errorf("there was an error while starting pre backup pod %v: %v", deployment.GetName(), last.Message)
-
+				return fmt.Errorf("there was an error while starting pre backup pod '%v' in 'ns/%v': %v", deployment.Name, deployment.Namespace, last.Message)
```
ccremer, line 219: we should stick to common resource naming conventions (there are simillar other errors being thrown, I'm just mentioning here)
```suggestion
				return fmt.Errorf("there was an error while starting pre backup pod in deployment '%s/%s': %s", deployment.Namespace, deployment.Name, last.Message)
```

cimnine: I've changed this, because when I saw it in the operator I was like: _Cool, `k8up-test/pod-XXXX-xxxx`, I can just copy it to `kubectl logs k8up-test/pod-XXXX-xxxx`_. But no, that's not the way it works. So I'm not sure about the convention you mention either, because the main convention for a `/` in K8s in my opinion is `resourceKind/resourceName`, not `namespaceName/resourceName`.

How would you like me to proceed?

job/job.go
cimnine: As far as I understood, Patch will update just the properties that changed, whereas Update will replace the whole resource, i.e. update every property. That's a neat way to avoid the error about the resource not being up-to-date, because – at least for status updates – I don't really care if some other operator modified (parts of) the resource in the meantime. I just want to update that particular status field.

prebackup/prebackup.go
ccremer: I guess there are different kinds of convention, or no convention at all :) Most operators I saw so far use the `<namespace>/<resource>` notation to indicate the location of a resource. AFAIK the `<kind>/<resource>` is only used in the CLI. 
Neither approach is directly copy-pastable, as your current context might default to a different namespace than the one logged, so you'd need the `--namespace` flag too anyway.
So, choose your poison? I've just never seen the `ns/<namespace>` notation. and kubectl seems to agree with me:
```
$ k -n ns/default get pods
invalid namespace "ns/default": [may not contain '/']
```

cimnine: Like I said, I understood the `/` convention as `<kind>/<resource>`, so it'd be `kubectl get ns/default`, which works perfectly ;)

Nevertheless, I'll switch back to `<ns>/<resource>` if that's what other operators use.

go.mod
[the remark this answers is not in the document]
cimnine: You'll now find an improved note in `api/v1alpha1/status.go`.

cimnine:
Rebased onto master and stashed the commits.

ccremer (approved):
I played around with it locally, though I haven't set up local backup repository so they always failed. at least the "JobCreated" condition is correctly being applied to the prune job that I set up.

Looking good so far! Thanks for the works!

tobru (approved):
Documentation update LGTM

corvus-ch (approved):

PR #207, opened 2020-12-14
Build k8up in Docker

cimnine:
This PR ensures that the k8up binary is built for the correct Arch.
I.e. that the packeged-up binary is not accidentially a macOS binary
when the host is a macOS computer.

## Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,16 @@
-.*
-*
-!k8up
+# dirs
+
+.git
+.github
+docs
+e2e
+testbin
+
+# files
+.*ignore
+.editorconfig
+.git*
+*.png
+*.md
+Dockerfile
+renovate.json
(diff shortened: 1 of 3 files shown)


ccremer:
We are using Goreleaser as the build and release tool. Goreleaser expects a different kind of build where the binary is built externally and added via COPY. See here: https://goreleaser.com/customization/docker/.

Obviously both approaches have their (dis)advantages...

I have a counter proposal: We should adapt the Makefile so that it always builds the Linux binary for the docker image by setting/exporting the correct GOOS and GOARCH vars.

If we require multi-arch platform support (we currently don't AFAIK), we should investigate more into docker manifests, for which [Goreleaser has now also support for it](https://goreleaser.com/customization/docker_manifest/) but we'd need to adapt .goreleaser.yml accordingly.

PR #218, opened 2020-12-21
Add JobSucceeded Condition to Status fields

cimnine:
Based on #204

Implements and closes https://github.com/vshn/k8up/issues/212

## Checklist

- [x] Update tests.

### DIFF
--- a/observer/observer_test.go
+++ b/observer/observer_test.go
@@ -6,10 +6,11 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/vshn/k8up/api/v1alpha1"
 	batchv1 "k8s.io/api/batch/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/vshn/k8up/api/v1alpha1"
 )
 
 func TestObserver_IsConcurrentJobsLimitReached(t *testing.T) {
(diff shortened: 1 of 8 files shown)


cimnine:
I will leave this issue in draft state until #204 is merged.

observer/observer_test.go
```diff
@@ -56,3 +54,98 @@ func TestObserver_IsConcurrentJobsLimitReached(t *testing.T) {
...
+		eventType           EventType
+		registeredCallbacks int
+		expectedInvocations int
+	}{
+		{
+			name:                "succeeded status",
+			eventType:           Suceeded,
+			registeredCallbacks: 1,
+			expectedInvocations: 1,
+		},
```
ccremer, line 70: I propose to use the *Given_When_Then* syntax. It's easier to read in the Test spec what is expected or not than the comments in the test logic that are the same for every case.
and make the test cases using a map[string]struct instead of a struct slice, it looks a little bit nicer IMO. See https://github.com/vshn/k8up/blob/v1.0.0-rc1/handler/schedule_test.go#L163 for inspiration

cimnine: Thank you for this input.

I've reworked the tests and split them into two separate tests, because the original code was testing two different things anyway: Are Callbacks invoked per Status, and are all registered callbacks invoked.

These are now two tests and I believe it's now simpler to understand what they do.

observer/observer_test.go
```diff
@@ -56,3 +55,123 @@ func TestObserver_IsConcurrentJobsLimitReached(t *testing.T) {
...
+
+func TestObserver_AreOnlyExpectedCallbacksInvoked(t *testing.T) {
+	tests := map[string]struct {
+		givenEventType   EventType
+		expectInvocation bool
+	}{
+		"GivenStatusSucceeded": {
+			givenEventType:   Suceeded,
+			expectInvocation: true,
+		},
```
ccremer, line 115: I was referring to this construct:
```suggestion
		"GivenStatusSucceeded_WhenHandleEvent_ThenExpectInvocation": {
			givenEventType:   Suceeded,
			expectInvocation: true,
		},
```
It might look a bit verbose and tedious to copy-paste, but if a test fails on GH action or command line, that sometimes gives better indication which test failed and why. Often the assertion output can be confusing. Also for consistency reasons (we have other tests following this scheme where reasonablee)

cimnine: I've adjusted to these messages.

PR #219, opened 2020-12-21
Grant additional RBAC permissions to K8up

### LINKED ISSUE #213: [Bug] K8up doesn't have the RBAC permissions to create Roles, ServiceAccounts and Rolebindings
## Describe the bug

K8up doesn't have the RBAC permissions to create roles and rolebindings for backup pods.

## Additional context

The RBAC rules generated with kubebuilder do not take into account that K8up creates Roles and Bindings at runtime.

## Logs

```console
2020-12-18T16:10:04.942Z	ERROR	controllers.Backup	cannot create job	{"backup": "default/backupjob-1608307800", "repository": "s3:/", "error": "roles.rbac.authorization.k8s.io is forbidden: User \"system:serviceaccount:k8up-system:default\" cannot create resource \"roles\" in API group \"rbac.authorization.k8s.io\" in the namespace \"default\""}
github.com/go-logr/zapr.(*zapLogger).Error	
    /home/chris/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128
github.com/vshn/k8up/executor.(*QueueWorker).loopRepositoryJobs	
    /home/chris/Documents/source/vshn/k8up/executor/worker.go:70
github.com/vshn/k8up/executor.(*QueueWorker).executeQueue	
    /home/chris/Documents/source/vshn/k8up/executor/worker.go:41
```

## Expected behavior

* K8up is able to create Roles, ServiceAccount and Rolebinding at runtime
* `make generate` also creates RBAC rules so that it can be imported into the Helm chart as well.

## To Reproduce

1. Run e2e tests
```bash
make setup_e2e_test e2e_test
```
2. Investigate logs and backup objects
```bash
kubectl -n k8up-system logs deployment/k8up-operator
kubectl default describe backup (see the conditions)
```

- Image Version: master
- K8s Version: e.g. v1.18
- K8s Distribution: Kind

ccremer:
Since move to Operator SDK, some permissions have to be added that are
needed during runtime

Closes #213
 
## Checklist
Remove items that do not apply. For completed items, change [ ] to [x].
-->

- [ ] Update tests.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -69,6 +69,26 @@ rules:
 - apiGroups:
   - backup.appuio.ch
   resources:
+  - prebackuppods
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - backup.appuio.ch
+  resources:
+  - prebackuppods/status
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
+  - backup.appuio.ch
+  resources:
   - prunes
   verbs:
   - create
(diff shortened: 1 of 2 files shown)


cimnine (approved):
I did not spot any blocker, but my knowledge on K8s RBAC is limited.

corvus-ch (approved):
I can not tell if those are the correct RBAC rules but the change is done using the correct tooling. When it comes to the usage of `*` instead of explicit list the verbs, I have my doubts. However, I can not argue why not to do so.

ccremer:
> I can not tell if those are the correct RBAC rules but the change is done using the correct tooling. When it comes to the usage of `*` instead of explicit list the verbs, I have my doubts. However, I can not argue why not to do so.

I can understand your reservations, I was also surprised by the need of it. However, [apparently when K8up creates such runtime role](https://github.com/vshn/k8up/blob/e6543fb3e6466d48b91bb164220955a4cf327088/executor/backup.go#L246), the SA running K8up needs the same permissions, otherwise kubernetes denies it for security reasons. Obviously, otherwise an SA could give itself more access than it has itself.

Maybe that can be revisited later or in K8up v2

PR #222, opened 2020-12-21
Enable pre release

### LINKED ISSUE #220: Enable us to create release candidates without touching the `latest` tag
## Summary

**As** a developer of K8up
**I want** to create release of type "pre-release"
**So that** in order to build and push assets to be used in testing.

## Context

Sometimes testing require a deployment to clusters managed with some sort of configuration management. Having a release, makes things easier. Certain release might also require to update surrounding code like Helm charts or Commodore components.

## Out of Scope

* Replacing the existing tooling (goreleaser, GitHub actions).

## Further links

* Case with Helm chart: https://github.com/appuio/charts/pull/224

## Acceptance criteria

*Given* a pre-release with git tag
*When* I push the git tag
*Then* all images get built and pushed in GitHub workflow
*Except* the `latest` container image tag

corvus-ch:
Resolves #220

## Checklist

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -34,6 +34,8 @@ changelog:
     - '^Merge pull request'
 
 release:
+  prerelease: auto
+
   github:
     owner: vshn
     name: k8up
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -23,12 +23,14 @@ dockers:
 # and therefore eliminates `latest` from the list. This is OK to do since
 # `goreleaser` builds the image only once and tagging is a cheap operation.
 - image_templates:
-  - "docker.io/vshn/k8up:{{ if .Prerelease }}v{{ .Version }}{{ else }}latest{{ end }}"
   - "docker.io/vshn/k8up:v{{ .Version }}"
-  - "docker.io/vshn/k8up:v{{ .Major }}"
-  - "quay.io/vshn/k8up:{{ if .Prerelease }}v{{ .Version }}{{ else }}latest{{ end }}"
   - "quay.io/vshn/k8up:v{{ .Version }}"
-  - "quay.io/vshn/k8up:v{{ .Major }}"
+
+  - "docker.io/vshn/k8up:{{ if .Prerelease }}v{{ .Version }}{{ else }}v{{ .Major }}{{ end }}"
+  - "quay.io/vshn/k8up:{{ if .Prerelease }}v{{ .Version }}{{ else }}v{{ .Major }}{{ end }}"
+
+  - "docker.io/vshn/k8up:{{ if .Prerelease }}v{{ .Version }}{{ else }}latest{{ end }}"
+  - "quay.io/vshn/k8up:{{ if .Prerelease }}v{{ .Version }}{{ else }}latest{{ end }}"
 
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -16,11 +16,17 @@ snapshot:
...
 dockers:
+# The `latest` is not touched for pre release.
+# The template string `{{ if .Prerelease }}v{{ .Version }}{{ else }}latest{{ end }}`
+# will compute to `latest` for regular release. For pre releases, it will
+# compute to `v{{ .Version }}`. This duplicates one of the other list items
+# and therefore eliminates `latest` from the list. This is OK to do since
+# `goreleaser` builds the image only once and tagging is a cheap operation.
 - image_templates:
-  - "docker.io/vshn/k8up:latest"
+  - "docker.io/vshn/k8up:{{ if .Prerelease }}v{{ .Version }}{{ else }}latest{{ end }}"
```
ccremer, line 26: what about this?
```suggestion
  {{ if .Prerelease }}- "docker.io/vshn/k8up:latest"{{ end }}
```

.goreleaser.yml
```diff
@@ -16,11 +16,17 @@ snapshot:
...
   - "docker.io/vshn/k8up:v{{ .Version }}"
   - "docker.io/vshn/k8up:v{{ .Major }}"
-  - "quay.io/vshn/k8up:latest"
+  - "quay.io/vshn/k8up:{{ if .Prerelease }}v{{ .Version }}{{ else }}latest{{ end }}"
   - "quay.io/vshn/k8up:v{{ .Version }}"
   - "quay.io/vshn/k8up:v{{ .Major }}"
```
ccremer, line 31: We also need to prevent a push to the floating Major version tag. In other words, only the `.Version` image should be pushed

ccremer (changes requested):

.goreleaser.yml
corvus-ch: This will be invalid YAML. And `- "{{ if not .Prerelease }}docker.io/vshn/k8up:latest{{ end }}` will result in an empty string and will make `docker tag` to fail.

.goreleaser.yml
corvus-ch: What about `v{{ .Major }}{{ if .Prerelease }}-{{ .Prerelease }}{{ end }}`?

ccremer: that would print something like `vshn/k8up:v1-rc1` with `v1.0.0-rc1` (just tested locally with goreleaser without publishing).
so, that means any rc1 releases are floating, e.g. `1.1-rc1`, `1.2-rc1`, `1.3-rc1`... I doubt that's useful

.goreleaser.yml
ccremer: right, just went across this error when locally testing

ccremer (approved):
i played around locally as well. I've allowed myself a commit regarding the major version. This should now only push the exact version for prereleases, but for real releases there's additionally the latest and major image tags.

.goreleaser.yml
corvus-ch: Agree, that is not useful.

PR #240, opened 2020-12-28
Fix listing of PreBackupPods in foreign namespaces

### LINKED ISSUE #238: k8up starts prebackuppods from other namespaces
## Describe the bug

It tries to start mariadb and solr prebackuppods:
```
❯ kubectl -n syn-backup-cluster-objects get pods
NAME                                    READY   STATUS                       RESTARTS   AGE
k8s-objects-8559547bd7-szcxz            1/1     Running                      0          2d21h
mariadb-prebackuppod-6c455fd5cd-kxgsm   0/1     CreateContainerConfigError   0          2d21h
solr-prebackuppod-5f49cf8dcc-wmskr      0/1     Pending                      0          6d17h
```
even though in this namespace there are no prebackuppods for mariadb and solr defined:
```
❯ kubectl -n syn-backup-cluster-objects get prebackuppod
NAME          AGE
k8s-objects   49d
```

## Additional context

* There should be some namespace filtering logic defined here:
https://github.com/vshn/k8up/blob/a257728e29b9b5b6fb255ea3fc46911429a3bcf6/prebackup/prebackup.go#L72-L77

## Expected behavior

K8up should not create Pods and Deployments in foreign namespaces.

## To Reproduce

1. Create 1 PrebackupPod CRDs in 2 different namespaces each
2. Create a Schedule in namespace 1
3. Observe pods being created from other namespace too

- Image Version: v1.0.0-rc1
- K8s Version: v1.18
- K8s Distribution: Rancher

ccremer:
Fixes #238

## Checklist

- [x] Update tests.

### DIFF
--- a/prebackup/prebackup_test.go
+++ b/prebackup/prebackup_test.go
@@ -1,0 +1,67 @@
+package prebackup
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	k8upv1alpha1 "github.com/vshn/k8up/api/v1alpha1"
+	"github.com/vshn/k8up/job"
+)
+
+func TestPreBackup_getPodTemplates(t *testing.T) {
+	ns := "test-namespace"
+	backupObj := &k8upv1alpha1.Backup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "my-backup",
+			Namespace: ns,
+		},
+	}
+	tests := map[string]struct {
+		initFakeObjects        []runtime.Object
+		expectPodTemplateNames []string
+	}{
+		"GivenPrebackupObjectInDifferentNamespaces_WhenListingPrebackupTemplates_ThenIgnoreTemplatesFromOtherNamespaces": {
+			initFakeObjects: []runtime.Object{
+				backupObj,
+				createPrebackupPod(ns, "backup-pod"),
+				createPrebackupPod("foreign-namespace", "foreign-pod"),
+			},
+			expectPodTemplateNames: []string{"backup-pod"},
+		},
+	}
+	scheme := runtime.NewScheme()
+	require.NoError(t, k8upv1alpha1.SchemeBuilder.AddToScheme(scheme))
+
+	for name, tt := range tests {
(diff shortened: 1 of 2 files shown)


prebackup/prebackup_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+func TestPreBackup_getPodTemplates(t *testing.T) {
+	ns := "test-namespace"
+	backupObj := &k8upv1alpha1.Backup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "my-backup",
+			Namespace: ns,
+		},
+	}
+	tests := map[string]struct {
```
Kidswiss, line 24: This map could be defined outside of the function. Makes it re-usable and the function smaller and easier to read.

Kidswiss (changes requested):
LGTM

Just curious: was the test file generated by some tool? Asking just so we all use more or less the same

prebackup/prebackup_test.go
ccremer: It depends on the test. Not all test cases have the same input and output, thus a dynamic type with specialised fields within the function gives most flexibility. e.g. `expectPodTemplateNames` might not be needed in all tests

ccremer:
> Just curious: was the test file generated by some tool? Asking just so we all use more or less the same

It's default when quick-generating tests with Goland (IDE). The only difference is, that Goland produces a slice of Test cases with the `name` property. However, to keep tests named uniquely I usually make that a `map[string]struct{}` where the string is the name of the test case and following the "GivenPrecondition_WhenAction_ThenResult" paradigma.

prebackup/prebackup_test.go
Kidswiss: Yes that's true, but moving it outside the function makes the code more readable IMHO.

ccremer: I tried to address your readability concern with a helper function, this should already help a lot.

If the test cases are defined outside, I would have to scroll up and down way more :) With table-like test cases, the test cases with different names, preconditions and expectations become more important than the test code, thus I prefer to have it together. If we can share the same cases with same preconditions and expectations for another test, I'll be happy to consolidate those outside, but in this single test I don't see much value.

Kidswiss: Okay let's move the discussions here: https://github.com/vshn/k8up/discussions/231 :)

Kidswiss (approved):

prebackup/prebackup_test.go
ccremer: Fine by me. We likely have to touch every test anyway once agreed on something, this shouldn't be a blocker for this PR though.

PR #242, opened 2020-12-28
Replace internal schedules when reconciling Schedule CRs

### LINKED ISSUE #237: [Bug] Updating schedule in Schedule spec does not update internal scheduler
## Describe the bug

Changed the schedule in a Schedule CR does not update the schedule in 

## Additional context

A restart of K8up is likely going to fix the problem, since the internal scheduler will be rebuilt from the Spec

## Logs

```console
2020-12-23T15:27:50.772+0100	DEBUG	controller	Successfully Reconciled	{"reconcilerGroup": "backup.appuio.ch", "reconcilerKind": "Schedule", "controller": "schedule", "name": "k8up-lagoon-backup-schedule", "namespace": "drupal-example-test1-dev"}
```

## Expected behavior

K8up should replace the internal schedule to the new spec

## To Reproduce

1. Create schedule spec
```yaml
spec:
  backup:
    schedule: '*/30 * * * *' # something bigger than 10min
```
1. Change schedule spec
```yaml
spec:
  backup:
    schedule: '* * * * *'
```
2. waited for 10 minutes and still nothing
also no backup object was created:
```
❯ k get backups
No resources found in drupal-example-test1-dev namespace.
```

- Image Version: e.g. v1.0.0-rc1
- K8s Version: v1.18
- K8s Distribution: Rancher

ccremer:
Fixes #237

Neither the Cron library nor our internal wrapping have a method to Update existing schedules, thus we recreate the schedules when reconciling Schedule CRs.

## Checklist

- [ ] Update tests.

### DIFF
--- a/api/v1alpha1/job_types.go
+++ b/api/v1alpha1/job_types.go
@@ -1,7 +1,9 @@
 package v1alpha1
 
-// JobType defines what job type this is.
-type JobType string
+type (
+	// JobType defines what job type this is.
+	JobType string
+)
 
 const (
 	BackupType   JobType = "backup"
--- a/handler/schedule.go
+++ b/handler/schedule.go
@@ -55,7 +55,7 @@ func (s *ScheduleHandler) Handle() error {
 	jobList := s.createJobList()
 
 	scheduler.GetScheduler().RemoveSchedules(namespacedName)
-	err = scheduler.GetScheduler().AddSchedules(jobList)
+	err = scheduler.GetScheduler().SyncSchedules(jobList)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


ccremer:
@Kidswiss did you have a look yet?

Regarding the 4 codeclimate issues: I do not see an easy way to reduce code duplication without creating a bunch of interfaces that are already very similar to existing ones, but don't match in the types. Would you be ok with accepting them and ignore until K8up v2?

Kidswiss:
Yes for the sake of fixing this bug let's stick with the quick solution. But please open a new issue to solve this specifically.

Rest LGTM

Kidswiss (approved):

PR #248, opened 2020-12-29
Replace Goreleaser changelog generator with PR-based generator

ccremer:
## Summary

Goreleaser builds changelog from commits, which are too detailed and uncategorized for human readers. 
The alternative changelog generator uses GitHub API to build a changelog from PRs, categorized by labels.

## Checklist

### DIFF
--- a/.github/changelog-configuration.json
+++ b/.github/changelog-configuration.json
@@ -1,0 +1,29 @@
+{
+    "pr_template": "- ${{TITLE}} (#${{NUMBER}})",
+    "categories": [
+        {
+            "title": "## 🚀 Features",
+            "labels": ["enhancement", "feature"]
+        },
+        {
+            "title": "## 🛠️ Minor Changes",
+            "labels": ["change"]
+        },
+        {
+            "title": "## 🔎 Breaking Changes",
+            "labels": ["breaking"]
+        },
+        {
+            "title": "## 🐛 Fixes",
+            "labels": ["bug", "fix"]
+        },
+        {
+            "title": "## 📄 Documentation",
+            "labels": ["documentation"]
+        },
+        {
+            "title": "## 🔗 Dependency Updates",
+            "labels": ["dependency"]
+        }
+    ]
+}
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,7 @@
+## Summary
+
+
+
 <!--
 Thank you for your pull request. Please provide a description above and
 review the checklist below.
@@ -9,6 +13,9 @@ Remove items that do not apply. For completed items, change [ ] to [x].
 -->
 
 - [ ] Keep pull requests small so they can be easily reviewed.
+- [ ] Categorize the PR by setting a good title and adding one of the labels:
+      `bug`, `enhancement`, `documentation`, `change`, `breaking`,
+      as they show up in the changelog
 - [ ] Update the documentation.
 - [ ] Update tests.
 - [ ] Link this PR to related issues.
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,9 +29,21 @@ jobs:
       run: docker login -u "${{ secrets.QUAY_IO_USERNAME }}" -p "${{ secrets.QUAY_IO_PASSWORD }}" quay.io
     - name: Generate artifacts
       run: make crd
+    - name: Build changelog from PRs with labels
+      id: build_changelog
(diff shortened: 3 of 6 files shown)


.github/changelog-configuration.json
```diff
@@ -0,0 +1,20 @@
+{
+    "categories": [
+        {
+            "title": "## 🚀 Features",
+            "labels": ["enhancement", "feature"]
+        },
+        {
+            "title": "## ⚠️ (Breaking) Changes",
+            "labels": ["change"]
```
ccremer, line 9: Should we make a further separation between non-breaking changes and breaking changes?
If so, what should the labels be named? Can we even keep this discipline? Ideally, only major releases should get breaking changes (hence the major release), but this would also list general, probably unrelated/uninteresting changes like "Refactor backup logic".

Or should we remove "(Breaking)" from the Title altogether and not make a distinction, relying on any official migration guides in the documentation?

Kidswiss: I'd say we stick to the semver as much as possible, so only major releases should break stuff. Which would make the breaking changes here obsolete.

ccremer: Going by semver alone would also mean that we don't categorize documentation PRs (even simple typos etc). I personally like a distinction between "changes that are infrastructure/quality (git, pipelines, makefile etc)" related and actual, new features of K8up itself. Now we could argue to ignore all quality or infra related stuff that aren't strictly related to semver, but I'd prefer to have them listed too, but not under bugfixes or new features.

.github/changelog-configuration.json
```diff
@@ -0,0 +1,20 @@
+{
+    "categories": [
```
ccremer, line 2: The config allows setting some formatting option per PR.
Default config looks like this:

image

Maybe put the PR number onto the same single line?
```suggestion
    "pr_template": "- ${{TITLE}} (PR #${{NUMBER}})",
    "categories": [
```

.github/changelog-configuration.json
ccremer: Also, there's a lot of merged Renovate PRs. I think for those alone it's worth to have an "uncategorized" or "dependency update" section. I doubt that we'll be able to label all Renovate PRs, as they're merged pretty fast while forgetting to set a label.

tobru: I'd remove the "(Breaking)" from the title and find another, less intrusive icon. Breaking changes should be documented separately if there are any, in a different category (label). I would also only mark changes "breaking" if there is any user-interaction required when upgrading K8up.

Listing infrastructure updates in a separate category would be cool, as well as documentation and dependency updates.

> I doubt that we'll be able to label all Renovate PRs
Afaik Renovate can be configured to automatically set labels.

Categorizing things in the changelog allows the reader to distinguish the changes and decide if they are interesting to the reader or not and can skip it then.

For inspiration: Other projects using icons in releases:
* https://github.com/TryGhost/Ghost/releases
* https://github.com/RocketChat/Rocket.Chat/releases/

.github/changelog-configuration.json
tobru: > Maybe put the PR number onto the same single line?

Yes, please :+1:

tobru (approved):

.github/changelog-configuration.json
Kidswiss: Right, I didn't think about those infrastructure fixes.

Listing them separately would be cool, I like how rocketchat does their changelog!

ccremer: > Afaik Renovate can be configured to automatically set labels.

Neat, didn't know! Adjusted PR with category.

I have categorized Infrastructure-related changes as "minor changes" and made a separate for breaking changes. This time with a magnifying glass, indicating that this is a section users should look into. Any other proposals?

tobru: Looks good, I'd say go for it :+1:

ccremer: I'm thinking about better title for breaking changes. Coming up with a "weaker" word for "Breaking" is not that easy:
* "Notable changes" (goes well with magnifying glass icon, but maybe too weak)
* "user-facing changes" (not my favorite)
* "Changes (Action needed)" or "Changes (Migration needed)"

any other? or we keep "breaking changes"?

Kidswiss: Well for breaking changes I'd keep the breaking to be honest.

ccremer: Ok, I think this is good enough for now :)

ccremer:
Thanks for the discussions :tada:

PR #250, opened 2020-12-30
Replace local registry by loading e2e images directly onto KIND cluster

ccremer:
## Summary

Sometimes there are pull/connection issues when cluster tries to load from docker registry when executing e2e tests multiple times.
This should now also restart the operator when the image is changed.
It also reduces disk usage by half by not having to store the images in cluster and ALSO on the local registry.

It's still recommended to delete the kind cluster from time to time, as older images are still within the cluster.

## Checklist

- [x] Update tests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,7 +60,7 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-go-
     - name: Run e2e tests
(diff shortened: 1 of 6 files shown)


Kidswiss:
> It's still recommended to delete the kind cluster from time to time, as older images are still within the cluster.

Shouldn't the kind cluster be deleted anyways after the e2e test ran? So every e2e run has a clean slate.

ccremer:
> Shouldn't the kind cluster be deleted anyways after the e2e test ran? So every e2e run has a clean slate.

No, that would just slow down local development and that's a pain, as each cluster install takes ~1min. There is already a `make clean_e2e_setup` that removes the namespace and CRD, usually sufficient

PR #253, opened 2021-01-04
Show link to changelog in the docs

ccremer:
## Summary

* This will display a link in the navigation pointing to GitHub release, but only when the Antora component was built from a tag in the playbook (not master for example), see https://docs.antora.org/antora/2.3/page/intrinsic-attributes/#page-attributes

## Checklist


tobru:
Cool. Never used that before, let's see how this works when we tagged the release.

tobru (approved):

Issue #198, opened 2020-12-10
Extract e2e Targets from /Makefile to /e2e/Makefile


cimnine:
## Summary

**As** developer\
**I want** code organised in a cohesive way, such that related code is located in about the same location\
**So that** I find the relevant pieces with ease.

## Context

Currently, the main `Makefile` contains a lot of e2e test related code. Yet there is also the `e2e/Makefile` which would be a more suitable place for the implementation of some of the make targets.

## Out of Scope

* Any changes to the actual e2e test suite implementation.

## Further links

* [Comment that sparked this issue](https://github.com/vshn/k8up/pull/189#pullrequestreview-547378906)

## Acceptance criteria

*Given* the `Makefile` in the root of the repository\
*When* looking at the `Makefile`\
*Then* most of the code should be regarding building and releasing k8up, and just little code should be about the e2e test suite.

*Given* a fresh checkout of the repository\
*When* `make e2e_test` is executed\
*Then* the e2e test suite should be initialised (i.e. dependencies installed and set up) and run

*Given* `make e2e_test` did run\
*When* `make clean` is executed\
*Then* the local repository should be in it's initial state and all e2e test suite dependencies should be removed.

## Implementation Ideas

Move most of the e2e related code from `/Makefile` to `/e2e/Makefile`.

PR #255, opened 2021-01-04
Separate e2e tests and KIND from main Makefile

cimnine:
## Summary

Re-organized the Makefile with regards to #198 .

Left to do:
- [x] Actually extract code from `./Makefile` into another file

## Checklist

- ~~Update tests.~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,12 +27,12 @@ CRD_ROOT_DIR ?= config/crd/apiextensions.k8s.io
 TESTBIN_DIR ?= ./testbin/bin
 KIND_BIN ?= $(TESTBIN_DIR)/kind
 KIND_VERSION ?= 0.9.0
-KIND_KUBECONFIG ?= ./testbin/kind-kubeconfig
 KIND_NODE_VERSION ?= v1.18.8
+KIND_KUBECONFIG ?= ./testbin/kind-kubeconfig-$(KIND_NODE_VERSION)
 KIND_CLUSTER ?= k8up-$(KIND_NODE_VERSION)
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -131,56 +131,70 @@ bundle: generate ## Generate bundle manifests and metadata, then validate genera
...
 
+###
+### KIND
+###
+
+.PHONY: rund_kind
+run_kind: export KUBECONFIG = $(KIND_KUBECONFIG)
+run_kind: export BACKUP_ENABLE_LEADER_ELECTION = $(ENABLE_LEADER_ELECTION)
+run_kind: $(SETUP_KIND) ## Runs the operator in kind
+	go run ./main.go
```
ccremer, line 150: I propose to move the KIND related stuff into e2e/Makefile and include it in the main Makefile with `include e2e/Makefile`. This would make main Makefile easier to maintain while clearly separating between "building" and "running"

ccremer (commented):
I would propose to introduce lower-case variables for "make internal" variables, not necessary intended to be edited by developers (e.g. `$(KIND_BIN)` to `$(kind_bin)`. Additionally, I'd propose to create a `Makevariables.mk` file, to be included with `include Makevariables.mk` at the top of `Makefile`. This variables file contains only the stuff to be edited by developer or per project (e.g. docker image names etc), so that the main makefile becomes easier to port to other repositories, and indicates better which variables are meant to be customized.

ccremer (approved):

ccremer (approved):

Issue #256, opened 2021-01-04
Operator starts wrong job types


ccremer:
## Describe the bug

During testing the newest RC of the k8up operator running on test1, it's observed that the operator seems to start a prune pod instead of a backup pod when a backup schedule is changed so that the backup runs at a specific time. 

it's assumed this behavior happens because a prune hasn't previously completed successfully for this project. This prune pod then starts and fails to run because it is missing the AWS secret key in its env vars. This prune pod will then remain in crashloop backoff state for about 30 minutes due to the missing AWS secret key before it is marked as failed by the operator. After this prune job fails, no backup is attempted afterwards.

## Additional context

* specifically observed this behavior on both the `drupal-example-test1-master` and `drupal-example-test1-develop` environments.
* verified that the missing secret key does seem to exist in the `syn-backup` namespace's `global-bucket-credentials` secret, so this may be an issue with how the operator is creating that prune pod specification?
* As the logs show, the controller correctly parsed out the timeframe the prune should run, but it still tried to schedule a prune instead of a backup. This prune then failed, and the controller did not attempt to perform a backup.
* tried to remove the `prune` key in the schedule object, and a check was attempted instead of a backup, which still failed for the same reason. After removing both the `prune` and `check` keys, it was found that a backup was started and the pre backup pods ran through successfully, but the same issue with the missing AWS secret key still persists, causing that backup pod to crashloop.

## Logs

If applicable, add logs to help explain your problem.
```console
2020-12-31T17:18:09.973+0100	INFO	controllers.Schedule	registering schedule for	{"schedule": "drupal-example-test1-master/k8up-lagoon-backup-schedule", "type": "backup", "schedule": "19 * * * *"}
2020-12-31T17:18:09.973+0100	INFO	controllers.Schedule	registering schedule for	{"schedule": "drupal-example-test1-master/k8up-lagoon-backup-schedule", "type": "check", "schedule": "28 3 * * 0"}
2020-12-31T17:18:09.973+0100	INFO	controllers.Schedule	registering schedule for	{"schedule": "drupal-example-test1-master/k8up-lagoon-backup-schedule", "type": "prune", "schedule": "28 3 * * 6"}
2020-12-31T17:18:09.973+0100	DEBUG	controller	Successfully Reconciled	{"reconcilerGroup": "backup.appuio.ch", "reconcilerKind": "Schedule", "controller": "schedule", "name": "k8up-lagoon-backup-schedule", "namespace": "drupal-example-test1-master"}
[...]

PR #258, opened 2021-01-05
Ensure backwards compatibility of GLOBALSECRETACCESSKEY variable

cimnine:
## Summary

One complaint in #256 was, that the AWS S3 Secret Access Key is not passed on to the container. This seems to be a regression from the refactoring work.

## Checklist


ccremer (changes requested):
I think it should be the other way round. It looks like with either the rewrite or with the new config library we introduced a breaking change.
In [v0.1.10](https://github.com/vshn/k8up/blob/v0.1.10/config/global.go#L44) we use the env vars without the `id` suffix.
IMO we should fix this in code (`koanf` tag)

Kidswiss:
I agree I'd currently go with the non-backwards breaking stuff.

PR #259, opened 2021-01-05
Add junit reporting for e2e tests

ccremer:
## Summary

* Bats supports formatting tests with junit, and there's a cool junit GH action, so let's marry them!
* On successful runs, the reports may be added to the "Lint" workflow suite. [This is a bug by GitHub](https://github.com/mikepenz/action-junit-report/issues/34) and just of a cosmetic nature.

An example of a failing test can be found here: https://github.com/vshn/k8up/pull/259/checks?check_run_id=1651595468

image

## Checklist

- [x] Update tests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,9 +60,13 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-go-
     - name: Run e2e tests
-      run: make crd e2e-test -e CRD_SPEC_VERSION=${{ matrix.crd-spec-version }} -e KIND_NODE_VERSION=${{ matrix.kind-node-version }} -e KIND_KUBECTL_ARGS=--validate=false
-    - name: Show e2e debug logs
-      run: cat e2e/debug/detik/*
+      run: make crd e2e-test -e CRD_SPEC_VERSION=${{ matrix.crd-spec-version }} -e KIND_NODE_VERSION=${{ matrix.kind-node-version }} -e KIND_KUBECTL_ARGS=--validate=false -e bats_args="-F junit || true"
+    - name: Publish Test Report
+      uses: mikepenz/action-junit-report@v1
+      with:
+        report_paths: '**/e2e/TestReport-*.xml'
+        github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -60,9 +60,13 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-go-
     - name: Run e2e tests
-      run: make crd e2e_test -e CRD_SPEC_VERSION=${{ matrix.crd-spec-version }} -e KIND_NODE_VERSION=${{ matrix.kind-node-version }} -e KIND_KUBECTL_ARGS=--validate=false
-    - name: Show e2e debug logs
-      run: cat e2e/debug/detik/*
+      run: make crd e2e-test -e CRD_SPEC_VERSION=${{ matrix.crd-spec-version }} -e KIND_NODE_VERSION=${{ matrix.kind-node-version }} -e KIND_KUBECTL_ARGS=--validate=false -e bats_args="-F junit || true"
```
ccremer, line 63: Had to append `bats_args="-F junit || true"` to the bats command in order to be able to publish test reports. Bats would otherwise exit on errors without having a chance to upload the reports because GH workflow cancels on failures.

Any better solutions?

e2e/.gitignore
```diff
@@ -0,0 +1 @@
+TestReport*.xml
```
ccremer, line 1: This should not even be applicable to local environment, since by default bats is called without the junit formatter. Should this be removed? It doesn't hurt, and maybe someone runs bats with local junit formatter, probably there are some IDE support for junit reports files.

Kidswiss: I'd keep it.

.github/workflows/build.yml
Kidswiss: Just for clarification: the `|| true` is there in case bats exits with errors, but still has something to upload?

ccremer: yes, exactly

Kidswiss: If there's no flag to catch it in bats directly I'd say that's the way to go.

Kidswiss (approved):

tobru (approved):

PR #260, opened 2021-01-05
Pass Scheduler Backend Config to Childresources

cimnine:
## Summary

This PR fixes a regression by which k8up did not apply the _backend_ section of a _Schedule_ custom resource to the created sub-resources, i.e. _Backup_, _Check_, _Prune_, etc.

This fixes one of the findings in #256.

## Checklist

- [x] Update tests.


handler/schedule.go
```diff
@@ -79,63 +79,77 @@ func (s *ScheduleHandler) createJobList() scheduler.JobList {
 		Jobs:   make([]scheduler.Job, 0),
 	}
 
-	if s.schedule.Spec.Archive != nil {
-		s.schedule.Spec.Archive.ArchiveSpec.Resources = s.mergeResourcesWithDefaults(s.schedule.Spec.Archive.Resources)
+	if archive := s.schedule.Spec.Archive; archive != nil {
```
Kidswiss, line 82: I've discussed this if cascade here with @ccremer already.

I feel like it's now the right time to try to get rid of this. WDYT?

ccremer:
Since you touched resource merging stuff I tested resource templates in this branch, to see how that works now.

We basically have 3 levels where we can specify resources, each overrides the previous one:
1. Global resource limits and requests
2. `resourceRequirementsTemplate` in the schedule Spec
3. `resources` of a job spec (e.g. prune)

This PR has the behaviour so that the `Schedule` after reconciling has additional resources specified in the Job spec.
e.g. 
```yaml
---
kind: Schedule
metadata:
  name: schedule-test
spec:
  resourceRequirementsTemplate:
    requests:
      cpu: "250m"
  prune:
    schedule: '*/1 * * * *'
```
becomes
```yaml
kind: Schedule
metadata:
  name: schedule-test
  namespace: default
spec:
...
  prune:
    resources:
      requests:
        cpu: 250m
    retention:
      keepDaily: 14
      keepLast: 5
    schedule: '*/1 * * * *'
  resourceRequirementsTemplate:
    requests:
      cpu: 250m
```
(only important thing: `spec.prune.resources` got added)
And the prune spec gets as expected:
```yaml
kind: Prune
metadata:
  name: prunejob-1609946760
  namespace: default
spec:
  resources:
    requests:
      cpu: 250m
...
```
All the while the operator has this env var set: `kubectl set env deployment/k8up-operator BACKUP_GLOBALMEMORY_LIMIT=76M -n k8up-system`. But the `prunejob-1609946760` CR doesn't get the memory limit.

There are 2 issues:
1. The `resources` of a specific Job spec in `test-schedule` gets added after reconciling. I think this shouldn't be as it would make changing a value in `resourceRequirementsTemplate` impossible because `spec.<job>.resources` is more specific. Only when creating a `Prune` object, the final values should be in there, but not in the `Schedule`
2. The global resource values aren't applied at all. Though this could already be the case in the current master branch. Insufficient testing? :thinking: In any case, if `spec.<job>.resources` again gets set in the spec, it would disallow cluster administrators to effectively alter resource requirements after the first reconciliation (given users don't override this in the initial source YAML themselves, then it's their responsibility to meet cluster resource policies).

In code this means that merged resources should be kept as runtime variables and stored in the `Prune` CR. And not stored in the Schedule spec after reconciliation.
See also #143

Would you like a separate issue for this bug? Or do you think you can handle this in this PR when you're already touching the code anyway?

handler/schedule.go
cimnine: A generic solution – one that's not just some hack on top – is not trivial unfortunately. I've tried before and tried this morning, but was not able to come up with a nice solution. I'm open to suggestions :)

cimnine:
> The global resource values aren't applied at all. Though this could already be the case in the current master branch.

I have not confirmed this, but since I did not actually touch that logic I suspect so. I believe that this can be handled in a separate issue.

> In code this means that merged resources should be kept as runtime variables and stored in the Prune CR. And not stored in the Schedule spec after reconciliation.

I have changed to logic accordingly. The reconciliation loop still creates the `Resources` key for some reason I have not been able to figure out yet (but I have stopped looking for now as it's just a _Schoenheitsfehler_, IMO).

handler/schedule.go
Kidswiss: Okay then let's move it into its own issue then.

api/v1alpha1/schedule_types.go
```diff
@@ -19,10 +19,13 @@ type ScheduleSpec struct {
 	ResourceRequirementsTemplate corev1.ResourceRequirements `json:"resourceRequirementsTemplate,omitempty"`
 }
 
+// ScheduleDefinition is the actual cron-type expression that defines the interval of the actions.
+type ScheduleDefinition string
```
cimnine, line 23: This change was not strictly necessary. It's a leftover from me trying to improve the critiqued 'if-cascade'. I figured it improved the code and therefore left this change in place.

ccremer: If we're making a type for a string, at least we should add a "String()" method, that's a bit easier to use and read than having to use string(...)

cimnine: Fair enough!

ccremer:
Running the current branch and applying the default sample actually throws panics:
```console
2021-01-08T10:21:15.355Z	INFO	controllers.Schedule	registering schedule for	{"schedule": "default/schedule-test", "type": "prune", "schedule": "*/2 * * * *"}
2021-01-08T10:21:15.355Z	DEBUG	controller	Successfully Reconciled	{"reconcilerGroup": "backup.appuio.ch", "reconcilerKind": "Schedule", "controller": "schedule", "name": "schedule-test", "namespace": "default"}
E0108 10:21:15.356032       1 runtime.go:78] Observed a panic: &reflect.ValueError{Method:"reflect.Value.Type", Kind:0x0} (reflect: call of reflect.Value.Type on zero Value)
goroutine 312 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x1514a80, 0xc00067fe20)
	/home/chris/go/pkg/mod/k8s.io/apimachinery@v0.18.10/pkg/util/runtime/runtime.go:74 +0xa6
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/home/chris/go/pkg/mod/k8s.io/apimachinery@v0.18.10/pkg/util/runtime/runtime.go:48 +0x89
panic(0x1514a80, 0xc00067fe20)
	/snap/go/6439/src/runtime/panic.go:969 +0x175
reflect.Value.Type(0x0, 0x0, 0x0, 0x18d19a0, 0x1625f20)
	/snap/go/6439/src/reflect/value.go:1894 +0x189
github.com/imdario/mergo.merge(0x15804a0, 0xc00067d6c0, 0x15804a0, 0x0, 0x0, 0x0, 0x0, 0x16e3327, 0xc)
	/home/chris/go/pkg/mod/github.com/imdario/mergo@v0.3.11/merge.go:358 +0x20a
github.com/imdario/mergo.Merge(...)
	/home/chris/go/pkg/mod/github.com/imdario/mergo@v0.3.11/merge.go:291
github.com/vshn/k8up/handler.(*ScheduleHandler).mergeBackendWithDefaults(0xc0006c7c48, 0xc00067d6c0)
	/home/chris/Documents/source/vshn/k8up/handler/schedule.go:152 +0x79
github.com/vshn/k8up/handler.(*ScheduleHandler).mergeWithDefaults(0xc0006c7c48, 0xc0004e3620)
	/home/chris/Documents/source/vshn/k8up/handler/schedule.go:139 +0x65
github.com/vshn/k8up/handler.(*ScheduleHandler).createJobList(0xc0006c7c48, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/chris/Documents/source/vshn/k8up/handler/schedule.go:91 +0x7bc
github.com/vshn/k8up/handler.(*ScheduleHandler).Handle(0xc0006c7c48, 0x8, 0xc0006c7b78)
	/home/chris/Documents/source/vshn/k8up/handler/schedule.go:54 +0x85
```

api/v1alpha1/schedule_types.go
cimnine: Although it isn't actually needed as I just figured out, as this simple example demonstrates:

```go
package main

import (
	"fmt"
)

type Foo string

func main() {
	foo := Foo("Hello, playground")
	fmt.Println(foo)
}
```

(Try it on https://play.golang.org/)

But there are other interesting things that can be done when that silly string is strongly typed. I've implemented two changes: _func (s ScheduleDefinition) IsRegularCron() bool_ and _func (s ScheduleDefinition) IsRandom() bool_. With it, IMO the flow of `getEffectiveSchedule` in `schedule.go` is easier to comprehend.

cimnine:
> Running the current branch and applying the default sample actually throws panics:

Interesting. My operator was running the whole night and I have not observed such a panic. Is it reproducible on your end? And would you mind sharing the actual schedule definition, since there are many floating around in the docs and the repo.

handler/schedule.go
```diff
@@ -155,25 +170,41 @@ func (s *ScheduleHandler) updateStatus() error {
...
+
+func (s *ScheduleHandler) getRandomSchedule(jobType k8upv1alpha1.JobType, originalSchedule k8upv1alpha1.ScheduleDefinition) (k8upv1alpha1.ScheduleDefinition, error) {
 	seed := s.createSeed(s.schedule, jobType)
-	schedule, err := randomizeSchedule(seed, originalSchedule)
+	randomizedSchedule, err := randomizeSchedule(seed, originalSchedule)
 	if err != nil {
-		s.Log.Info("Could not randomize schedule, ignoring and try original schedule", "error", err.Error())
-	} else {
-		s.Log.V(1).Info("Randomized schedule", "seed", seed, "from_schedule", originalSchedule, "effective_schedule", schedule)
+		s.Log.Info("Could not randomize schedule", "schedule", originalSchedule, "error", err.Error())
```
ccremer, line 199: I propose to move this log line to the caller in `getEffectiveSchedule`, so that we follow better the "either-log-or-throw-but-not-both" paradigma.

scheduler/scheduler.go
```diff
@@ -111,7 +111,7 @@ func (s *Scheduler) SyncSchedules(jobs JobList) error {
 
 // addSchedule adds the given newJobs to the cron scheduler
 func (s *Scheduler) addSchedule(jb Job, namespacedName types.NamespacedName, cmd func()) error {
-	id, err := s.cron.AddFunc(jb.Schedule, cmd)
+	id, err := s.cron.AddFunc(string(jb.Schedule), cmd)
```
ccremer, line 114: is the function already there?
```suggestion
	id, err := s.cron.AddFunc(jb.Schedule.String()), cmd)
```

ccremer (changes requested):

ccremer:
> Interesting. My operator was running the whole night and I have not observed such a panic. Is it reproducible on your end? And would you mind sharing the actual schedule definition, since there are many floating around in the docs and the repo.

```bash
make clean e2e-test
k apply -f config/samples/k8up_v1alpha1_schedule.yaml
```

api/v1alpha1/schedule_types.go
```diff
@@ -104,3 +109,17 @@ func (s *Schedule) GetStatus() *Status {
 func (s *Schedule) GetResources() corev1.ResourceRequirements {
 	return s.Spec.ResourceRequirementsTemplate
 }
+
+// IsNonStandard returns true if the value begins with "@",
+// indicating a special definition.
+// An example is '@weekly-random'
```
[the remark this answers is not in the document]
cimnine: I've chosen the term _non-standard_ because it corresponds to the documentation at https://k8up.io/k8up/1.0.0/references/schedule-specification.html.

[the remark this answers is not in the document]
ccremer: I understand where your reasoning comes from. I think it makes sense to open a PR so that we separate between non-standard and k8up-specific schedules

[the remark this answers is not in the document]
ccremer: there you go: https://github.com/vshn/k8up/pull/267

cimnine:
When I run `k apply -f config/samples/k8up_v1alpha1_schedule.yaml` against the correct cluster it indeed crashed my k8up as well. This has now been fixed.

ccremer (approved):

cimnine:
Rebased onto current master.

Issue #257, opened 2021-01-04
K8up spams the same jobs to the queue


ccremer:
## Describe the bug

The operator logs is filled with messages like below

## Additional context

* This might somehow be caused by reconciliation loops
* Also, the check is older than 1.0.0-rc2, maybe a version migration issue?

## Logs

If applicable, add logs to help explain your problem.
```console
2021-01-04T16:24:55.650+0100	INFO	controllers.Check	adding job to the queue	{"check": "drupal-example-test1-controller-master/checkjob-1608436800"}
2021-01-04T16:25:25.650+0100	INFO	controllers.Check	adding job to the queue	{"check": "drupal-example-test1-controller-master2/checkjob-1608433440"}
2021-01-04T16:25:25.650+0100	INFO	controllers.Check	adding job to the queue	{"check": "drupal-example-test1-controller-master/checkjob-1608436800"}
2021-01-04T16:25:55.650+0100	INFO	controllers.Check	adding job to the queue	{"check": "drupal-example-test1-controller-master2/checkjob-1608433440"}
2021-01-04T16:25:55.650+0100	INFO	controllers.Check	adding job to the queue	{"check": "drupal-example-test1-controller-master/checkjob-1608436800"}
2021-01-04T16:26:25.650+0100	INFO	controllers.Check	adding job to the queue	{"check": "drupal-example-test1-controller-master2/checkjob-1608433440"}
2021-01-04T16:26:25.650+0100	INFO	controllers.Check	adding job to the queue	{"check": "drupal-example-test1-controller-master/checkjob-1608436800"}
2021-01-04T16:26:55.651+0100	INFO	controllers.Check	adding job to the queue	{"check": "drupal-example-test1-controller-master2/checkjob-1608433440"}
2021-01-04T16:26:55.651+0100	INFO	controllers.Check	adding job to the queue	{"check": "drupal-example-test1-controller-master/checkjob-1608436800"}
2021-01-04T16:27:25.651+0100	INFO	controllers.Check	adding job to the queue	{"check": "drupal-example-test1-controller-master2/checkjob-1608433440"}
2021-01-04T16:27:25.651+0100	INFO	controllers.Check	adding job to the queue	{"check": "drupal-example-test1-controller-master/checkjob-1608436800"}
2021-01-04T16:27:55.651+0100	INFO	controllers.Check	adding job to the queue	{"check": "drupal-example-test1-controller-master2/checkjob-1608433440"}
[...]

PR #268, opened 2021-01-08
Avoid reconciliation loops with predicates

ccremer:
## Summary

From the godoc:

---
GenerationChangedPredicate implements a default update predicate function on Generation change.
This predicate will skip update events that have no change in the object's metadata.generation field. 
The metadata.generation field of an object is incremented by the API server when writes are made to the spec field of an object. 
This allows a controller to ignore update events where the spec is unchanged, and only the metadata and/or status fields are changed.
For CustomResource objects the Generation is only incremented when the status subresource is enabled.
Caveats:
* The assumption that the Generation is incremented only on writing to the spec does not hold for all APIs. 
  E.g For Deployment objects the Generation is also incremented on writes to the metadata.annotations field. 
  For object types other than CustomResources be sure to verify which fields will trigger a Generation increment when they are written to.
* With this predicate, any update events with writes only to the status field will not be reconciled. 
  So in the event that the status block is overwritten or wiped by someone else the controller will not self-correct to restore the correct status.

---
* In short: This should avoid a reconciliation loop when we update the status by ourselves. but also means that if someone else manually updates the status field then no reconciliation will be triggered either. I don't a reason why a user would do that, but hey...
* May also be relevant for #257 

## Checklist

### DIFF
--- a/controllers/archive_controller.go
+++ b/controllers/archive_controller.go
@@ -9,6 +9,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	k8upv1alpha1 "github.com/vshn/k8up/api/v1alpha1"
 	"github.com/vshn/k8up/cfg"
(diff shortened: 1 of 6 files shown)


cimnine:
> May also be relevant for #257

#257 should be fixed with https://github.com/vshn/k8up/pull/266, it wasn't a problem with too many reconciliations but rather a problem of updating the status information. This PR would not have fixed that problem.

Still, I believe this PR will avoid unnecessary reconciliation loops which improves many factors of k8up (i.e. load and log output, possibly others)

cimnine (approved):

PR #269, opened 2021-01-08
Load the e2e image into KIND in its own make-target

ccremer:
## Summary

* This should allow cases for local development where the image should be updated
without having to run the e2e tests themselves (e.g. manually)
* Further checks in the e2e whether the new image is being seen, and not the old pod from the previous replicaset.

## Checklist

- [x] Update tests.

### DIFF
--- a/Makefile.vars.mk
+++ b/Makefile.vars.mk
@@ -1,6 +1,5 @@
 # Current Operator version
-VERSION ?= $(shell date +%Y-%m-%d_%H-%M-%S)
-VERSION := $(VERSION)
+VERSION ?=
 
 # Default bundle image tag
 BUNDLE_IMG ?= controller-bundle:$(VERSION)
--- a/Makefile.vars.mk
+++ b/Makefile.vars.mk
@@ -1,9 +1,3 @@
-# Current Operator version
-VERSION ?=
-
-# Default bundle image tag
-BUNDLE_IMG ?= controller-bundle:$(VERSION)
-
 IMG_TAG ?= latest
 
 BIN_FILENAME ?= $(PROJECT_ROOT_DIR)/k8up
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ MAKEFLAGS += --no-builtin-variables
 PROJECT_ROOT_DIR = .
 include Makefile.vars.mk
 
-e2e_make := $(MAKE) -C e2e -e 'VERSION=$(VERSION)'
+e2e_make := $(MAKE) -C e2e
 go_build ?= CGO_ENABLED=0 go build -o $(BIN_FILENAME) main.go
 
 # Options for 'bundle-build'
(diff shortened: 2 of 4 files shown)


Makefile.vars.mk
```diff
@@ -1,6 +1,5 @@
 # Current Operator version
-VERSION ?= $(shell date +%Y-%m-%d_%H-%M-%S)
-VERSION := $(VERSION)
+VERSION ?=
```
cimnine, line 2: Are you sure that setting this to empty by default does not affect the `./Makefile`'s `bundle` target?

I argue that [this command on line 106](https://github.com/vshn/k8up/blob/master/Makefile#L106) will break when the `$(VERSION)` variable is suddenly empty:

```
$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
```

Makefile.vars.mk
```diff
@@ -1,6 +1,5 @@
...
 
 # Default bundle image tag
 BUNDLE_IMG ?= controller-bundle:$(VERSION)
```
cimnine, line 5: Since I've been looking into the `$(VERSION)` variable, I've come to believe that the `BUNDLE_IMG` variable is not used anymore.

Makefile.vars.mk
ccremer: We currently aren't using the bundle target at all, it's a leftover from Operator SDK setup.
However, it may become relevant again when we want to enable OLM integration, but I don't know. So I'm unsure whether we should keep the target or get rid of it and recreate it when relevant again (either from git or from a new scaffold).

cimnine:
```
$ make bundle
Warning: default unsupported in CRD version v1beta1, v1 required. Removing defaults.
Warning: default unsupported in CRD version v1beta1, v1 required. Removing defaults.
Warning: default unsupported in CRD version v1beta1, v1 required. Removing defaults.
2021/01/11 10:37:46 Running post-generate in /Users/cimnine/src/vshn/k8up
2021/01/11 10:37:46 Reading file config/crd/apiextensions.k8s.io/v1beta1/backup.appuio.ch_prebackuppods.yaml
2021/01/11 10:37:46 Removed 'protocol' in line 777
2021/01/11 10:37:46 Removed 'protocol' in line 2271
2021/01/11 10:37:46 Writing new file to config/crd/apiextensions.k8s.io/v1beta1/backup.appuio.ch_prebackuppods.yaml
operator-sdk generate kustomize manifests -q
make: operator-sdk: No such file or directory
make: *** [bundle] Error 1
```

cimnine: (Sorry, did not see your comment in the meantime.

cimnine: TBH, I'd toss it. Code that's not used should generally be removed, IMO, because even if it's not used it still need maintenance and may block other development. Like getting rid of the `VERSION` variable here, who's only remaining purpose is the `bundle` task.

ccremer: you need the operator-sdk binary installed for that target.
But yes, let's remove the stuff

cimnine (approved):
I love it when a PR removes more code than it adds 👍

PR #279, opened 2021-01-12
Show partial release notes for PreReleases

ccremer:
## Summary

when creating a tag with 'rc' in its name, this should build a changelog that shows a partial changelog since the last prerelease, and the full one when fully released. That makes it easier to follow changes between prereleases.

This is a suggestion from https://github.com/mikepenz/release-changelog-builder-action/issues/154#issuecomment-758117389, which I quite like and didn't think of earlier.

## Checklist


tobru (approved):
Makes sense, yes :+1:

PR #280, opened 2021-01-12
Fix Application of Global Memory Limit

### LINKED ISSUE #276: Merged resources are stored in the wrong CR object
## Describe the bug

With #175 we have added a feature that allows setting default resources. However, the resources that are merged with the template from the Schedule object and also specified in the job spec end up being stored in the Schedule CR, instead of being updated in the job CR.

this will remove the possibility to alter global default resources for cluster admins after the first reconciliation.

## Additional context

See discussion here #260 

## Logs

The schedule gets updated with the following spec after reconcile:
```yaml
kind: Schedule
metadata:
  name: schedule-test
  namespace: default
spec:
...
  prune:
    resources:
      requests:
        cpu: 250m
    retention:
      keepDaily: 14
      keepLast: 5
    schedule: '*/1 * * * *'
  resourceRequirementsTemplate:
    requests:
      cpu: 250m
```
and prune:
```yaml
kind: Prune
metadata:
  name: prunejob-1609946760
  namespace: default
spec:
  resources:
    requests:
      cpu: 250m
```

## Expected behavior

* `.spec.prune.resources` should be empty in `Schedule`, also the `.spec.resourceRequirementsTemplate` should not be touched

## To Reproduce

1. Schedule Specs
```yaml
---
kind: Schedule
metadata:
  name: schedule-test
spec:
  resourceRequirementsTemplate:
    requests:
      cpu: "250m"
  prune:
    schedule: '*/1 * * * *'
```
2. Set global default memory limit
```bash
kubectl set env deployment/k8up-operator BACKUP_GLOBALMEMORY_LIMIT=76M
```

- Image Version: v1.0.0-rc2
- K8s Version: v1.18
- K8s Distribution: KIND

cimnine:
## Summary

The global memory limit property was accidentally applied as the CPU request. Related to #276.

## Checklist

- [x] Update tests.

### DIFF
--- a/cfg/config.go
+++ b/cfg/config.go
@@ -109,9 +109,9 @@ func (c Configuration) GetGlobalDefaultResources() (res corev1.ResourceRequireme
 	}
 	if r, err := resource.ParseQuantity(c.GlobalMemoryResourceLimit); err == nil && c.GlobalMemoryResourceLimit != "" {
 		if res.Limits == nil {
-			res.Requests = make(corev1.ResourceList)
+			res.Limits = make(corev1.ResourceList)
 		}
-		res.Requests[corev1.ResourceCPU] = r
+		res.Limits[corev1.ResourceMemory] = r
 	}
 	if r, err := resource.ParseQuantity(c.GlobalCPUResourceLimit); err == nil && c.GlobalCPUResourceLimit != "" {
 		if res.Limits == nil {


ccremer:
Nice catch, but does solve the issue that I reported. The issue is that **all merged** resources gets **stored** into the Schedule CR, which shouldn't be.

cimnine:
I've changed the description so that merging this PR does not close #276.

ccremer (approved):

Issue #281, opened 2021-01-12
PreBackupPod CRD cannot be applied with kubectl


ccremer:
## Describe the bug

The PreBackupPod CRD cannot be applied. It fails with error
`spec.preserveUnknownFields: Invalid value: true: must be false in order to use defaults in the schema`

## Additional context

The same error also happens with `--validate=false`
Initial investigation shows it's related to v1 of the CRD: https://github.com/kubernetes-sigs/kubebuilder/issues/933

In our e2e test case, this doesn't happen. It could also relate to the fact that previous versions of K8up (0.x) was installed on the cluster, and now some upgrade of CRD v1beta1 to v1 doesn't quite work out.

~It might be interesting if controller-runtime 0.7.x (see #235 ) already solves this for us.~ it doesn't

## Logs

If applicable, add logs to help explain your problem.
```console
kubectl apply -f https://github.com/vshn/k8up/releases/download/v1.0.0-rc3/k8up-crd.yaml
customresourcedefinition.apiextensions.k8s.io/archives.backup.appuio.ch configured
customresourcedefinition.apiextensions.k8s.io/backups.backup.appuio.ch configured
customresourcedefinition.apiextensions.k8s.io/checks.backup.appuio.ch configured
customresourcedefinition.apiextensions.k8s.io/prunes.backup.appuio.ch configured
customresourcedefinition.apiextensions.k8s.io/restores.backup.appuio.ch configured
customresourcedefinition.apiextensions.k8s.io/schedules.backup.appuio.ch configured
customresourcedefinition.apiextensions.k8s.io/snapshots.backup.appuio.ch configured
The CustomResourceDefinition "prebackuppods.backup.appuio.ch" is invalid: spec.preserveUnknownFields: Invalid value: true: must be false in order to use defaults in the schema
```

## Expected behavior

The CRD can be applied without errors.

## To Reproduce

unclear yet

## Environment (please complete the following information):

- Image Version: v1.0.0-rc3
- K8s Version: v1.18.9-eks-d1db3c
- K8s Distribution: Rancher

PR #283, opened 2021-01-12
Upgrade of Operator SDK to 1.3.0 / Kubebuilder to v3

cimnine:
## Summary

Applied the following advice:

* https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.3.0/
* https://master.book.kubebuilder.io/migration/project/v2_v3.html

Fixes https://github.com/vshn/k8up/issues/235

## Checklist

- [x] Update tests.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -13,9 +13,13 @@ spec:
       labels:
         control-plane: controller-manager
     spec:
+      securityContext:
+        runAsUser: 1001
       containers:
       - name: k8up
         image: quay.io/vshn/k8up:latest
+        securityContext:
+          allowPrivilegeEscalation: false
         resources:
           limits:
             cpu: 300m
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -14,7 +14,7 @@ spec:
         control-plane: controller-manager
     spec:
       securityContext:
-        runAsUser: 1001
+        runAsUser: 65532
       containers:
       - name: k8up
         image: quay.io/vshn/k8up:latest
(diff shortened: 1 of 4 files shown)


ccremer:
According to Kubebuilder migration guide, we can also remove this bit from leader_election_role.yaml:
```yaml
  resources:
  - configmaps/status
  verbs:
  - get
  - update
  - patch
```

config/manager/manager.yaml
```diff
@@ -13,9 +13,13 @@ spec:
       labels:
         control-plane: controller-manager
     spec:
+      securityContext:
+        runAsUser: 1001
```
ccremer, line 17: For us, there's no reason why we couldn't use
```
USER 65532:65532
```
(and this security context) as suggested by the guide, since we don't mount any volumes. Openshift will override the user and group anyway.

ccremer:
So, related to #281 , I tried whether upgrading to controller-runtime to 0.7 will resolve the issue linked. It did not, but I've discovered other findings related to the upgrade.

I made a separate commit for it, but maybe it's probably best to upgrade this in a separate PR, because of the findings that we first should discuss.

ccremer:
Opened discussion here: https://github.com/vshn/k8up/discussions/288

cimnine:
> According to Kubebuilder migration guide, we can also remove this bit from leader_election_role.yaml:

This will become irrelevant when https://github.com/vshn/k8up/pull/289 is merged and when I've rebased, right?

cimnine:
Rebased onto master

ccremer:
I think you can remove my commit about upgrading the controller runtime. This should go into its own PR, e.g. #293 . this was for a test, but I can cherry-pick the commit locally to the other branch.

PR #285, opened 2021-01-13
Rework GitHub workflows

ccremer:
## Summary

* Running e2e tests and image build again when merging a PR to master does not make much sense, since the tests ran in the PR before. Thus skip them in master branch to not waste resources and time
* We still run unit tests in the master branch solely for the purpose of publishing CodeClimate test report. And the docker images are also built anyway since we push master tags.
* Adds more badges
* Enabled branch protection rule so that a PR has to be up to date before merging. This is to avoid merging feature branches with changes that weren't part of the PR and possibly conflict with eachother.

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,42 +1,11 @@
 name: Build
 
 on:
-  push:
-    branches:
-    - master
-    - development
-    tags-ignore:
-    - "*"
   pull_request:
     branches:
     - master
-    - development
 
 jobs:
-  test:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v2
-    - name: Determine Go version from go.mod
-      run: echo "GO_VERSION=$(grep "go 1." go.mod | cut -d " " -f 2)" >> $GITHUB_ENV
-    - uses: actions/setup-go@v2
-      with:
-        go-version: ${{ env.GO_VERSION }}
-    - uses: actions/cache@v2
-      with:
-        path: ~/go/pkg/mod
-        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
-        restore-keys: |
-          ${{ runner.os }}-go-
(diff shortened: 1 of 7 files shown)


cimnine:
> Running e2e tests and image build again when merging a PR to master does not make much sense, since the tests ran in the PR before. Thus skip them in master branch to not waste resources and time

I disagree. When multiple PRs are merged, and they were not rebased onto master before merging them, they could interfere with each other.

.github/workflows/test.yml
```diff
@@ -0,0 +1,35 @@
+name: Test
```
cimnine, line 1: I like that this is in it's own file!

.github/workflows/build.yml
```diff
@@ -1,42 +1,11 @@
 name: Build
```
cimnine, line 1: Maybe this could be re-named to _e2e-test_?

```suggestion
name: E2E-Test
```

... and the _image_ target could be separated as well?

ccremer: yes, this makes sense. I would actually even remove the `image` target, since it's part of e2e build anyway

ccremer:
> > Running e2e tests and image build again when merging a PR to master does not make much sense, since the tests ran in the PR before. Thus skip them in master branch to not waste resources and time
> 
> I disagree. When multiple PRs are merged, and they were not rebased onto master before merging them, they could interfere with each other.

Right. We could set up a [branch protection](https://docs.github.com/en/github/administering-a-repository/about-protected-branches#require-status-checks-before-merging) rule with "Require branch to be up to date". We must then select which checks absolutely must pass (I'd suggest the `lint` only) before merging, and so we can only merge branches that are up to date. This makes a rebase a requirement before merging, but I believe GitHub offers a one-click-rebase-branch button if the PR is outdated.

Would that be ok?

cimnine (approved):

PR #289, opened 2021-01-13
Add K8up viewer and editor roles to default built-in roles

### LINKED ISSUE #284: Viewer and Editor roles are disabled
## Describe the bug

Previous 0.x versions of K8up (helm chart) come with Viewer and Editor roles included.
This is not the case anymore.

## Additional context

* The Viewer and Editor roles are [already there, but not activated in Kustomize](https://github.com/vshn/k8up/tree/master/config/rbac)
* Old versions defined a catch-all ClusterRole like:
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    app.kubernetes.io/instance: k8up
    app.kubernetes.io/managed-by: Tiller
    app.kubernetes.io/name: k8up
    app.kubernetes.io/version: v0.1.10
    argocd.argoproj.io/instance: backup-k8up
    helm.sh/chart: k8up-0.6.1
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
    rbac.authorization.k8s.io/aggregate-to-edit: "true"
  name: k8up-edit
rules:
- apiGroups:
  - backup.appuio.ch
  resources:
  - '*'
  verbs:
  - '*'
```
* Since Helm charts is basing its RBAC objects on Kustomize, the newest Chart PR doesn't include viewer and editor roles either at this time.

## Expected behavior

Viewer and Editor roles should be included again.
However, we could discuss following options:

1. Use the fine-granular roles (for each CRD `archive`, `backup`, `check`. `restore`. `prune`, `schedule` `prebackuppod` its own viewer/editor role)
2. Use the catch-all roles mentioned like before
3. Combine both options 1) and 2) to support easy, but also granular role bindings
4. Split into "user-facing" and "system" roles, e.g. user facing are `schedule`, `prebackuppod` and `restore`(in one role definition), the rest is "system" (also in 1 role definition).
5. [ClusterRole aggregation](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles)

## To Reproduce

1. Specs
```bash
kustomize build config/rbac
```

ccremer:
## Summary

As described in https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles, those will be added automatically, so that any "viewer" or "editor" can view/edit K8up resources also.

Closes #284 

## Checklist


ccremer:
This creates a bunch of clusterroles:
```console
$ k get clusterrole | grep k8up
k8up-archive-editor-role                                               2m3s
k8up-archive-viewer-role                                               2m3s
k8up-backup-editor-role                                                2m3s
k8up-backup-viewer-role                                                2m3s
k8up-check-editor-role                                                 2m3s
k8up-check-viewer-role                                                 2m3s
k8up-manager-role                                                      15m
k8up-prebackuppod-editor-role                                          2m3s
k8up-prebackuppod-viewer-role                                          2m3s
k8up-prune-editor-role                                                 2m3s
k8up-prune-viewer-role                                                 2m3s
k8up-restore-editor-role                                               2m3s
k8up-restore-viewer-role                                               2m3s
k8up-schedule-editor-role                                              2m3s
k8up-schedule-viewer-role                                              2m3s
k8up-snapshot-editor-role                                              2m3s
k8up-snapshot-viewer-role                                              2m3s
```
I'm starting to think that a consolidated viewer and editor role makes more sense than having individual cluster roles.

Makefile
```diff
@@ -38,10 +38,12 @@ run: fmt vet ## Run against the configured Kubernetes cluster in ~/.kube/config
 	go run ./main.go
 
 .PHONY: install
+install: export KUBECONFIG = $(KIND_KUBECONFIG)
```
cimnine, line 41: This would make it impossible to easily install the CRDs in a non-kind cluster using this Makefile.

I suggest to rather create a `kind-install` target:

```Makefile
.PHONY: kind-install
e2e-install: export KUBECONFIG = $(KIND_KUBECONFIG)
e2e-install: install ## Install the CRDs into the e2e test cluster
```

cimnine (commented):
I'm not quite certain I've fully understood the aggregated roles, so I can say little to that other than lgtm.

Makefile
ccremer: argh, this change was here due to a PEBKAC issue, this should not be in here :)

ccremer:
> I'm not quite certain I've fully understood the aggregated roles, so I can say little to that other than lgtm.

It basically means that any user can view or edit K8up resources without having to create explicit rolebindings for that. Provided they have the built-in-kubernetes-default view or edit roles resp.

PR #295, opened 2021-01-15
Add preserveUnknownFields to generated PreBackupPod CRD

### LINKED ISSUE #281: PreBackupPod CRD cannot be applied with kubectl

ccremer:
## Summary

Fixes #281 

This occurs only in PreBackupPods, because there we use the upstream PodSpec that uses properties that are somewhat too advanced

## Checklist

### DIFF
--- a/config/crd/apiextensions.k8s.io/v1/backup.appuio.ch_prebackuppods.yaml
+++ b/config/crd/apiextensions.k8s.io/v1/backup.appuio.ch_prebackuppods.yaml
@@ -8,6 +8,7 @@ metadata:
   creationTimestamp: null
   name: prebackuppods.backup.appuio.ch
 spec:
+  preserveUnknownFields: false
   group: backup.appuio.ch
   names:
     kind: PreBackupPod
--- a/config/crd/apiextensions.k8s.io/v1beta1/backup.appuio.ch_prebackuppods.yaml
+++ b/config/crd/apiextensions.k8s.io/v1beta1/backup.appuio.ch_prebackuppods.yaml
@@ -8,6 +8,7 @@ metadata:
   creationTimestamp: null
   name: prebackuppods.backup.appuio.ch
 spec:
+  preserveUnknownFields: false
   group: backup.appuio.ch
   names:
     kind: PreBackupPod
--- a/generate.go
+++ b/generate.go
@@ -15,7 +15,7 @@ import (
 	"os"
 )
 
-var patchFiles = []string{"v1beta1/backup.appuio.ch_prebackuppods.yaml"}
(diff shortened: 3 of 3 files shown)


config/crd/apiextensions.k8s.io/v1/backup.appuio.ch_prebackuppods.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
   creationTimestamp: null
   name: prebackuppods.backup.appuio.ch
 spec:
+  preserveUnknownFields: false
```
cimnine, line 11: Just to double check: You've meant to set those to `false`, right?

cimnine (approved):

config/crd/apiextensions.k8s.io/v1/backup.appuio.ch_prebackuppods.yaml
ccremer: Yes. If it's set to true, then the cluster wouldn't accept the CRD when applying with kubectl, as described in the linked issue.

PR #296, opened 2021-01-15
Upgrade to controller-runtime 0.8 and K8s 1.20

ccremer:
## Summary

Upgrade to controller-runtime 0.8 and K8s 1.20

## Checklist

- [x] Update tests.


.github/workflows/build.yml
```diff
@@ -42,7 +42,7 @@ jobs:
     strategy:
       matrix:
         include:
-        - kind-node-version: v1.18.8
```
cimnine, line 45: I would suggest to have the minimal version, that supports the leader election, tested as well.

cimnine (approved):

.github/workflows/build.yml
ccremer: We'd need an e2e test that actually verifies leader election, because this can only determined at runtime. Currently we don't have that so a 1.16 cluster is "useless"

tobru (approved):

PR #298, opened 2021-01-15
Adds a test for mergeBackendWithDefaults

cimnine:
## Summary

Adds a test for mergeBackendWithDefaults

## Checklist

- [x] Update tests.


handler/schedule_test.go
```diff
@@ -146,12 +146,161 @@ func TestScheduleHandler_mergeResourcesWithDefaults(t *testing.T) {
...
+				Resources: tt.givenResourceTemplate,
+			}
+			schedule.mergeResourcesWithDefaults(res)
+			assert.Equal(t, tt.expectedTemplate, res.Resources)
+		})
+	}
+}
+
+func TestScheduleHandler_mergeBackendWithDefaults(t *testing.T) {
+	tests := []struct {
```
ccremer, line 162: just a minor thing to be ready for our reference tests, would you mind changing this to a struct and move the `name` property of each struct to the key of the map?
```suggestion
	tests := map[string]struct {
```

handler/schedule_test.go
```diff
@@ -146,12 +146,161 @@ func TestScheduleHandler_mergeResourcesWithDefaults(t *testing.T) {
...
+		name                 string
+		globalS3Bucket       string
+		givenScheduleBackend v1alpha1.Backend
+		givenResourceBackend v1alpha1.Backend
+		expectedBackend      v1alpha1.Backend
+	}{
+		{
+			name: "Given_NoGlobalDefaults_And_NoScheduleDefaults_When_Spec_Then_UseSpec",
+			givenResourceBackend: v1alpha1.Backend{
```
ccremer, line 171: I suggest to create a small helper function that improves readability
```suggestion
			givenResourceBackend: createBackend("https://resource-url", "resource-bucket")
```

handler/schedule_test.go
```diff
@@ -21,53 +21,53 @@ func TestScheduleHandler_mergeResourcesWithDefaults(t *testing.T) {
...
-		resources                   v1.ResourceRequirements
-		expected                    v1.ResourceRequirements
+		givenScheduleTemplate       v1.ResourceRequirements
+		givenResourceTemplate       v1.ResourceRequirements
+		expectedTemplate            v1.ResourceRequirements
 	}{
 		{
-			name:     "Given_NoGlobalDefaults_And_NoScheduleDefaults_When_NoSpec_Then_LeaveEmpty",
-			expected: v1.ResourceRequirements{},
+			name:             "Given_NoGlobalDefaults_And_NoScheduleDefaults_When_NoSpec_Then_LeaveEmpty",
```
ccremer, line 29: Can you also move this test to the `map[string]struct` layout?

cimnine:
Rebased and Refactored ✅

handler/schedule_test.go
```diff
@@ -90,72 +84,149 @@ func TestScheduleHandler_mergeResourcesWithDefaults(t *testing.T) {
...
+			expectedBackend:      newS3Backend("https://schedule-url", "schedule-bucket"),
+		},
+		"Given_NoGlobalDefaults_And_ScheduleDefaults_When_Spec_Then_UseSpec": {
+			givenScheduleBackend: newS3Backend("https://schedule-url", "schedule-bucket"),
+			givenResourceBackend: newS3Backend("https://resource-url", "resource-bucket"),
+			expectedBackend:      newS3Backend("https://resource-url", "resource-bucket"),
+		},
+		"Given_GlobalDefaults_And_NoScheduleDefaults_When_NoSpec_Then_UseGlobalDefaults": {
+			globalS3Bucket: "global-bucket",
+			givenScheduleBackend: v1alpha1.Backend{
```
ccremer, line 152: Is the `newS3Backend` function not usable here? en empty string "" is equal to not specifying in the struct, provided the property is not a pointer

handler/schedule_test.go
cimnine: I've applied the same for the other test and the `ResourceList` resource.

ccremer (approved):

handler/schedule_test.go
cimnine: It certainly is, but my regex to did not catch this case 😅

PR #310, opened 2021-01-20
Introduce Integration Tests using envtest

cimnine:
## Summary

An attempt to implement https://github.com/vshn/k8up/issues/274

## Todo

- [x] Github Action to trigger integration tests
- [x] Document how to run integration tests in an IDE
- [x] Re-integrate fixups

## Checklist

- [x] Update tests.


controllers/suite_test.go
```diff
@@ -1,93 +0,0 @@
...
-
-func TestAPIs(t *testing.T) {
-	RegisterFailHandler(Fail)
-
-	RunSpecsWithDefaultAndCustomReporters(t,
-		"Controller Suite",
-		[]Reporter{printer.NewlineReporter{}})
-}
-
-var _ = BeforeSuite(func(done Done) {
```
ccremer, line 40: I'm interested on a proposal how to replicate this suite in plain testing.

ccremer (commented):
I'm a bit skeptical about this approach. at least for the specific function.
I see problems in the maintainability of test case itself, but that may be a design flaw in the restore logic itself.
If we are asserting every env var, volume name, image etc. then I'd suggest to refactor the `buildRestoreObject` in a way that allows more granular testing. That results in the build function to be more composed of subfunctions. Maybe even a Builder pattern could be of use here. e.g. `newJobBuilder().WithVolume(...).WithEnvs(...).Build()`

But also some aspects of the function should be done in an actual integration test with a server backing it. e.g. OwnerReference. With a unit test resp. programmatic approach we assume that the server behaves as we write our expectation logic. However our assumption can be wrong and we won't see important side effects. The intent behind setting owner reference is that the resource is being garbage collected when the parent resource is being deleted. As such, we should test whether that works as expected, since a test like that automatically verifies that all properties of the owner reference are correct.

Generally speaking, I think this particular test should be separated into more granular unit tests, but that may require some minor design changes/refactors. An integration test or even e2e test verifies if all parts are glued together correctly as we can focus more on the higher level business functions.

WDYT?

executor/restore_test.go
```diff
@@ -2,194 +2,181 @@ package executor
...
-	Expect(err).NotTo(HaveOccurred())
+var yes = true
 
-	close(done)
-})
+func Test_buildRestoreObject(t *testing.T) {
+	err := batchv1.AddToScheme(testScheme)
+	if err != nil {
+		t.Errorf("Failed to register batchv1 scheme: %s", err)
+	}
```
ccremer, line 45: This can be simplified with `require.NoError(batchv1.AddToScheme(testScheme))`
`require` immediately aborts if there's an error.

executor/restore_test.go
```diff
@@ -2,194 +2,181 @@ package executor
...
-		Expect(err).To(BeNil())
-		Expect(j).NotTo(BeNil())
-		Expect(j).To(PointTo(c.JobMatcher))
-	},
-		Entry("builds a job object for S3 restore", buildRestoreObjectCase{
-			Restore: &k8upv1alpha1.Restore{
-				Spec: k8upv1alpha1.RestoreSpec{
-					RestoreMethod: &k8upv1alpha1.RestoreMethod{
-						S3: &k8upv1alpha1.S3Spec{
+	tests := map[string]struct {
```
corvus-ch, line 52: I prefer to have the test tables defined on the package level for two reasons:

1. It makes the test function easier to read. The test table can be a uge distraction herer.
2. The test data can be used in more than one test function. I actually used this quite a lot (for example in two way data conversions but there are also other cases this comes in handy).

corvus-ch (commented):
For an integration test, we need to ask the question: what do we need the server for? So an integration test should only test things that can not be tested in unit tests. I have the impression, that we have things mixed up here. Things should probably be splitt up.

cimnine:
I've now migrated the existing test to a unit test style test and started the new foundation for the future integration tests, of which I intent to implement the one for the Restore resource.

cimnine:
A basic integration test for `Restore` is now implemented.

integration/restore_test.go
```diff
@@ -0,0 +1,137 @@
...
+	return result
+}
+
+func (r *RestoreTestSuite) expectAJobEventually() {
+	for i := 0; i < 10; i++ {
+		if i > 0 {
+			r.T().Logf("⏱ No job after %d seconds", i)
+			time.Sleep(time.Second)
+		}
+
```
ccremer, line 125: I think we're going to use this sort of timeouts multiple times.
I suggest to add a timeout function to the base test suite by making use of `context.WithTimeout(...)` so we can just call like `r.waitForResource(r. T(), &batchv1.Job, 20s)`

integration/restore_test.go
```diff
@@ -0,0 +1,137 @@
...
+			RestoreMethod: &k8upv1a1.RestoreMethod{
+				S3: &k8upv1a1.S3Spec{
+					Bucket:   "backups",
+					Endpoint: "https://s3-endpoint.local",
+				},
+			},
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      RestoreName,
+			Namespace: NS,
```
ccremer, line 47: I recommend to use different namespace for each test. can even be generated ones, main point is that they don't interfere with eachother with sideeffects or resources we might not know about at programming time.

integration/restore_test.go
cimnine: Unfortunately this is not a straight forward approach. The reason is that `r.Client.List` expects an `ObjectList`. But this type does not provide a method to access the individual `.Items`. Those are only on the concrete types, i.e. `JobList` or `RestoreList` for example.

However, I now moved the logic for managing the repetition and the timeout to `common_test.go` and the `EnvTestSuite` type. The specific implementation for the assertion must be provided to this logic (i.e. `RepeatedAssert`) by passing a function. That function is then invoked every so often to perform the actual assertion.

integration/restore_test.go
cimnine: I'll look into that in the afternoon.

But I believe this might be something that we can delay until it becomes a problem. Because currently, each test suite gets it's own envtest. And – as I understand – every individual test suite is executed sequentially. So if every test of each suite properly cleans up after itself, we won't have a problem soon.

ccremer: I was referencing to each test within the same suite so that they're side effects free within the suite. It might be enough to just use the test name as the namespace

.github/workflows/integration.yml
```diff
@@ -0,0 +1,36 @@
+name: Integration Test
```
ccremer, line 1: Instead of adding a new test suite, why not just use the existing Test workflow that just executes `integration-test` instead of `test`? AFAIK, the normal unit tests also run, and so we get only 1 coverage report.

Makefile
```diff
@@ -23,10 +23,21 @@ test: fmt vet ## Run tests
...
+integration-test: generate $(testbin_created) ## Run integration tests with envtest
+	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || \
+		curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/master/hack/setup-envtest.sh
+	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; \
+		fetch_envtest_tools $(ENVTEST_ASSETS_DIR); \
+		setup_envtest_env $(ENVTEST_ASSETS_DIR); \
+		go test -tags=integration ./integration -coverprofile cover.out
+
+integration-test-debug: export KUBEBUILDER_ATTACH_CONTROL_PLANE_OUTPUT = true
+integration-test-debug: integration-test ## Run integration tests with envtest debug output enabled
```
ccremer, line 40: Is this really necessary? Why can't we use `make integration-test -e DEBUG_INTEGRATION_TEST=true` with default value being false?

integration/common_test.go
```diff
@@ -0,0 +1,120 @@
...
+	err := ts.Client.List(ts.Ctx, list)
+	assert.NoError(ts.T(), err)
+
+	ts.T().Logf("Deleting %d Jobs", len(list.Items))
+
+	for _, j := range list.Items {
+		err := ts.Client.Delete(ts.Ctx, &j) // DeleteAllOf seems not implemented in envtest
+		assert.NoError(ts.T(), err)
+	}
+}
```
ccremer, line 93: Do we really need to delete resources? If the tests are executed per namespace, and we remove the namespace, all resources in that namespace get deleted automatically

And even if there are reasons, I would prefer a generic solution that works for all kinds of resources. The function could accept a ListInterface, since ObjectList from controller-client basically aliases just that.

integration/restore_test.go
```diff
@@ -0,0 +1,123 @@
+// +build integration
+
+package integration
```
ccremer, line 3: I would actually move this test closer to restore package as well and name it `restore_integration_test.go`.
To me, integration tests feel more closely to unit tests than e2e tests. The only difference is the API server

integration/common_test.go
cimnine: The envtest does not implement garbage collection. Deleting the NS will not remove resources associated with that NS.

ccremer: Ok. But even then, we can leave them around. Less code to maintain

integration/restore_test.go
cimnine: I'll move them to the controller package then, since that's the entrypoint to the reconciliation.

controllers/suite_test.go
cimnine: It's now not plain testing, but a [testify suite](https://godoc.org/github.com/stretchr/testify/suite)

integration/restore_test.go
cimnine: I've implemented it, but since garbage collection is not running in envtest it actually provides little value. When removing a NS, it does not remove the respective resources. Same goes for owner references, btw.

docs/modules/ROOT/pages/explanations/ide.adoc
```diff
@@ -0,0 +1,31 @@
...
+. Then try to run a specific integration test.
+  You must either run the tests of a single directory or file – or run a specific test method.
+  The execution will fail at first – that's expected.
+. Now go to _Run – Edit Configurations..._ and find the automatically created run configuration for the test.
+. Edit the _Working directory:_ field.
+  Set it to the same directory as the integration test lies which you want to test.
+. Edit the _Environment:_ field.
+  It should contain `KUBEBUILDER_ASSETS=../testbin/bin`.
+. Edit the _Go tools arguments:_ field.
+  It should contain `-tags=integration`.
```
ccremer, line 21: Can we make this even better?
Can we set this programmatically? see https://godoc.org/sigs.k8s.io/controller-runtime/pkg/envtest#Environment (`BinaryAssetsDirectory`)

docs/modules/ROOT/pages/explanations/ide.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+Now try to re-run the test.
+It should now succeed.
+You will see test results in your IDE and be able to use the integrated debugger as well.
+
+IMPORTANT: Remember that you need to re-run `make integration-test` when you ran `make clean` or when you deleted the `./testbin` directory.
+
+NOTE: If you're in doubt about your configuration, check whether `make clean integration-test` succeeds.
+  If it does, you've broken your IDE configuration.
+  Otherwise, you've broken something else.
```
ccremer, line 31:
```suggestion
[NOTE]
====
If you're in doubt about your configuration, check whether `make clean integration-test` succeeds.
If it does, you've broken your IDE configuration, otherwise you've broken something else.
====
```

controllers/restore_integration_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (r *RestoreTestSuite) SetupTest() {
+	r.RestoreName = "restore-integration-test"
+}
+
+func (r *RestoreTestSuite) TearDownTest() {
+	r.CleanUpJobs()
+	r.cleanUpRestores()
+	r.CleanUpNamespaces()
+}
```
ccremer, line 51: Please help me understand. Why do we cleanup resources, if each test within a suite is executed in a different namespace, and each suite has its own API server process that lives only as long as the tests run? Is there a reason why we add so much code to cleanup resources that we may no longer care about anyway? Namespaces provide enough isolation between tests.

We could remove 30% of the testing code.

Makefile
```diff
@@ -51,9 +62,11 @@ deploy: generate ## Deploy controller in the configured Kubernetes cluster in ~/
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
 .PHONY: generate
-generate: ## Generate manifests e.g. CRD, RBAC etc.
+generate: $(testbin_created) ## Generate manifests e.g. CRD, RBAC etc.
 	@CRD_ROOT_DIR="$(CRD_ROOT_DIR)" go generate -tags=generate generate.go
 	@rm config/*.yaml
+	@cp -r config/crd/apiextensions.k8s.io $(TESTBIN_DIR)/
+	@rm $(TESTBIN_DIR)/apiextensions.k8s.io/{v1,v1beta1}/kustomiz*.yaml || true
```
ccremer, line 69: I recommend restructuring the CRD dirs a bit:
```
/config/crd/apiextensions.k8s.io/{v1,v1beta1}/kustomize.yaml
/config/crd/apiextensions.k8s.io/{v1,v1beta1}/bases/backup.appuio.ch_*.yaml
```
So we can tell envtest to search CRDs in `/config/crd/apiextensions.k8s.io/v1/bases`, where only CRD yaml files are stored. That way we don't need to copy files around.

controllers/restore_integration_test.go
cimnine: Now that each test is in it's own NS, this is certainly a valid argument. I like to clean up after a test, so that it's not possible to have side-effects in later test. But this is certainly not a _must_ anymore.

config/crd/apiextensions.k8s.io/v1/kustomization.yaml
```diff
@@ -2,40 +2,14 @@
...
 resources:
-- backup.appuio.ch_schedules.yaml
-- backup.appuio.ch_backups.yaml
-- backup.appuio.ch_restores.yaml
-- backup.appuio.ch_archives.yaml
-- backup.appuio.ch_checks.yaml
-- backup.appuio.ch_prunes.yaml
-- backup.appuio.ch_snapshots.yaml
-- backup.appuio.ch_prebackuppods.yaml
-# +kubebuilder:scaffold:crdkustomizeresource
```
ccremer, line 13: I'd recommend to keep this scaffold comment, in case we add more CRDs

config/crd/apiextensions.k8s.io/v1beta1/kustomization.yaml
```diff
@@ -2,40 +2,14 @@
...
 resources:
-- backup.appuio.ch_schedules.yaml
-- backup.appuio.ch_backups.yaml
-- backup.appuio.ch_restores.yaml
-- backup.appuio.ch_archives.yaml
-- backup.appuio.ch_checks.yaml
-- backup.appuio.ch_prunes.yaml
-- backup.appuio.ch_snapshots.yaml
-- backup.appuio.ch_prebackuppods.yaml
-# +kubebuilder:scaffold:crdkustomizeresource
```
ccremer, line 13: same here

docs/modules/ROOT/pages/explanations/ide.adoc
```diff
@@ -0,0 +1,35 @@
...
+  It should contain `-tags=integration`.
+
+Now try to re-run the test.
+It should now succeed.
+You will see test results in your IDE and be able to use the integrated debugger as well.
+
+[IMPORTANT]
+====
+Remember that you need to re-run `make integration-test` when you ran `make clean` or when you deleted the `./testbin` directory.
+====
```
ccremer, line 28: FYI: One-line sentences could be left as `IMPORTANT: ...` but no hard feelings here.
In the case below it makes sense to keep the block admonition

ccremer (approved):

controllers/restore_integration_test.go
```diff
@@ -0,0 +1,107 @@
+// +build integration
+
+package controllers
```
corvus-ch, line 3: Generaly, I prefer to use `_test` packages for tests. I deem this even more important for integration tests. I ensures, that only public things of the subject under test can be used.

corvus-ch (approved):
Looking good for now.

controllers/restore_integration_test.go
ccremer: I find it a bit hard to generalize this idea. Sometimes I need to test package internal helper functions to keep test scope small without having to expose them. For me, same package visibility is fine as long as the file name suffix stays `_test.go`

cimnine: We've just discussed this in person, actually. I argued this as well. But I believe for an integration test it is a fair argument to make that only public APIs should be invoked.

PR #311, opened 2021-01-20
Change default backup image to quay.io/vshn/wrestic

ccremer:
## Summary

The backup image with hardcoded IP and example repository does not bring any value.
For generic purpose, the image being used is wrestic anyway.

Technically, this is a breaking change. However, since all users were required to set this property anyway, this should not have a major impact.

## Checklist

- [x] Update tests.


tobru (approved):

Kidswiss (approved):
LGTM

PR #312, opened 2021-01-21
Avoid name clash by using scheduler name as prefix and random suffix

### LINKED ISSUE #306: could not trigger k8up newJobs: Backup job already exists
## Describe the bug

When reconciling the Schedule object after a Pod restart, K8up cannot reconcile the Schedule to finish, as it encounters an error.

## Additional context

I was trying to upgrade K8up 0.x to 1.x. After restart of K8up, this error below occurred.
Note: I expected the example schedule to have an updated status field with the effective schedule from the check job. But after starting K8up 1.0.0-rc4, this doesn't happen.

## Logs

```console
2021-01-20T09:57:00.000+0100	INFO	controllers.Schedule	running schedule for	{"schedule": "default/schedule-global", "job": "backup"}
2021-01-20T09:57:00.000+0100	INFO	controllers.Schedule	running schedule for	{"schedule": "default/schedule-test", "job": "backup"}
2021-01-20T09:57:00.007+0100	INFO	controllers.Backup	adding job to the queue	{"backup": "default/backupjob-1611133020"}
2021-01-20T09:57:00.007+0100	ERROR	controllers.Schedule	could not trigger k8up newJobs	{"schedule": "default/schedule-test", "name": "default/backupjob-1611133020", "error": "backups.backup.appuio.ch \"backupjob-1611133020\" already exists"}
github.com/go-logr/zapr.(*zapLogger).Error
	/home/runner/go/pkg/mod/github.com/go-logr/zapr@v0.2.0/zapr.go:132
github.com/vshn/k8up/scheduler.(*Scheduler).createObject
	/home/runner/work/k8up/k8up/scheduler/scheduler.go:172
github.com/vshn/k8up/scheduler.(*Scheduler).getScheduleCallback.func1
	/home/runner/work/k8up/k8up/scheduler/scheduler.go:120
github.com/robfig/cron/v3.FuncJob.Run
	/home/runner/go/pkg/mod/github.com/robfig/cron/v3@v3.0.1/cron.go:136
github.com/robfig/cron/v3.(*Cron).startJob.func1
	/home/runner/go/pkg/mod/github.com/robfig/cron/v3@v3.0.1/cron.go:312
```

## Expected behavior

The schedule should either update the backup job, or delete and recreate

## To Reproduce

2. Commands
```bash
make kind-setup
export KUBECONFIG=${PWD}/testbin/bin/kind-kubeconfig-v1.20.0
kubectl create ns k8up-system
helm -n k8up-system install k8up appuio/k8up
kubectl apply -f config/samples/k8up_v1alpha1_prebackuppod.yaml
# wait a few mins to generate jobs
helm uninstall k8up
make crd
kubectl apply -f k8up-crd.yaml
kustomize build config/default | kubectl apply -f -
# now watch the logs
```

- Image Version: v1.0.0-rc4
- K8s Version: v1.20
- K8s Distribution: KIND

ccremer:
## Summary

* When multiple schedules are created in the same namespace, the names could clash if the are scheduled in the same second.
* It replaces the unix timestamp suffix with a 5-char randomized string in the job name. Since the creation date is in the metadata anyway, no data is lost. This is similar how replicaSets work

Closes #306 

## Checklist

- [x] Update tests.

### DIFF
--- a/job/job.go
+++ b/job/job.go
@@ -26,8 +26,11 @@ const (
 	// K8uplabel is a label that is required for the operator to differentiate
 	// batchv1.job objects managed by k8up from others.
 	K8uplabel = "k8upjob"
-	// K8upExclusive is needed to determine if a given job is consideret exclusive or not.
+	// K8upExclusive is needed to determine if a given job is considered exclusive or not.
 	K8upExclusive = "k8upjob/exclusive"
+
+	// K8upTypeLabel is the label key that identifies the job type.
+	K8upTypeLabel = "k8up.io/type"
 )
 
 // Config represents the whole context for a given job. It contains everything
@@ -70,7 +73,8 @@ func GetGenericJob(obj Object, config Config) (*batchv1.Job, error) {
 			Name:      obj.GetMetaObject().GetName(),
 			Namespace: obj.GetMetaObject().GetNamespace(),
 			Labels: map[string]string{
-				K8uplabel: "true",
+				K8uplabel:     "true",
+				K8upTypeLabel: obj.GetType().String(),
 			},
 		},
 		Spec: batchv1.JobSpec{
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -4,13 +4,14 @@ import (
 	"errors"
 	"fmt"
 	"sync"
-	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/robfig/cron/v3"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/utils/strings"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
@@ -152,7 +153,7 @@ func (s *Scheduler) RemoveSchedules(namespacedName types.NamespacedName) {
 
 func (s *Scheduler) createObject(jobType k8upv1alpha1.JobType, namespace string, obj ObjectCreator, config job.Config) {
 
-	name := fmt.Sprintf("%sjob-%d", jobType, time.Now().Unix())
+	name := generateName(jobType, config.Obj.GetMetaObject().GetName())
 
 	rtObj := obj.CreateObject(name, namespace)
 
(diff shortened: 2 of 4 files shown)


scheduler/scheduler.go
```diff
@@ -174,6 +175,13 @@ func (s *Scheduler) createObject(jobType k8upv1alpha1.JobType, namespace string,
 
 }
 
+func generateName(jobType k8upv1alpha1.JobType, prefix string) string {
+	lenRandom := 8
```
cimnine, line 179: The job and pod names have a 5 character suffix. I believe that's enough.

scheduler/scheduler_test.go
```diff
@@ -107,3 +107,29 @@ func TestScheduler_SyncSchedules(t *testing.T) {
...
+			jobType:        k8upv1alpha1.ArchiveType,
+			prefix:         "my-schedule-with-a-really-long-name-that-could-clash-with-max-length",
+			expectedPrefix: "my-schedule-with-a-really-long-name-that-could-archive-",
+		},
+	}
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			name := generateName(tt.jobType, tt.prefix)
+			assert.Contains(t, name, tt.expectedPrefix)
+			assert.LessOrEqual(t, len(name), 63)
```
cimnine, line 132: I suggest to add an assert for the total length as well, i.e. `len(tt.expectedPrefix)+8`.

cimnine (approved):

scheduler/scheduler_test.go
ccremer: I'm not sure what you mean? the return value is the full name, and that should not exceed total length of 63 chars, which includes prefix, type and suffix

cimnine: As I read the test:
- The test tests that the prefix is as expected
- It tests that the allowed maxima is not exceeded

It does not test however whether the random suffix is actually generated. If the suffix were of length zero, it would still pass the test.
Therefore I suggest a test that checks whether the generated name's length is equal to the length of the prefix plus the expected length of the suffix, i.e. `assert.Equal(t, len(name), len(expectedPrefix)+8`.

job/job.go
```diff
@@ -70,7 +70,8 @@ func GetGenericJob(obj Object, config Config) (*batchv1.Job, error) {
 			Name:      obj.GetMetaObject().GetName(),
 			Namespace: obj.GetMetaObject().GetNamespace(),
 			Labels: map[string]string{
-				K8uplabel: "true",
+				K8uplabel:      "true",
+				"k8up.io/type": obj.GetType().String(),
```
cimnine, line 74: I suggest to extract this label into a const as well, as are `K8uplabel` and `K8upExclusive`.

scheduler/scheduler_test.go
ccremer: ah, right

job/job.go
ccremer: I was too lazy :D

cimnine (approved):

Kidswiss (commented):
LGTM

The addition of the timestamp was always intended to introduce a bit of randomness in the pod names. Similar how k8s native cronjobs handle things.

But no hard feelings in replacing this with something better.

Issue #274, opened 2021-01-11
Migrate Ginkgo tests to plain Go testing


ccremer:
## Summary

**As** K8up developer\
**I want** to migrate the test cases done with Ginkgo over to plain Go `testing`\
**So that** I do not need to know another framework besides the default.

## Context

* The current Ginkgo test cases were set up by Operator SDK and are mainly used with Kubebuilder's `envtest` library
* `envtest` is the integration test layer to test K8s controllers and operators by using an in-memory K8s API server. Should more resemble an actual cluster than the K8s fake client.
* It's been concluded that we should replace Ginkgo since its BDD syntax has low to no support for common IDEs and the value it provides is not deemed worthy to continue with it.

## Out of Scope

* e2e tests
* Adding more tests. Just migrate the necessary ones, even if CodeClimate might complain.

## Further links

* Discussion: https://github.com/vshn/k8up/discussions/231

## Acceptance criteria

* Given K8up code base, when I want to contribute PR with unit or integration test, then I can do that with plain Go `testing` library
* Given local git checkout of K8ups code base, when I want to integrate and run K8up's integration tests in my IDE, then I can read the K8up documentation that has instructions for Goland and VS code.
* Given GH action workflows, when a PR is created or updated, then run the test suite with integration tests enabled (currently, only unit tests run)

## Implementation Ideas

* Since Go's `testing` executes each test case in its own PID, there is no possibility to have "BeforeAll" and "AfterAll". We could work around this with table-driven test -> few test functions, but calling `t.Run()` multiple times within that function

PR #330, opened 2021-01-25
Convert executor/restore_test.go to unit test

cimnine:
## Summary

`executor/restore_test.go` was basically a huge unit test all along. It has now been taken apart and re-implemented as an actual unit test in preparations for #274.

## Checklist

- [x] Update tests.


executor/restore_test.go
```diff
@@ -2,306 +2,270 @@ package executor
...
 }
 
-func newConfig() *job.Config {
-	cfg := job.NewConfig(context.TODO(), nil, nil, &k8upv1alpha1.Restore{}, scheme, "")
-	return &cfg
+func assertVolumes(t *testing.T, ex PVCExpectation, volumes []corev1.Volume) {
+	if !ex.Expected {
+		assert.Len(t, volumes, 0)
+		return
+	}
```
ccremer, line 223: I don't see any tests now that use this flag. Maybe we can remove it?

executor/restore_test.go
```diff
@@ -2,306 +2,270 @@ package executor
...
+			e := NewRestoreExecutor(*tt.GivenJobConfig)
+			volumes, mounts := e.volumeConfig(tt.GivenResource)
+
+			assertVolumes(t, tt.ExpectedPVC, volumes)
+			assertVolumeMonts(t, tt.ExpectedVolumeMount, mounts)
 		})
 	}
+}
+
+func assertVolumeMonts(t *testing.T, ex VolumeMountExpectation, mounts []corev1.VolumeMount) {
```
ccremer, line 206:
```suggestion
func assertVolumeMounts(t *testing.T, ex VolumeMountExpectation, mounts []corev1.VolumeMount) {
```

executor/restore_test.go
```diff
@@ -2,306 +2,270 @@ package executor
...
+			ExpectedPVC: PVCExpectation{
+				Expected:  true,
+				ClaimName: "test",
+				ReadOnly:  false,
+			},
+			ExpectedVolumeMount: VolumeMountExpectation{
+				Expected: true,
+				Name:     "test",
+				Path:     "/restore",
+			},
```
ccremer, line 191: Maybe we can even remove the struct and assert directly in the test?
```suggestion
			ExpectedPVCClaimName: "test",
			ExpectedVolumeMountName: "test",
			ExpectedVolumeMountPath: "/restore"
```

ccremer (commented):
I like it! The tests are now much more readable

executor/restore_test.go
cimnine: `givenS3RestoreResource_whenVolumeConfig_expectNoPVCAndNoMount` implicitly sets them `false`. I've now made this explicit.

executor/restore_test.go
cimnine: I find the structure cleaner, especially when passing them to `assertVolumeMounts` and `assertVolumes`.

cimnine:
I've implemented the feedback. There have not been any other changes.

ccremer (approved):

Issue #214, opened 2020-12-18
[Feature] Deduplicate Check and Prunes for the same backup repository


ccremer:
## Summary

**As** K8up user\
**I want** to deduplicate Jobs that target the same repository \
**So that** exclusive Jobs are not run excessively

## Context

Check and Prunes are Restic Jobs that need exclusive access to the backend repository: Only one job can effectively run at the same time. However, multiple backups can target the same Restic repository.

The operator should deduplicate prune jobs that are managed by a smart schedule. So for example if there are multiple schedules with `@daily-random` prunes to the same S3 endpoint the scheduler should only register one of these. 

But if the prunes have explicit cron patterns like `5 4 * * *` and `5 5 * * *` they should NOT be deduplicated. This will ensure maximum flexibility if for some reason a user explicitly wants multiple prune runs.

## Out of Scope

* 

## Further links

* #116 

## Acceptance criteria

*Given* a `Check` or `Prune` object\
*When* the same non-standardized predefined cron syntax is specified that targets the same backup repository\
*Then* ignore the duplicated schedule of the same job type

## Implementation Ideas

* If applicable, shortly list possible implementation ideas

cimnine:
Some ideas to spin:

- Deduplicate when scheduling, i.e. before adding a new entry to the internal cron
- Deduplicate when creating job objects,
  - by checking an internal reference
  - by querying k8s controller if a similar object was already scheduled
  - combination of them.

ccremer:
This ain't exactly easy. I started implementing, but soon discovered that
* we need to check for deduplication first in all other Schedule CRs, as only those contain the `@...` schedule definitions.
* also with other schedules, we check if the backend is "same" -> equal comparison implementations necessary
* then, we don't know whether the checks or prunes of the **other** schedules have already been deduplicated or not, thus we **also** need to verify if they're not already in the cron scheduler.

So, back to drawing board:
![image](https://user-images.githubusercontent.com/12159026/105507114-1a0b5980-5ccb-11eb-9168-50f907d12f35.png)

The graphic is in the PR for editing, though it's not intended to stay there in that form.

Kidswiss:
One thought I just had:

If we do this, should we provide a stable deduplication for this? Example:

[...]

PR #333, opened 2021-01-26
new CRD: EffectiveSchedule

### LINKED ISSUE #338: Deduplication pt 1: Generate effective schedules in its own custom resource
## Summary

**As** K8up user\
**I want** persistent schedules that were generated with random schedule specs\
**So that** the effective schedules stay stable across K8up restarts.

## Context

As an outcome of #214 , we decided to implement deduplication in 2 steps. The current K8up design with its internal state handling is not ready for deduplication when it comes to smart/randomized schedules.

We decided to split deduplication feature into 2 parts:
Part 1: Introduce a new CRD that stores the generated schedule definition coming from the `Schedule` resources. (only when `@x-random` is used). There is no further logic/controller behind this CR (like `PrebackupPod`), it's just a means to store information.
Part 2: Implement the deduplication logic with the new CRD will hold all references to `Schedules`, but only 1 such schedule is added to the internal scheduler, achieving deduplication.

Since part 1 will be the new location where the effective Schedule is stored, it would be a "breaking" change because currently this information is in the Status field of the `Schedule`. however, since that feature is not yet released in K8up v1, it's not breaking unless we release K8up v1 before this new CRD is also part of v1.0.

## Out of Scope

* Deduplication logic. While we design the new CRD with deduplication in mind, we explicitly don't deduplicate in part 1.

## Further links

* See [PR](#333 )
* [Discussion](#214 )

## Acceptance criteria

For the purpose of naming, the new CRD is named `EffectiveSchedule`, but may be changed for a better name.

* *Given* a `Schedule` with `@...-random` specification ,*When* that resource is being reconciled, *Then* a new `EffectiveSchedule` CR is created in the same namespace as the Operator currently runs in.
* *Given* an `EffectiveSchedule` that holds the reference to a specific job type (backup, prune, etc) in a `Schedule`, *When* changing the schedule def from `@...-random` to standard syntax, *Then* remove the reference to the job in `EffectiveSchedule`
* *Given* an  `EffectiveSchedule`, *When* no more references to a job in a `Schedule` is needed, *Then* delete the `EffectiveSchedule`

ccremer:
## Summary

As discussed in #214 , step 1 for Deduplication
* Adds new CRD `EffectiveSchedule`
  * Reverts adding status field to `Schedule`
* Adds integration tests for `Schedule`
  * Adds more integration test utility functions
* Adds new **required** config var: BACKUP_OPERATOR_NAMESPACE
  * This is a breaking change, since the startup fails if left empty. But since we can default this to installation namespace in Kustomize, Helm and Commodore, this shouldn't be too much an issue.
* Updates documentation
  * Removes status example
  * Updates K8up architecture diagram

Closes #338 

## Checklist

- [x] Update tests.


ccremer:
RFC first round

api/v1alpha1/effectiveschedule_types.go
```diff
@@ -0,0 +1,59 @@
...
+		// smart schedules.
+		EffectiveSchedules []JobRef `json:"effectiveSchedules,omitempty"`
+	}
+
+	// JobRef represents a reference to a job in a Schedule object
+	JobRef struct {
+		Name      string             `json:"name,omitempty"`
+		Namespace string             `json:"namespace,omitempty"`
+		JobType   JobType            `json:"jobType,omitempty"`
+		Schedule  ScheduleDefinition `json:"schedule,omitempty"`
```
cimnine, line 45: Why do we need the schedule definition, if we have the job ref?

ccremer: This is the generated, effective Schedule for that job type, e.g. `@hourly-random` translated to `23 * * * *`.
But maybe `EffectiveSchedule` is a better name?

cimnine: Makes sense. Maybe GeneratedSchedule to not confuse it with the EffectiveSchedule definition?

cimnine:
I did not have a `k8up-system` NS, but applied the new sample schedule:

```
```

https://gist.github.com/cimnine/4ff4f16ee715d44020b96c984062caca

a) IMO, this must not lead to a panic
b) I believe that k8up must check when it starts whether there is the k8up-system namespace (or whatever is configured) and if that namespace does not exist, it should `exit(1)` immediately.

An alternative may be that k8up (optionally, but by default) tries to create it's namespace.

ccremer:
Hmm, right.
I agree on the Panic, that should be catched.
Yes, the default is hardcoded `k8up-system` atm. But maybe it's better to leave it internally empty, and require the admin to explicitly set it? Then we can exit early if the setting is empty. In case of the Helm chart and Kustomize, we can set a default. However, checking whether the namespace exists at startup doesn't cover the case where the namespace can be deleted after the startup, so stuff may still break.

handler/schedule.go
```diff
@@ -206,6 +233,80 @@ func (s *ScheduleHandler) getRandomSchedule(jobType k8upv1alpha1.JobType, origin
...
+	})
+	s.effectiveSchedule.Spec.EffectiveSchedules = schedules
+	s.requireStatusUpdate = true
+}
+
+func (s *ScheduleHandler) createNewEffectiveScheduleObj() {
+	newSchedule := &k8upv1alpha1.EffectiveSchedule{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: cfg.Config.OperatorNamespace,
+			Name:      rand.String(32),
```
cimnine, line 254: For debugging purposes, I suggest to find a more meaningful naming convention than just 32 random characters, e.g. `random-weekly-$(RAND_5)` (i.e. the original schedule plus a short random suffix).

tobru:
> the default is hardcoded k8up-system atm

I wouldn't make this hard-coded, should be configurable. In the Helm Chart we can make use of the [Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#capabilities-of-the-downward-api) to make it "automatically" to be the same namespace as K8up is installed in to.

ccremer:
> I wouldn't make this hard-coded, should be configurable.

It's the *default* that is hardcoded. The value itself is already configurable.

handler/schedule.go
ccremer: I thought about it too. but it's not feasible:

1. If the only/last schedule changes from `@weekly-random` to `@daily-random` then it's even more confusing. We would have to delete the old, and then create a new object. Sounds like overengineering just for naming the users shouldn't care about much?
2. The `Schedule` can specify a `check` schedule and `backup` schedule with different `@...-random` schedules, but both will make it into the same `EffectiveSchedule` currently. So which one to pick? 

Thinking about the 2nd point. There might be a flaw in the current code if both job types end up in the same effective schedule object. especially later for deduplication maybe it's better to have all backup schedules in one, all Checks in another... I need to look into it.

controllers/schedule_controller.go
```diff
@@ -41,18 +45,50 @@ func (r *ScheduleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return reconcile.Result{}, err
 	}
 
+	effectiveSchedule, err := r.FetchEffectiveScheduleResource(ctx, schedule)
+	if err != nil {
+		r.Log.Info("could not retrieve list of effective schedules, try later again", "error", err.Error())
+		return ctrl.Result{Requeue: true, RequeueAfter: 60 * time.Minute}, err
```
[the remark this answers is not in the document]
ccremer: that's actually wrong :see_no_evil:  I wanted 60 seconds, then switched to minute but forgot to change 60 to 1 :)

handler/schedule.go
```diff
@@ -206,6 +233,80 @@ func (s *ScheduleHandler) getRandomSchedule(jobType k8upv1alpha1.JobType, origin
...
 	s.requireStatusUpdate = true
+	s.Log.Info("Creating new EffectiveSchedule", "name", k8upv1alpha1.GetNamespacedName(newSchedule))
+	err := s.Client.Create(s.CTX, newSchedule)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		s.Log.Error(err, "could not persist effective schedules", "name", newSchedule.Name)
+		s.SetConditionFalseWithMessage(k8upv1alpha1.ConditionReady, k8upv1alpha1.ReasonCreationFailed, "could not persist effective schedules: %s", err.Error())
+	} else {
+		s.SetConditionTrueWithMessage(k8upv1alpha1.ConditionReady, k8upv1alpha1.ReasonReady, "effective schedule created: '%s'", k8upv1alpha1.GetNamespacedName(newSchedule))
+		s.effectiveSchedule = newSchedule
+	}
```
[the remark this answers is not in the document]
ccremer: Removing the else clause means that the following code is executed regardless of the outcome in the if-condition, but that's not the intention here. I can only set true condition on successes. I would need a return statement in the if-clause to prevent that.

handler/schedule.go
ccremer: After a 1-on-1 discussion we concluded that there should be an effective Schedule resource for each Job type, not per schedule itself. Otherwise this would be harder to deduplicate for different job types.

controllers/schedule_controller.go
```diff
@@ -41,18 +45,50 @@ func (r *ScheduleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 func (r *ScheduleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&k8upv1alpha1.Schedule{}).
 		WithEventFilter(predicate.GenerationChangedPredicate{}).
 		Complete(r)
 }
+
+// FetchEffectiveScheduleResource retrieves a list of EffectiveSchedules and filter the one that matches the given schedule.
+// Returns an error if the listing failed, but nil when no matching EffectiveSchedule object was found.
+func (r *ScheduleReconciler) FetchEffectiveScheduleResource(ctx context.Context, schedule *k8upv1alpha1.Schedule) (*k8upv1alpha1.EffectiveSchedule, error) {
```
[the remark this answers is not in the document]
ccremer: I wonder if it makes more sense to have a bidirectional link, e.g. each Schedule object also maintains a link to the EffectiveSchedule with name, namespace and jobtype info. That way, we won't be needing to fetch and filter all effective Schedules, possibly lowering the load on the API a bit. However, bidirectional links are easier to break.

WDYT ?

handler/schedule.go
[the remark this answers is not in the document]
cimnine: You're right, that should have actually be part of my suggestion.

I suggest to return at the end of the `if` case and not use the `else` case here. it reduces one branch in the code – if not actually, then at least mentally.

controllers/schedule_controller.go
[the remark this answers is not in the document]
cimnine: I wouldn't go for the two-way binding. It only helps us very little but IMO adds additional complexity. Also we'd have to think about what happens if they get out of sync. Also there could be race-conditions as well.

handler/schedule.go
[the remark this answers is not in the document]
ccremer: after last dicussions I might need to rework this anyway

controllers/schedule_controller.go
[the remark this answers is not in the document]
ccremer: Agreed. Though race conditions could happen with the fetch-and-filter approach too, at latest when trying to update a resource that has been deleted for reason in the meantime. But yes, what to do with out-of-syncs...

I've added locally more printcolumns so that `k get effectiveschedule` prints the job type, generated schedule and the name and namespace of the first schedule ref. Do you think that's enough for debugging? One would need access to the operator namespace though.
I would not store the name of the effective schedule in the Ready condition's message anymore, only when there was an error.

ccremer:
Latest PoC looks promising:
```console
k get effectiveschedules -n k8up-system
NAME                      SCHEDULE NAMESPACE   SCHEDULE NAME   GENERATED SCHEDULE   JOB TYPE   AGE
backup-n27qhnqkcgstt49c   default              schedule-test   4 * * * *            backup     2m41s
check-llj8p6x4z7ssh4p6    default              schedule-test   44 * * * *           check      2m41s
```

controllers/schedule_controller.go
```diff
@@ -68,27 +68,28 @@ func (r *ScheduleReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		return nil, err
+		return map[k8upv1alpha1.JobType]k8upv1alpha1.EffectiveSchedule{}, err
 	}
 	return filterEffectiveSchedule(list, schedule), nil
 }
 
-func filterEffectiveSchedule(list k8upv1alpha1.EffectiveScheduleList, schedule *k8upv1alpha1.Schedule) *k8upv1alpha1.EffectiveSchedule {
-	for _, effectiveSchedule := range list.Items {
-		if effectiveSchedule.GetDeletionTimestamp() != nil {
+func filterEffectiveSchedule(list k8upv1alpha1.EffectiveScheduleList, schedule *k8upv1alpha1.Schedule) map[k8upv1alpha1.JobType]k8upv1alpha1.EffectiveSchedule {
```
cimnine, line 82: I don't like about the method name that it does not say what it filters for exactly.
Maybe `onlyReferencesOfSchedule` or alike?

controllers/schedule_it_test.go
```diff
@@ -0,0 +1,70 @@
...
+	resultSchedule.Spec.Backup.Schedule = "* * * * *"
+	ts.UpdateResources(resultSchedule)
+
+	ts.whenReconciling(resultSchedule)
+
+	resultSchedule = ts.newScheduleSpec(ts.givenSchedule.Name)
+	ts.FetchResource(k8upv1alpha1.GetNamespacedName(resultSchedule), resultSchedule)
+	ts.thenAssertCondition(resultSchedule, k8upv1alpha1.ConditionReady, k8upv1alpha1.ReasonReady, "resource is ready")
+
+	newEffectiveSchedules := ts.whenListEffectiveSchedules()
```
cimnine, line 56: Perhaps `actualEffectiveSchedules` could be a better term, as it _actual_ is a term often used to compare the actual state with the expected state.

controllers/schedule_it_utils_test.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+}
+
+func (ts *ScheduleControllerTestSuite) thenAssertEffectiveScheduleExists(expectedScheduleName string) {
+	list := ts.whenListEffectiveSchedules()
+	ts.Require().NotEmpty(list)
+	spec := list[0].Spec
+	ts.Require().NotEmpty(spec.EffectiveSchedules)
+	ref := spec.EffectiveSchedules[0]
+	ts.Assert().Equal(expectedScheduleName, ref.Name)
```
cimnine, line 39: Is there an advantage over this:

```suggestion
	assert.Equal(ts.T(), expectedScheduleName, ref.Name)
```

handler/effectiveschedule.go
```diff
@@ -0,0 +1,158 @@
...
+			continue
+		}
+		schedules = append(schedules, ref)
+	}
+	es.Spec.EffectiveSchedules = schedules
+	s.effectiveSchedules[jobType] = es
+}
+
+// synchronizeEffectiveSchedulesResources ensures that the effective schedules are created, updated or deleted depending on the Spec.
+// If no effective Schedule references the reconcile schedule resource anymore, it will be deleted.
```
cimnine, line 35:
```suggestion
// If no Schedule references the EffectiveSchedule resource anymore, it will be deleted.
```

handler/effectiveschedule.go
```diff
@@ -0,0 +1,158 @@
...
+// On errors, the Ready condition will be set to false.
+func (s *ScheduleHandler) synchronizeEffectiveSchedulesResources() error {
+	for _, es := range s.effectiveSchedules {
+		if len(es.Spec.EffectiveSchedules) <= 0 {
```
cimnine, line 39:
```suggestion
		if len(es.Spec.EffectiveSchedules) == 0 {
```

`len()` can not go below 0, can it?

handler/effectiveschedule.go
```diff
@@ -0,0 +1,158 @@
...
+		schedules = append(schedules, ref)
+	}
+	es.Spec.EffectiveSchedules = schedules
+	s.effectiveSchedules[jobType] = es
+}
+
+// synchronizeEffectiveSchedulesResources ensures that the effective schedules are created, updated or deleted depending on the Spec.
+// If no effective Schedule references the reconcile schedule resource anymore, it will be deleted.
+// On errors, the Ready condition will be set to false.
+func (s *ScheduleHandler) synchronizeEffectiveSchedulesResources() error {
```
cimnine, line 37: What I don't like in this function is that its implicitly modifying the state of ScheduleHandler. I.e. it modifies or deletes resources in `s.effectiveSchedules`, which leaves `s.effectiveSchedule` in an invalid state behind.

controllers/schedule_it_utils_test.go
ccremer: it's exactly the same. just kind of a shortcut.

controllers/schedule_controller.go
ccremer: Better now?

handler/effectiveschedule.go
ccremer: > which leaves `s.effectiveSchedule` in an invalid state behind.

how so? There is no `s.effectiveSchedule` anymore. I refactored it to be a map with name `s.effectiveSchedules`, since there can now be more than one EffectiveSchedule resource that match the `s.schedule` if using multiple job types.

Similar to SyncSchedules(), it's intended that this function actually tries match what is given from `s.effectiveSchedules`. It's updating the values since Client.Update() also updates the given resource with the latest object metadata.

ccremer:
@tobru wanna review docs?

docs/modules/ROOT/pages/references/object-specifications.adoc
```diff
@@ -365,3 +363,22 @@ spec:
 * `backupCommand`: command that should get executed within the pod. *Attention* the command should output its data to stdout so that wrestic can pick it up correctly
 * `fileExtension`: as this leverages the stdin backup capabilities of Restic it will generate a virtual file. That file name is by default just the name of the PreBackup pod. But to make restores easier you can define a file extension, that gets appended to the filename. For example: ".sql" for a mysql dump
 * `pod`: pod is default `podTemplateSpec` of https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/#pod-templates[Kubernetes].
+
+== EffectiveSchedule
+
+An `EffectiveSchedule` is a status object that persists generated schedule definitions when using <<references/schedule-specification.adoc#,K8up specific schedules>> like `@daily-random`.
```
tobru, line 369:
```suggestion
An `EffectiveSchedule` is a status object that persists generated schedule definitions when using xref:references/schedule-specification.adoc[K8up specific schedules] like `@daily-random`.
```

tobru (approved):
Docs LGTM

docs/modules/ROOT/pages/references/object-specifications.adoc
```diff
@@ -365,3 +363,22 @@ spec:
...
+They are completely self-managed in a single namespace, users should not have the need to interact with them.
```
cimnine, line 370:
```suggestion
They are completely controller-managed in the controller namespace.
Users should not have the need to interact with them.
```

ccremer:
Thanks for the pair review @cimnine 
I'll start organizing the commits now

ccremer:
@cimnine want to approve?

I noticed that the coverage tool does not count the coverage done with integration tests. that's probably also the reason why this PR's diff is so low.
```
$ make integration-test
...
kubebuilder/bin/
kubebuilder/bin/etcd
kubebuilder/bin/kube-apiserver
kubebuilder/bin/kubectl
setting up env vars
?   	github.com/vshn/k8up	[no test files]
?   	github.com/vshn/k8up/api/v1alpha1	[no test files]
ok  	github.com/vshn/k8up/cfg	0.027s	coverage: 23.3% of statements
ok  	github.com/vshn/k8up/controllers	15.596s	coverage: 27.1% of statements
ok  	github.com/vshn/k8up/executor	0.017s	coverage: 13.8% of statements
ok  	github.com/vshn/k8up/handler	0.028s	coverage: 14.1% of statements
?   	github.com/vshn/k8up/job	[no test files]
ok  	github.com/vshn/k8up/observer	0.042s	coverage: 52.6% of statements
ok  	github.com/vshn/k8up/prebackup	0.028s	coverage: 3.8% of statements
ok  	github.com/vshn/k8up/queue	0.016s	coverage: 87.5% of statements
ok  	github.com/vshn/k8up/scheduler	0.013s	coverage: 55.2% of statements
```

```
$ make test
go fmt ./...
go vet ./...
go test ./... -coverprofile cover.out
?   	github.com/vshn/k8up	[no test files]
?   	github.com/vshn/k8up/api/v1alpha1	[no test files]
ok  	github.com/vshn/k8up/cfg	0.039s	coverage: 23.3% of statements
ok  	github.com/vshn/k8up/controllers	0.028s	coverage: 6.8% of statements
ok  	github.com/vshn/k8up/executor	0.012s	coverage: 13.8% of statements
ok  	github.com/vshn/k8up/handler	0.014s	coverage: 14.1% of statements
?   	github.com/vshn/k8up/job	[no test files]
ok  	github.com/vshn/k8up/observer	0.016s	coverage: 52.6% of statements
ok  	github.com/vshn/k8up/prebackup	0.023s	coverage: 3.8% of statements
ok  	github.com/vshn/k8up/queue	0.046s	coverage: 87.5% of statements
ok  	github.com/vshn/k8up/scheduler	0.018s	coverage: 55.2% of statements
```
When doing tests with Goland + coverage, the total test coverage should increase from 17% to 31%
Lemme investigate a bit and make a new PR

cimnine (approved):

ccremer:
> I noticed that the coverage tool does not count the coverage done with integration tests

false alarm. I didn't notice the diff in the controllers package

Issue #339, opened 2021-02-01
e2e-test: sha1sum command is not found on macOS


marcellanz:
## Describe the bug

While running the e2e-test on macOS, the e2e images tag `E2E_TAG` is not calculated properly as the `sha1sum` command is not found on macOS.

## Additional context

None.

## Logs

While running the e2e test, log entries about can be found like: `missing /bin/bash: sha1sum: command not found`

```console
user@host> make e2e-test
...
CGO_ENABLED=0 go build -o ./k8up main.go
/bin/bash: sha1sum: command not found
docker build . -t docker.io/vshn/k8up:latest -t quay.io/vshn/k8up:latest -t local.dev/k8up/e2e:e2e_
Sending build context to Docker daemon  46.13MB
Step 1/5 : FROM docker.io/library/alpine:3.13 as runtime
 ---> e50c909a8df2
Step 2/5 : ENTRYPOINT ["k8up"]
 ---> Using cache
 ---> eb6cf1eb416b
Step 3/5 : RUN     apk add --no-cache curl bash tzdata
 ---> Using cache
 ---> 23b58add1f0c
Step 4/5 : COPY k8up /usr/bin/
 ---> db87615ffcc6
Step 5/5 : USER 65532:65532
 ---> Running in 21f1191084c2
Removing intermediate container 21f1191084c2
 ---> 5d7ed51f0748
Successfully built 5d7ed51f0748
Successfully tagged vshn/k8up:latest
Successfully tagged quay.io/vshn/k8up:latest
Successfully tagged local.dev/k8up/e2e:e2e_
Warning: default unsupported in CRD version v1beta1, v1 required. Removing defaults.
Warning: default unsupported in CRD version v1beta1, v1 required. Removing defaults.
Warning: default unsupported in CRD version v1beta1, v1 required. Removing defaults.
2049/06/30 15:37:52 Running post-generate in /Users/user/tmp/v/k8up
2049/06/30 15:37:52 Reading file config/crd/apiextensions.k8s.io/v1beta1/base/backup.appuio.ch_prebackuppods.yaml
2049/06/30 15:37:52 Added  'preserveUnknownFields' after line 9
2049/06/30 15:37:52 Removed 'protocol' in line 777
2049/06/30 15:37:52 Removed 'protocol' in line 2295
2049/06/30 15:37:52 Writing new file to config/crd/apiextensions.k8s.io/v1beta1/base/backup.appuio.ch_prebackuppods.yaml
[...]

PR #340, opened 2021-02-01
Fix "sha1sum command is not found" in e2e-tests on macOS

marcellanz:
## Summary

Fixes Issue #339 – "e2e-test: sha1sum command is not found on macOS"
While running the e2e-test on macOS, the e2e images tag `E2E_TAG` is not calculated properly as the `sha1sum` command is not found on macOS.

## Checklist

### DIFF
--- a/Makefile.vars.mk
+++ b/Makefile.vars.mk
@@ -18,7 +18,8 @@ KIND_KUBECONFIG ?= $(TESTBIN_DIR)/kind-kubeconfig-$(KIND_NODE_VERSION)
 KIND_CLUSTER ?= k8up-$(KIND_NODE_VERSION)
 KIND_KUBECTL_ARGS ?= --validate=true
 
-E2E_TAG ?= e2e_$(shell sha1sum $(BIN_FILENAME) | cut -b-8)
+SHASUM ?= $(shell command -v sha1sum > /dev/null && echo "sha1sum" || echo "shasum -a1")
+E2E_TAG ?= e2e_$(shell $(SHASUM) $(BIN_FILENAME) | cut -b-8)
 E2E_REPO ?= local.dev/k8up/e2e
 E2E_IMG = $(E2E_REPO):$(E2E_TAG)
 


Makefile.vars.mk
```diff
@@ -18,7 +18,12 @@ KIND_KUBECONFIG ?= $(TESTBIN_DIR)/kind-kubeconfig-$(KIND_NODE_VERSION)
 KIND_CLUSTER ?= k8up-$(KIND_NODE_VERSION)
 KIND_KUBECTL_ARGS ?= --validate=true
 
-E2E_TAG ?= e2e_$(shell sha1sum $(BIN_FILENAME) | cut -b-8)
+ifeq ($(shell which sha1sum 2>&1; echo $$?),0)
```
cimnine, line 21: `which` should be avoided in favour of e.g. `command` or `hash` because it is not consistent among operating systems: https://stackoverflow.com/a/677212/172132

Makefile.vars.mk
```diff
@@ -18,7 +18,12 @@ KIND_KUBECONFIG ?= $(TESTBIN_DIR)/kind-kubeconfig-$(KIND_NODE_VERSION)
...
+  SHA1SUM = sha1sum
+else
+  SHA1SUM = shasum -a1
+endif
+E2E_TAG ?= e2e_$(shell $(SHA1SUM) $(BIN_FILENAME) | cut -b-8)
```
cimnine, line 26: Alternatively, this could also be written more concise:

```suggestion
SHASUM ?= $(shell command -v sha1sum > /dev/null && echo "sha1sum" || echo "shasum -a1")
E2E_TAG ?= e2e_$(shell $(SHASUM) $(BIN_FILENAME) | cut -b-8)
```

cimnine:
Thank you for taking the time to suggest this PR. I think by just switching to `shasum -a1` on all platforms we could avoid most of the pain this PR addresses. What do you think?

Makefile.vars.mk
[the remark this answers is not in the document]
marcellanz: Today I learned `shasum` on macOS is from its perl installation :)
Just tried on debian and others to run it. Even if perl is installed on the `debian:latest` image, I had to install it with `apt-get install libdigest-sha-perl`. 
To use the same command would be ideal I think, but it seems simple OS images do not have it installed. wdyt?

marcellanz: > Alternatively, this could also be written more concise:

I'm open for the most idiomatic and portable way to do it. I'd prefer readability over conciseness. I've seen the approach to try the first command and then the next one if that fails. It works, but also is less expressive.

Makefile.vars.mk
marcellanz: Great point.

marcellanz:
> Thank you for taking the time to submit this PR. I think by just switching to `shasum -a1` on all platforms we could avoid most of the pain this PR addresses. What do you think?

You're most welcome. I did comment in the suggested changes of yours. I'd prefer any option that is most portable.

marcellanz:
In general, including OS-dependent .mk.vars files could be the most explicit and clean way. No use of 'which' or try to run commands.

ccremer:
If you happen to know a clever solution how we could get the SHA sum of the tagged docker image (you know, like `local.dev/k8up/e2e@sha256:blahblubb`), and then load this into KIND, we wouldn't need to generate a shasum on our own. The current solution where we checksum the binary was kind of lazy but good-enough until now :)

ccremer:
> which also could be cleaned up I think.

`make e2e-clean` or generally `make clean` cleans those up. Run `make help` to see a list of targets :)

> Could't we use a random ID instead, uses for a run and then being regenerated at the next run?

We used a date-based tag generation before. The problem is that it's always a new tag, even if nothing changed. But if the image stays the same (because the binary hasn't changed), local repeated e2e-tests are faster. The random ID would be similar to date.

Makefile.vars.mk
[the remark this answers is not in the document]
cimnine: I'd say let's not bikeshed any longer:

- Either go with `sha1sum` if available and else `shasum`.
- Or we just politely ask macOS developers to `brew install coreutils` or `brew install md5sha1sum`.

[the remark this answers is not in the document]
marcellanz: lets choose: "Either go with sha1sum if available and else shasum"

marcellanz:
@ccremer see the accepted suggestion of yours. It should be good enough.

cimnine (approved):

ccremer:
Can you squash your commits? After that we can merge

marcellanz:
I've squashed the commits.

ccremer:
thx!

PR #342, opened 2021-02-02
Speed up `make test`

cimnine:
## Summary

The target `make lint` depends on `make vet fmt`.
The target `make test` also depends on `make vet fmt`, which is redundant in that case.
To speed up running the unit tests, `make vet fmt` have been removed as dependencies from `make test`.

## Checklist

- [x] Update tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
 all: build ## Invokes the build target
 
 .PHONY: test
-test: fmt vet ## Run tests
+test: ## Run tests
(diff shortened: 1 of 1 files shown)


ccremer:
The `vet` and `fmt` targets are already run in the `lint` workflow. Is that not enough?

cimnine:
> The vet and fmt targets are already run in the lint workflow. Is that not enough?

Sure. Should we remove it from `make test` as well then?

ccremer:
fine by me

cimnine:
Refactored as discussed and rebased onto latest `master`.

ccremer (approved):

PR #344, opened 2021-02-05
Deduplication for Prune and Check jobs

### LINKED ISSUE #214: [Feature] Deduplicate Check and Prunes for the same backup repository

ccremer:
## Summary

WIP

Closes #214 

## Checklist

- [ ] Update tests.


Kidswiss:
Did some testing with this.

It currently doesn't detect when someone changes the repository on an existing schedule.

I created two identical random schedules for check in two namespaces:

```
k get schedules.backup.appuio.ch --all-namespaces
NAMESPACE   NAME                     AGE
default     schedule-deduplication   24s
test        schedule-deduplication   12s
```

It was deduplicated properly:

```
k get effectiveschedules.backup.appuio.ch
NAME                     SCHEDULE NAMESPACE   SCHEDULE NAME            GENERATED SCHEDULE   JOB TYPE   AGE
check-lsnrqxzf9cjrbkf2   default              schedule-deduplication   54 * * * *           check      31m
```

Changing the backend on one of the schedules doesn't de-deduplicate the effective schedule. It still remains one. Also after deleting and re-creating the schedule it still stays the same.

I have to delete all schedules and re-create them to get it work again.

I did some debugging and it looks like `upsertEffectiveScheduleInternally()`, doesn't create a new effective schedule in that case.

ccremer:
> It currently doesn't detect when someone changes the repository on an existing schedule.

argh darn. I thought of changisch a schedule, but not changing of a backend. I't simply not implemented

handler/effectiveschedule.go
```diff
@@ -46,69 +46,58 @@ func (s *ScheduleHandler) synchronizeEffectiveSchedulesResources() error {
...
+	es, found := s.effectiveSchedules[ctx.jobType]
 	if !found {
-		es = s.newEffectiveSchedule(jobType)
+		es = s.newEffectiveSchedule(ctx.jobType)
 	}
-	es.Spec.GeneratedSchedule = schedule
-	es.Spec.JobType = jobType
-	es.Spec.OriginalSchedule = originalSchedule
+	es.Spec.GeneratedSchedule = ctx.effectiveSchedule
+	es.Spec.JobType = ctx.jobType
```
Kidswiss, line 87: Not sure if it actually makes sense to update `JobType` and `backendString` here? Because by our definition this is what defines the scope of the deduplication. Instead of changing the existing one it should create a new one that would handle the new type and backend. Ig there are no more schedules with this type and backend combo it can be deleted.

Kidswiss (changes requested):

Kidswiss:
> argh darn. I thought of changing a schedule, but not changing of a backend. I't simply not implemented

Changing the schedule is the easier use case. Changing the backend could get more complex as it creates a new scope for deduplication and needs to be removed from existing effective schedules.

handler/effectiveschedule.go
ccremer: This piece of code is used for both creating and updating the ES ("upsert" being an abbrev for 'update or insert').
Upon updating, the context should be equal to what is already existing, so those lines become basically no-ops.

ccremer: my point is, the detection whether a backend or schedule has changed, should happen before this func is called.

ccremer:
> Changing the schedule is the easier use case. Changing the backend could get more complex as it creates a new scope for deduplication and needs to be removed from existing effective schedules.

I'd say it's about the same complexity. changing the original schedule (e.g. from daily-random to hourly-random) also requires to be removed from existing effective schedules

handler/effectiveschedule.go
Kidswiss: Fair enough. So with the implementation of backend detection it should also be a no-op?

ccremer: I think so, but I'd need to look into it.

PR #345, opened 2021-02-05
Introduce BackendInterface and document more backend settings

ccremer:
## Summary

While preparing for #344 I was able to isolate improvements regarding to backends in this separate PR.
Also addresses some CodeClimate issues while at it.

## Checklist
- [x] Update tests.

### DIFF
--- a/api/v1alpha1/backend.go
+++ b/api/v1alpha1/backend.go
@@ -2,59 +2,52 @@ package v1alpha1
 
 import (
 	"fmt"
+	"reflect"
+
 	"github.com/vshn/k8up/cfg"
 
 	corev1 "k8s.io/api/core/v1"
 )
 
-type Backend struct {
-	// RepoPasswordSecretRef references a secret key to look up the restic repository password
-	RepoPasswordSecretRef *corev1.SecretKeySelector `json:"repoPasswordSecretRef,omitempty"`
-	Local                 *LocalSpec                `json:"local,omitempty"`
-	S3                    *S3Spec                   `json:"s3,omitempty"`
-	GCS                   *GCSSpec                  `json:"gcs,omitempty"`
-	Azure                 *AzureSpec                `json:"azure,omitempty"`
-	Swift                 *SwiftSpec                `json:"swift,omitempty"`
-	B2                    *B2Spec                   `json:"b2,omitempty"`
-	Rest                  *RestServerSpec           `json:"rest,omitempty"`
-}
+type (
+	// Backend allows configuring several backend implementations.
+	// It is expected that users only configure one storage type.
+	Backend struct {
+		// RepoPasswordSecretRef references a secret key to look up the restic repository password
+		RepoPasswordSecretRef *corev1.SecretKeySelector `json:"repoPasswordSecretRef,omitempty"`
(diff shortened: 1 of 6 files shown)


api/v1alpha1/backend.go
```diff
@@ -2,147 +2,149 @@ package v1alpha1
...
-		return b.GCS.EnvVars(vars)
-	}
+	// +k8s:deepcopy-gen=false
 
-	if b.Local != nil {
-		return b.Local.EnvVars(vars)
+	// BackendInterface represents a Backend for internal use.
+	BackendInterface interface {
+		EnvVars(vars map[string]*corev1.EnvVarSource) map[string]*corev1.EnvVarSource
+		GetRepositoryString() string
```
Kidswiss, line 32: Any objections in simply calling this `String()`? Then the implementations would also implement the Stringer interface.

api/v1alpha1/backend.go
```diff
@@ -2,147 +2,149 @@ package v1alpha1
...
-
-	if b.Azure != nil {
-		return b.Azure.EnvVars(vars)
-	}
-
-	if b.B2 != nil {
-		return b.B2.EnvVars(vars)
+type (
+	// Backend allows configuring several backend implementations.
+	// It is expected that users only configure one storage type.
```
Kidswiss, line 14: We should probably document that expectation somewhere, if that hasn't happened already.

Kidswiss (changes requested):

api/v1alpha1/backend.go
ccremer: Hm, personally I find `String()` to be a very vague name as to what to expect. But I see the argument for the Stringer interface.

api/v1alpha1/backend.go
ccremer: Nope, in fact there's barely any backends documented.

ccremer:
@tobru want to have a look at the documentation?

docs/modules/ROOT/pages/references/object-specifications.adoc
```diff
@@ -325,6 +337,52 @@ Settings:
...
+* `path`: Path of the B2 instance
+* `bucket`: name of the bucket that should be used
+* `accountIDSecretRef`: Kubernetes secret reference containing the account ID
+* `accountKeySecretRef`: Kubernetes secret reference containing the account key
+
+=== Local
+
+Settings:
+
+* `mountPath`: MountPath inside wrestic
```
ccremer, line 369: @Kidswiss is that right?

tobru (approved):
Documentation LGTM

docs/modules/ROOT/pages/references/object-specifications.adoc
Kidswiss: Yes that should be right. Although the local backend was never really tested. It may need more fields in the future. For example to specify what it should mount into the container...

ccremer: Hm okay, since this is already "released" in 0.x, I think we should leave it for 1.x and do improvements/fixes in normal release schedule. I'd expect bug reports if something is missing here.

PR #352, opened 2021-02-10
Introduce ScheduleSpecInterface

ccremer:
## Summary

* While implementing deduplication, this piece of code has become unmaintainable. Thus adding a new interface
* Addresses CodeClimate issues

## Checklist

- [x] Update tests.

### DIFF
--- a/api/v1alpha1/common_types.go
+++ b/api/v1alpha1/common_types.go
@@ -2,6 +2,7 @@ package v1alpha1
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 )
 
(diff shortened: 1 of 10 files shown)


api/v1alpha1/common_types.go
```diff
@@ -14,6 +15,24 @@ type (
 
 	// ConditionReason is a static/programmatic representation of the cause of a status condition.
 	ConditionReason string
+
+	// +k8s:deepcopy-gen=false
```
cimnine, line 19: Is this related to the `ScheduleSpecInterface`? Does it need the extra line break, or could it be defined below the `// ScheduleSpecInterface represents [...]` comment? (Same for the same situation below.)

api/v1alpha1/common_types.go
```diff
@@ -14,6 +15,24 @@ type (
...
+
+	// ScheduleSpecInterface represents a Job for internal use.
+	ScheduleSpecInterface interface {
+		GetDeepCopy() ScheduleSpecInterface
+		GetRunnableSpec() *RunnableSpec
+		GetSchedule() ScheduleDefinition
+		GetObjectCreator() ObjectCreator
+	}
+
+	// +k8s:deepcopy-gen=false
```
cimnine, line 29: Same here.

api/v1alpha1/prune_types.go
```diff
@@ -56,6 +56,15 @@ type Prune struct {
 	Status Status    `json:"status,omitempty"`
 }
 
+// +kubebuilder:object:root=true
```
cimnine, line 59: Would be same situation here as well.

cimnine (approved):
This changes does make the code more DRY and more concise, but I'm not so sure about the improvement of the readability and 'understandability'.

Nevertheless, since I don't know a better solution myself right now, I'm in no place to critise. And it's not that the unchanged code was very easy to grok, either. So if – at it's core – it's not an improvement, then it's at least as-good-as-before. Now, with reduced repetition and improved linter compliance, this change is a net-improvement after all.

api/v1alpha1/common_types.go
ccremer: Yes, otherwise there would be DeepCopy methods generated for those as well, but we don't need that.
the extra line break is to make sure the `// +k8s:deepcopy-gen=false` comment doesn't get part of the function godoc, otherwise it's part of the godoc documentation which I find annoying/irrelevant (same if it's below the type comment).

ccremer:
> This changes does make the code more DRY and more concise, but I'm not so sure about the improvement of the readability and 'understandability'.

I agree. This PR is a preparation for #344 , where the `createJobList()` func becomes much easier to maintain, but it doesn't overhaul K8up's design that it desperately needs for v2.

cimnine (approved):

Issue #272, opened 2021-01-11
Add e2e tests that creates a backup and restores from it


ccremer:
## Summary

**As** K8up developer\
**I want** an automated e2e test case that creates a backup and immediately restores from that backup\
**So that** I verify that my contribution doesn't break K8up's core functionality.

## Context

First, and probably most important e2e test case: Does the backup and restore of a simple file or folder work at all times? Until now, backups and restore were tested manually

## Out of Scope

* Don't test the same application stack as described in the tutorial. As a first step, we just want to test K8up, not the tutorial

## Further links

* Test Framework: https://github.com/bats-core/bats-detik
* Related Discussion: https://github.com/vshn/k8up/discussions/231
* current e2e tests: https://github.com/vshn/k8up/tree/master/e2e

## Acceptance criteria

all mentioned preconditions are scoped to local KIND-cluster set up by `make`

* Given a deployment of K8up operator, when a Schedule CR object with backup spec is created, then K8up creates an backup in the restic repository.
* Given an existing backup in the restic repository and K8up deployment, when a Restore CR object is created, then K8up restores the backup

## Implementation Ideas

* In all e2e test cases, the K8up operator deployment is a precondition, maybe deploy the operator and it's e2e docker image as part of the test setup, instead of doing it in its own test case (current).
* Install a Minio S3 server in local KIND cluster to have free and isolated S3 buckets. Use the installation method that offers the best maintainability vs complexity balance.
* The data to backup can be simple dummy file(s), but it needs to be verifiable for the restore test.
* Do the test in its own namespace and leave the remaining artifacts. That should help with local development when trying to debug an issue.

PR #369, opened 2021-02-18
Refactor e2e tests

cimnine:
## Summary

This change aims to simplify the actual test cases and allows
to extend them more granuarly, e.g. by allowing to
re-use the resource definitions by composing them as needed.

The idea is that all the e2e tests can use the resource definitions in the new `definitions` folder to their liking.
Otherwise, e.g. the `deployment.yaml`, would have to be duplicated for every test.
Or the upcoming test of a restore will probably have to do a `backup` as well. So it can share that with the e2e test that tests a backup.

Related to #272.

## Checklist

- [x] ~Update tests.~


e2e/lib/k8up.bash
```diff
@@ -1,12 +1,27 @@
...
+	cp -r /tmp/detik debug || true
+}
+
+kustomize() {
+	go run sigs.k8s.io/kustomize/kustomize/v3 "${@}"
+}
+
+prepare() {
+	mkdir -p "debug/${1}"
+	kustomize build "${1}" > "debug/${1}/main.yml"
```
ccremer, line 20:
```suggestion
	kustomize build "${1}" -o "debug/${1}/main.yml"
```

e2e/test-deployment.bats
```diff
@@ -0,0 +1,21 @@
...
+load "lib/detik"
+load "lib/k8up"
+
+DETIK_CLIENT_NAME="kubectl"
+DETIK_CLIENT_NAMESPACE="k8up-system"
+DEBUG_DETIK="true"
+
+@test "reset the debug file" {
+	reset_debug
+}
```
ccremer, line 13: Have you tried moving this reset to setup() or teardown() ? Seems kinda out of place as a "test case"

cimnine: I guess this was originally copied from the DETIK example. I've moved it to a `setup_file` function in the `k8up.bash` file.

cimnine:
Integrated feedback and rebased onto latest `master`.

ccremer (approved):

PR #371, opened 2021-02-18
WIP POC prebackup handling

Kidswiss:
## Summary

Closes: https://github.com/vshn/k8up/issues/137

## Checklist

- [ ] Update tests.

### DIFF
--- a/prebackup/prebackup.go
+++ b/prebackup/prebackup.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 
 	batchv1 "k8s.io/api/batch/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/utils/pointer"
 
 	k8upv1alpha1 "github.com/vshn/k8up/api/v1alpha1"
 	"github.com/vshn/k8up/cfg"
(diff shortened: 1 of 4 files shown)


prebackup/prebackup.go
```diff
@@ -35,41 +32,43 @@ func NewPrebackup(config job.Config) *PreBackup {
...
 const (
-	// ConditionPreBackupPodsReady is True if Deployments for all Container definitions were created and are ready
-	ConditionPreBackupPodsReady k8upv1alpha1.ConditionType = "PreBackupPodsReady"
-
-	// ReasonNoPreBackupPodsFound is given when no PreBackupPods are found in the same namespace
-	ReasonNoPreBackupPodsFound k8upv1alpha1.ConditionReason = "NoPreBackupPodsFound"
+	// ConditionPreBackupJobsReady is True if Deployments for all Container definitions were created and are ready
+	ConditionPreBackupJobsReady k8upv1alpha1.ConditionType = "PreBackupJobsReady"
+	// ReasonNoPreBackupJobsFound is given when no PreBackupPods are found in the same namespace
+	ReasonNoPreBackupJobsFound k8upv1alpha1.ConditionReason = "NoPreBackupJobsFound"
```
ccremer, line 38: Since [conditions are part of the API](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status), we would be introducing breaking changes with this renaming. I suggest to leave them as they are. After all, they are not exactly wrong anyway. If they were named "PrebackupDeployment", then we would have a problem.

PR #372, opened 2021-02-18
E2E-Test Backup

cimnine:
## Summary

The goal of this PR is to test a backup E2E. It's part of #272 .

## Checklist

- [x] Update tests.


e2e/lib/k8up.bash
```diff
@@ -19,13 +30,57 @@ kustomize() {
...
 prepare() {
 	mkdir -p "debug/${1}"
 	kustomize build "${1}" -o "debug/${1}/main.yml"
-	sed -i -e "s|\$E2E_IMAGE|'${E2E_IMAGE}'|" "debug/${1}/main.yml"
+	sed -i \
+		-e "s|\$E2E_IMAGE|'${E2E_IMAGE}'|" \
+		"debug/${1}/main.yml"
+	sed -i \
+		-e "s|\$WRESTIC_IMAGE|'${WRESTIC_IMAGE}'|" \
+		"debug/${1}/main.yml"
```
ccremer, line 61:
```suggestion
	sed -i \
		-e "s|\$E2E_IMAGE|'${E2E_IMAGE}'|" \
		-e "s|\$WRESTIC_IMAGE|'${WRESTIC_IMAGE}'|" \
		"debug/${1}/main.yml"
```

e2e/definitions/subject/pvc.yaml
```diff
@@ -0,0 +1,16 @@
...
+  accessModes:
+  - ReadWriteMany
+  resources:
+    requests:
+      storage: 1Gi
+  storageClassName: hostpath
+  volumeMode: Filesystem
+  selector:
+    matchLabels:
+      pv.kubernetes.io/type: target
```
ccremer, line 16: this label is just arbitrary, but it does have to match with the PV
```suggestion
      pv.kubernetes.io/type: e2e-test
```

e2e/definitions/subject/pv.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: subject-pv
+  labels:
+    pv.kubernetes.io/type: target
```
ccremer, line 6:
```suggestion
    pv.kubernetes.io/type: e2e-test
```

e2e/definitions/subject/pv.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  accessModes:
+  - ReadWriteMany
+  capacity:
+    storage: 1Gi
+  hostPath:
+    path: /var/local-path-provisioner/pvc-subject
```
ccremer, line 13: Where is the data being stored exactly? How do I inspect storage contents manually for debugging purposes or after e2e-testing?

e2e/definitions/backup/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+resources:
+- ../secrets
+- backup.yaml
+
+namespace: k8up-system
```
ccremer, line 5: Maybe we can test the stuff in a different namespace than the one the operator runs in? Like simulating a user namespace.

e2e/definitions/deployment/deployment.yaml
```diff
@@ -13,3 +13,5 @@ spec:
         env:
         - name: BACKUP_LOG_LEVEL
           value: debug
+        - name: BACKUP_IMAGE
+          value: vshn/wrestic:v0.2.0
```
ccremer, line 17:
```suggestion
          value: quay.io/vshn/wrestic:v0.2.0
```

e2e/definitions/secrets/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+resources:
+- secrets.yaml
```
ccremer, line 2: there is no namespace for the secrets? If not, the `default` namespace is being used. I propose to make this explicit

e2e/definitions/subject/alpine.yaml
```diff
@@ -0,0 +1,26 @@
...
+      labels:
+        app: subject
+    spec:
+      containers:
+      - name: subject-container
+        image: alpine
+        imagePullPolicy: IfNotPresent
+        volumeMounts:
+        - name: volume
+          mountPath: /pvc/
```
ccremer, line 22: I propose a real app. Besides, `alpine` again pulls from Docker Hub, which I would avoid.
Redis has reasonably low image size (40mb) and fast container startup (1s)
```suggestion
      - name: subject-container
        image: quay.io/bitnami/redis:latest
        imagePullPolicy: IfNotPresent
        env:
        - name: REDIS_PASSWORD
          value: e2e-test
        volumeMounts:
        - name: volume
          mountPath: /bitnami/redis
```

ccremer (changes requested):

e2e/definitions/subject/pv.yaml
cimnine: You can either `docker exec` into the kind controller container to check `/var/local-path-provisioner/pvc-subject` or you can `kubectl exec` into the `subject-container` pod and inspect the mount there.

ccremer: Any reasons against mounting a `testbin/data` folder directly? would be a better dev experience than exec' into containers

e2e/lib/k8up.bash
```diff
@@ -19,13 +30,56 @@ kustomize() {
...
 }
+
+given_a_subject() {
+	kubectl create namespace k8up-e2e-subject || true
+	apply definitions/subject
+}
+
+given_s3_storage() {
+	# TODO HELM REPO
+	helm install minio \
```
ccremer, line 76: Don't forget to also install this in the e2e GH workflow, e.g. with https://github.com/Azure/setup-helm

ccremer: actually, we might not need to do this.
The GH machine might already have this: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md#package-management. any helm v3 should be fine for us.

e2e/lib/k8up.bash
```diff
@@ -19,13 +30,82 @@ kustomize() {
...
+		jq -e 'map(select(.name == "minio")) | length == 1'
+	return ${?}
+}
+
+ensure_minio_helm_repo() {
+	if is_minio_helm_repo_installed; then
+		return
+	fi
+
+	helm repo add minio https://helm.min.io/
```
ccremer, line 86: Why not just call `helm repo add minio https://helm.min.io/` and `helm repo update` every time? I don't really see the need to do complex checking whether minio repo already exists etc and these commands are pretty fast.

e2e/kind.mk
```diff
@@ -25,7 +25,11 @@ $(KIND): $(testbin_created)
 
 $(KIND_KUBECONFIG): export KUBECONFIG = $(KIND_KUBECONFIG)
 $(KIND_KUBECONFIG): $(KIND)
-	$(KIND) create cluster --name $(KIND_CLUSTER) --image kindest/node:$(KIND_NODE_VERSION)
+	@mkdir -p debug/data
```
ccremer, line 28:
```suggestion
	@mkdir -p debug/data/pvc-subject
```

cimnine:
Green 💚  Finally 🎉

e2e/lib/k8up.bash
```diff
@@ -19,13 +30,68 @@ kustomize() {
...
 }
 
 apply() {
 	prepare "${@}"
 	kubectl apply -f "debug/${1}/main.yml"
 }
+
+given_a_subject() {
+	kubectl delete namespace k8up-e2e-subject || true
+	kubectl create namespace k8up-e2e-subject || true
```
ccremer, line 81:
```suggestion
	kubectl delete namespace k8up-e2e-subject --ignore-not-found
	kubectl delete pv subject-pv --ignore-not-found
	kubectl create namespace k8up-e2e-subject || true
```

e2e/test-deployment.bats
```diff
@@ -9,8 +9,11 @@ DETIK_CLIENT_NAMESPACE="k8up-system"
 DEBUG_DETIK="true"
 
 @test "verify the deployment" {
-	apply definitions/deployment
-	echo "$output"
+	# Remove traces of k8up deployments from other tests
+	kubectl delete namespace "$DETIK_CLIENT_NAMESPACE" || true
+	kubectl create namespace "$DETIK_CLIENT_NAMESPACE" || true
+
+	apply definitions/k8up
```
ccremer, line 16: I would just delete this file. It was there as the first skeleton test since we didn't any e2e tests back then.
Besides, the operator is already transparently tested with this backup test now

e2e/definitions/secrets/kustomization.yaml
```diff
@@ -0,0 +1,3 @@
...
+namespace: k8up-system
```
ccremer, line 3:
```suggestion
namespace: k8up-e2e-subject
```

e2e/test-backup.bats
```diff
@@ -0,0 +1,31 @@
...
+	given_s3_storage
+	given_a_subject
+
+	apply definitions/backup
+	try "at most 10 times every 1s to get backup named 'k8up-k8up-test' and verify that '.status.started' is 'true'"
+	try "at most 10 times every 1s to get job named 'k8up-k8up-test' and verify that '.status.active' is '1'"
+	try "at most 10 times every 2s to find 1 pod named 'k8up-k8up-test-.....' with 'status' being 'Running'"
+
+	# Wait for completion
+	try "at most 10 times every 2s to get backup named 'k8up-k8up-test' and verify that '.status.finished' is 'true'"
```
ccremer, line 25: I would check for a condition. The "finished" flag is kind of deprecated.
you could also use this:
`kubectl -n $DETIK_CLIENT_NAMESPACE wait --for condition=ready backup/k8up-k8up-test --timeout 1m`
But I'm fine with the `try` command

e2e/test-deployment.bats
cimnine: I like that it explicitly tests the deployment. When the other test fails, and it's because of the deployment, then it will be hard to interpret that. For example, because there are no `try` statements regarding the deployment, as these only apply to the current `DETIC_CLIENT_NAMESPACE`. (And this namespace is different in other tests, as we operator in `k8up-e2e-subject` there and not `k8up-system` as here.)

ccremer: Fine. I would prefer though a different file name. So that Bats takes the deployment test first instead of backup. Bats seems to choose the test files alphabetically by filename. Maybe like `00-test-deployment.bats`

Maybe we can use also another namespace for this? Or is that too much to configure?

e2e/test-backup.bats
cimnine: I guess you meant condition=completed?

ccremer (approved):

ccremer:
Maybe squash a few commits? (not all into one though :)

PR #380, opened 2021-02-23
Add missing finalizer RBAC rules

### LINKED ISSUE #379: OpenShift 3.11 missing RBAC for jobs/finalizers
## Describe the bug

When deploying to OpenShift 3.11 we found that the operator can't create the jobs. We found following errors on the backup objects:
```
status:
    conditions:
    - lastTransitionTime: "2021-02-22T20:24:30Z"
      message: The resource is ready
      reason: Succeeded
      status: "True"
      type: PreBackupPodsReady
    - lastTransitionTime: "2021-02-22T20:24:07Z"
      message: 'unable to create  ''namespace/backup-schedule-backup-mmnr4'':
        jobs.batch "backup-schedule-backup-mmnr4" is forbidden: cannot set blockOwnerDeletion
        if an ownerReference refers to a resource you can''t set finalizers on: no
        RBAC policy matched, <nil>'
      reason: CreationFailed
      status: "False"
      type: Ready
```

It seems that the operator can't update finalizers on job objects. We adjusted the RBAC as follows (https://github.com/vshn/k8up/blob/master/config/rbac/role.yaml):
```
- apiGroups:
  - batch
  resources:
  - jobs
  - jobs/finalizers
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
```

Which seemed to help.

Found some more instances where the permissions are needed: for backup `backup` objects. I'd suggest to go through the code and add the finalizer permissions accordingly for object where we use it.

## Additional context

Seems only to happen on 3.11 as we have other deployments on k8s that don't suffer this issue.

## Expected behavior

K8up can create jobs on OpenShift 3.11 without issues.

## To Reproduce

1. Install k8up via helmfile on OpenShift 3.11
2. Run a backup
3. The job won't be created and k8up will retry every 30 seconds (as expected)

- Image Version: v1.0
- K8s Version: 1.11
- K8s Distribution: OpenShift 3.11

ccremer:
## Summary

* Fixes #379 
* Also see https://sdk.operatorframework.io/docs/faqs/

## Checklist

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -33,6 +33,7 @@ rules:
 - apiGroups:
   - backup.appuio.ch
   resources:
+  - archives/finalizers
   - archives/status
   verbs:
   - get
@@ -53,6 +54,7 @@ rules:
 - apiGroups:
   - backup.appuio.ch
   resources:
+  - backups/finalizers
   - backups/status
   verbs:
   - get
@@ -73,6 +75,7 @@ rules:
 - apiGroups:
   - backup.appuio.ch
   resources:
+  - checks/finalizers
   - checks/status
   verbs:
   - get
@@ -111,6 +114,7 @@ rules:
 - apiGroups:
   - backup.appuio.ch
   resources:
+  - prebackuppods/finalizers
   - prebackuppods/status
   verbs:
   - get
@@ -131,6 +135,7 @@ rules:
 - apiGroups:
   - backup.appuio.ch
   resources:
+  - prunes/finalizers
   - prunes/status
   verbs:
   - get
@@ -151,6 +156,7 @@ rules:
 - apiGroups:
   - backup.appuio.ch
   resources:
+  - restores/finalizers
   - restores/status
   verbs:
   - get
@@ -171,6 +177,7 @@ rules:
 - apiGroups:
   - backup.appuio.ch
   resources:
+  - schedules/finalizers
   - schedules/status
   verbs:
   - get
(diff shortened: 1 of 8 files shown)


tobru:
@Kidswiss Would you mind giving that one a review?

Kidswiss:
I amended #379 a bit yesterday.

Looks like other finalizers are also missing. At least for the backup object it also needs the finalizer RBAC. I added it the finalizer RBAC to all CRDs as well, but we need to check what CRDs actually use finalizers and amend accordingly.

Also I feel like we need some automated testing for 3.11 clusters, too. We still maintain quite a lot of those with k8up...

ccremer:
> Also I feel like we need some automated testing for 3.11 clusters, too. We still maintain quite a lot of those with k8up...

We [are already testing e2e-testing for K8s 1.13](https://github.com/vshn/k8up/blob/master/.github/workflows/e2e.yml#L16). But we lack actual test cases, the first test is #372

Kidswiss:
Is that just a 1.13 K8s cluster or an actual OpenShift? I assume the finalizer RBAC stuff is OpenShift specific because we haven't encountered it before...

ccremer:
Unfortunately, K8s 1.13 is the closest we can get to OpenShift 3.11. It's just so horribly old it's nowhere supported with tooling anymore.

ccremer:
@Kidswiss can you re-review?

Kidswiss:
The we should get rid of the OS 3.11 clusters ASAP :D

Kidswiss (approved):
LGTM

PR #384, opened 2021-03-03
Rework job status and PreBackup handling

### LINKED ISSUE #370: stuck PreBackupPod can cause long backlog of backups which then all start at the same time
## Describe the bug

If a namespace has a PreBackupPod defined and that Pod is stuck in creation for any reason (in our situation the PVC could not been mounted), k8up will just wait forever until this pod has started, sometimes days. In the meantime additional backups Objects are created from the Schedule and are all just waiting for the stuck pod.
As soon as the stuck pod is deleted (either manually or during a maintenance) the k8up operator starts all waiting backups at the same time, which in some cases can be hundreds of pods (like if you do backups every 30mins and the backups where stuck for a week).
This causes a huge influx of pods and also many backups at the same time, which I don't think is good for the clusters. See screenshot:

Lens_-_The_Kubernetes_IDE

I was able to capture this on video as well:

https://drive.google.com/file/d/173ySK7sJz9IUlTRMRC24BNbY1Y3U9-EQ/view?usp=sharing

## Expected behavior

I'm not exactly sure what I would expect, I could imagine two things:
1. the k8up should realize that an PreBackupPod is stuck and maybe react to that right when the next backup should be started or so. As waiting will just make it worse (more backups piling up). I could imagine that if a new backup is about to be started that the old stuck backup is maybe cancelled (especially if the pod is not created yet, meaning that we know we didn't start a backup yet, so the risk of cancelling a running backup is not there)
2. if we are in a situation where many backups are waiting, it should run them one by one and not at the same time

## To Reproduce

1. create a schedule that runs every 5 mins with PreBackupPod
2. make one PreBackupPod Stuck in creation (not exactly sure how to reproduce, maybe delete the underlining PV for the PVC of the PreBackupPod?)
3. wait 60 mins
4. delete stuck PreBackupPod
5. there should be 11 backups started at the same time

- Image Version: quay.io/vshn/k8up:v1.0.0-rc6
- K8s Version: v1.18
- K8s Distribution: AWS EKS

ccremer:
## Summary

* Don't watch a single PreBackup Deployments with its own kubeclient anymore. Instead, the readiness is checked upon every reconcile (30s, polling). This also gets rid of go routines within reconciles, thus reducing risk of memory leak due to stuck goroutines.
* Fully integrate the prebackup business logic into the backup executor.
* Split helper functions from `backup.go` to `backup_utils.go` to keep business logic somewhat easier to read
* More integration tests. Also split some helper functions in a new file.
* Jobs are now using Conditions to determine whether they have started or finished, not using `status.started` or `status.finished` anymore. Though they are still being set, but they can be considered deprecated.

Closes #370 

## Checklist

- [x] Update tests.


executor/prebackup.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+		if exists {
+			deployment, err := b.fetchPreBackupDeployment(deployment.Name, deployment.Namespace)
+			if err != nil {
+				return false, err
+			}
+			existingDeployments = append(existingDeployments, deployment)
+		} else {
+			b.Log.Info("started pre backup pod", "namespace", deployment.Namespace, "name", deployment.Name)
+		}
```
cimnine, line 54: Maybe switch this:

```suggestion
		if !exists {
			b.Log.Info("started pre backup pod", "namespace", deployment.Namespace, "name", deployment.Name)
			continue
		}
		
		deployment, err := b.fetchPreBackupDeployment(deployment.Name, deployment.Namespace)
		if err != nil {
			return false, err
		}
		existingDeployments = append(existingDeployments, deployment)
```

executor/prebackup.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	if ready, err := b.allDeploymentsAreReady(existingDeployments); ready {
+		b.Log.Info("pre backup pod(s) now ready")
+		b.SetConditionTrue(k8upv1alpha1.ConditionPreBackupPodsReady, k8upv1alpha1.ReasonReady)
+		return true, nil
+	} else if err == nil {
+		b.SetConditionUnknownWithMessage(k8upv1alpha1.ConditionPreBackupPodsReady, k8upv1alpha1.ReasonWaiting, "waiting for %d PreBackupPods to become ready", len(deployments))
+	}
```
cimnine, line 62:
```suggestion
	ready, err := b.allDeploymentsAreReady(existingDeployments)
	if err != nil {
		return false, nil
	}
	if ready {
		b.Log.Info("pre backup pod(s) now ready")
		b.SetConditionTrue(k8upv1alpha1.ConditionPreBackupPodsReady, k8upv1alpha1.ReasonReady)
		return true, nil
	}
	b.SetConditionUnknownWithMessage(k8upv1alpha1.ConditionPreBackupPodsReady, k8upv1alpha1.ReasonWaiting, "waiting for %d PreBackupPods to become ready", len(deployments))
```

In your solution you check for `ready` but never check if `err != nil`. This – to me – is at least not ideomatic, but it might even be wrong here. At least I wouldn't rely on `ready == false && err != nil` being `true` in all cases and from here on out.

executor/prebackup.go
```diff
@@ -0,0 +1,142 @@
...
+// If one of the deployment has failed (determined via "Progressing" condition), then it returns false with an error.
+// In that case, a backup is considered failed completely and status conditions will be set accordingly.
+// If one deployment is neither failed nor ready, then it returns false without error, indicating that it's still waiting.
+func (b *BackupExecutor) allDeploymentsAreReady(deployments []*appsv1.Deployment) (bool, error) {
+	readyCounter := 0
+	for _, deployment := range deployments {
+		ready, err := isPreBackupDeploymentReady(deployment)
+		log := b.Log.WithValues("preBackup", deployment.Name+"/"+deployment.Namespace)
+		if err != nil {
+			log.Info("backup failed: deadline exceeded on pre backup deployment")
```
cimnine, line 76: Perhaps you could add some more information to this log statement, like which pre-backup deployment's deadline has exceeded.

executor/prebackup.go
```diff
@@ -0,0 +1,142 @@
...
+			b.deletePreBackupDeployment(deployment)
+			return false, err
+		}
+		if ready {
+			log.V(2).Info("pre backup pod is in ready state")
+			readyCounter++
+		} else {
+			log.Info("still waiting on pre backup pod...")
+			return false, nil
+		}
```
cimnine, line 88:
```suggestion
		if !ready {
			log.Info("still waiting on pre backup pod...")
			return false, nil
		}
		
		log.V(2).Info("pre backup pod is in ready state")
		readyCounter++
```

I prefer writing these if-situations as suggested, because I find it reduces my mental load when reading the code. Because every condition opens a new branch in my mental model, and a `return` gets rid of it. So if a certain condition is not met and a return happens, that mental branch can be reduced again.

executor/prebackup.go
```diff
@@ -0,0 +1,142 @@
...
+			b.SetConditionFalseWithMessage(k8upv1alpha1.ConditionPreBackupPodsReady, k8upv1alpha1.ReasonFailed, err.Error())
+			b.SetConditionTrueWithMessage(k8upv1alpha1.ConditionReady, k8upv1alpha1.ReasonFailed, err.Error())
+			b.deletePreBackupDeployment(deployment)
+			return false, err
+		}
+		if ready {
+			log.V(2).Info("pre backup pod is in ready state")
+			readyCounter++
+		} else {
+			log.Info("still waiting on pre backup pod...")
```
cimnine, line 86:
```suggestion
			log.Info("still waiting on pre backup pod...")
```

Perhaps you could add here which pre backup pod we're currently wating for. It could make it easier to debug a scenario where such a pre backup pod is stuck.

executor/prebackup.go
```diff
@@ -0,0 +1,142 @@
...
+			return false, nil
+		}
+	}
+	return readyCounter == len(deployments) && readyCounter > 0, nil
```
cimnine, line 90: I would invert this: Instead of a `readyCounter` you could use a `readyCountdown`:

Initialize the countdown with `len(deployments)` and subtract 1 for every deployment that is ready.
If – in the end – the countdown is 0, then all deployments are ready.

cimnine: After looking at it again, I think it's redundant anyway. Because every non-ready condition leads to an abort of the `for` loop. So if I see this correctly, neither a countdown nor a countup is required and this can be shortened to:

```suggestion
	return true, nil
```

executor/prebackup.go
```diff
@@ -0,0 +1,142 @@
...
+	for _, deployment := range deployments {
+		// Avoid exportloopref
+		deployment := deployment
+		b.deletePreBackupDeployment(&deployment)
+	}
+
+	b.SetConditionTrue(k8upv1alpha1.ConditionPreBackupPodsReady, k8upv1alpha1.ReasonReady)
+}
+
+// deletePreBackupDeployment deleted the given deployment, if existing.
```
cimnine, line 130:
```suggestion
// deletePreBackupDeployment deletes the given deployment, if existing.
```

executor/prebackup_utils.go
```diff
@@ -0,0 +1,121 @@
...
+				Template: template.Spec.Pod.PodTemplateSpec,
+				Selector: &metav1.LabelSelector{
+					MatchLabels: podLabels,
+				},
+			},
+		}
+
+		err := controllerutil.SetOwnerReference(b.Config.Obj.GetMetaObject(), &deployment, b.Scheme)
+		if err != nil {
+			b.Config.Log.Error(err, "cannot set the owner reference", "name", b.Config.Obj.GetMetaObject().GetName(), "namespace", b.Config.Obj.GetMetaObject().GetNamespace())
```
cimnine, line 69: Should this perhaps be either a skip or an abort condition? Because if we can't set the owner reference, this object will not be garbadge collected by k8s.

I think that skipping this particular deployment or straight-out failing this function might be the safer choice.

cimnine (commented):
I realy like this PR, as mentioned in the chat. It's concise and has a good level of abstraction, and these abstractions have clear names. The tests are particularly nice to read. It's almost like a test-DSL has evolved. Very cool.

My comments and suggestions are mostly of only only minor importance; their mostly about coding style details and logging.

api/v1alpha1/common_types.go
```diff
@@ -52,6 +53,8 @@ const (
 	ConditionScrubbed ConditionType = "Scrubbed"
 	// ConditionProgressing is given when the resource is in the process of doing its main function.
 	ConditionProgressing ConditionType = "Progressing"
+	// ConditionPreBackupPodsReady is True if Deployments for all Container definitions were created and are ready
+	ConditionPreBackupPodsReady ConditionType = "PreBackupPodsReady"
```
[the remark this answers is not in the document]
ccremer: hmm, not necessarily :D
I guess my brain thought it's normal to have multiple prebackup deployments, but that's probably a very case.

api/v1alpha1/common_types.go
```diff
@@ -72,6 +75,11 @@ const (
 	// ReasonRetrievalFailed indicates that dependent resource(s) could not be retrieved for further processing
 	ReasonRetrievalFailed ConditionReason = "RetrievalFailed"
 
+	// ReasonNoPreBackupPodsFound is given when no PreBackupPods are found in the same namespace
+	ReasonNoPreBackupPodsFound ConditionReason = "NoPreBackupPodsFound"
+	// ReasonWaiting is given when PreBackupPods are waiting to be started
+	ReasonWaiting ConditionReason = "Waiting"
```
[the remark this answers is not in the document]
ccremer: My thought was that this reason would be reusable for other conditions also (similar to "ready", "succeeded" or "failed"). Since currently this reason is only used in the `PreBackupPodReady` condition, this could be renamed. But OTOH, it should be clear enough from the condition's name and the message that the backup is waiting for a prebackup.

`NoPreBackupPodsFound` is the only reason the term `PreBackup` is mentioned, since none of the other reasons fit. 
But I have no hard feelings here.

api/v1alpha1/common_types.go
```diff
@@ -113,3 +121,60 @@ type Status struct {
...
+}
+
+func matchAnyReason(condition metav1.Condition, reasons ...ConditionReason) bool {
+	for _, reason := range reasons {
+		if condition.Reason == reason.String() {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
ccremer: I wonder if I should move the functions `SetStarted`, `SetFinished` also to the Status struct? Or at least the bits that set the conditions? We would have "getter" and "setter" in one place.

executor/prebackup.go
ccremer: jup, that's easier to read. thanks!

executor/prebackup.go
ccremer: that's set with `b.Log.WithValues("preBackup", deployment.Name+"/"+deployment.Namespace)` right above, but in its own log field compared to encapsulated in the message.

executor/prebackup.go
ccremer: that is also set from `b.Log.WithValues("preBackup", deployment.Name+"/"+deployment.Namespace)` above. The return value of `WithValues` is a logger instance that has this field set in all following Info() calls

executor/prebackup.go
ccremer: you might be right, I'll check

executor/prebackup_utils.go
ccremer: tbh, I don't see a reason this error can ever occur. The docs says that the error can occur if there is another owner that has the controller flag set. If K8up creates a new deployment, there can't be another owner anyway. Since we never update a deployment (only create or delete), there shouldn't be multiple owners.
Thinking about it, I would even go as far to write `_ = controllerutil.SetOwnerReference()...`

Kidswiss:
Unfortunately I wasn't able to invest much time in this review.

I just skimmed through the code and looked at what the code does and not how it does it. I have to agree with @cimnine this makes it look cleaner.

Kidswiss (approved):

executor/prebackup.go
ccremer: I was able to optimize even more and combine into `fetchOrCreate` function, so no additional list either. should be even more understandable now

api/v1alpha1/common_types.go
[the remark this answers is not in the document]
ccremer: there, it's done. WDYT?

executor/prebackup_utils.go
cimnine: So I have looked into it. There's no unexpected API call, it's all local code. I'd suggest to leave it like it is then.

api/v1alpha1/common_types.go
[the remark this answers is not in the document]
cimnine: You will have to call these via `c.Obj().GetStatus()` then. I think the argument is valid, but I don't think that the functions are on the wrong object either. The logic for me goes that "I want to update the _Resource_, so that it's status indicates that the corresponding action was successful". We have this method on the _Resource_ now, which seems perfectly fine to, as _Status_ is just an implementation detail of the _Resource_. On the other hand, it would also be fine in _Status_, as it's the actual implementation.

For now, I'd leave them where they are, because they are used all over the place.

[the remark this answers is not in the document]
cimnine: FYI: I didn't see your "it's done" when I typed my comment.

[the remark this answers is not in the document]
cimnine: Ok. So, disregard my lengthy comment https://github.com/vshn/k8up/pull/384#discussion_r587652912. It's a nice and clean implementation now. 👍

cimnine (approved):

PR #387, opened 2021-03-05
E2E-Test for Restore operation

cimnine:
## Summary

E2E-tests a Restore.

Todo:
- [x] Verify that the restore was actually successful

Relies on https://github.com/vshn/k8up/pull/388
Closes https://github.com/vshn/k8up/issues/272

## Checklist

- [x] Update tests.


e2e/definitions/backup_repository/README.md
```diff
@@ -0,0 +1,122 @@
...
+## Re-create the backup
+
+To create the backup, I ran the test `test-01-backup.bats`.
+But I wanted to have the string `MagicString` as content of the file `/data/expectation.txt` in the `subject` container,
+which is actually the `/expectation.txt` file in the `subject-pv`.
+So I changed the deployment of the subject:
+
+```patch
+--- a/e2e/definitions/subject/deployment.yaml
++++ b/e2e/definitions/subject/deployment.yaml
```
ccremer, line 24: Is this some sort of blog, manual or your personal notes? I'm not quite sure what the intent with this README is :)

ccremer (commented):
tbh I wouldn't want to maintain the restore test in this current PR.

It looks quite complicated to manually set up "golden" testdata for the restore test. By now I'd favor a restore test that just reuses the backup created from the backup test and accept that there is a test dependency for a successful backup test.

As a concrete suggestion, I'd be okay when the backup test creates a magic file that says "backup successful" and the restore test checks if that file exists before attempting a restore. That way, we can fail-fast the restore test saying that preconditions aren't met yet. Or [skip the test conditionally](https://bats-core.readthedocs.io/en/latest/writing-tests.html#skip-easily-skip-tests) so it doesn't count as failed, since one needs to fix the backup test before.
After all, _all_ tests need to pass before we merge a PR, so it doesn't really matter if a restore succeeds but a backup test fails.

For me, the cost of test isolation is too high in the current form.
WDYT @Kidswiss ?

e2e/Makefile
```diff
@@ -27,6 +27,7 @@ bats_args ?=
 test: export KUBECONFIG = $(KIND_KUBECONFIG)
 test: export E2E_IMAGE = $(E2E_IMG)
 test: export WRESTIC_IMAGE = $(WRESTIC_IMG)
+test: export BUSYBOX_IMAGE = $(BUSYBOX_IMG)
```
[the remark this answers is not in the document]
cimnine: Not realy.

e2e/definitions/backup_repository/README.md
cimnine: There are two intents:
- It's a note what this folder's purpose is, since it contains "weird" files.
- It's the instruction how to re-create that backup should it be needed.

cimnine: Would it be better to put it somewhere in the asciidoc?

cimnine:
> By now I'd favor a restore test that just reuses the backup created from the backup test and accept that there is a test dependency for a successful backup test.

I dislike this idea, because it makes it impossible to distinguish wheter
- the Backup test failed, but it was not noticed, and therefore the Restore failed
- or whether the restore test failed on it's own.

> We also need to consider changes in the restic code or data structures that we could need to maintain over version updates.

Yes, I agree. Maybe I should create a script that updates the test backup instead of writing it down for a human to do.

On the other hand, when you create a backup _in the real world_, you also don't want it to break when a new restic version is released. Old backups have to be restorable with newer versions anyway.

So I rate this argument as not being too important.

> We should rather treat restic itself as a black box and not need to maintain golden testdata as well.

It's treatet as a black box. Just instead of the backup being generated by restic every time from scratch, a well-defined state is imported to the S3 `backup` bucket. This does not rely on restic at all.

>  After all, all tests need to pass before we merge a PR, so it doesn't really matter if a restore succeeds but a backup test fails.

Yes, but: If you see that the Restore test failed, you want to look into the Restore test. You don't want to find out after some time of debugging that the Restore test is actually fine, but the Backup test failed in an unpredicted way.

---

Now with all this said, I see arguments for both sides. I built it this way, because I find it easier to formulate an expectation for a known and static state, rather than a dynamic state. I will now (or probably tomorrow) go about to change it.

ccremer:
> I dislike this idea, because it makes it impossible to distinguish wheter
> 
>     * the Backup test failed, but it was not noticed, and therefore the Restore failed
>     * or whether the restore test failed on it's own.

Why do we keep distinguishing a backup from a restore? a backup is only as good its restore. If a backup fails, the restore is worth nothing. If a restore fails but a backup works, the backup is worth nothing either.

Consider a new restic version. The backup fails but restore succeeds, we need to fix the backup, but also (maybe) recreate new testdata. Then the restore might also fail, possibly needing a fix too.
If the restore fails but backup succeeds, we need to recreate testdata and fix the restore. 

With the dynamic testdata approach, we need to fix both restore and backup. But we'd know already ahead. I just think that we are going to have more of integration problems with K8up and less one with restic.

> This does not rely on restic at all.

It does. It relies on the fact that we have correctly checked in that well-defined, black box state and data structure that may or may not be version dependent since it has become static.

I also see both sides. But I don't think it's K8up's job to test restic, only it's integration with K8up and wrestic. I trust that restic maintainers test their application and version compatibilities on their own.

I guess it all depends on what we want to test. Do we just test a case where we backup & restore an app so we can verify K8up's integration with restic, or do we want to ensure that restic backups and restore done with K8up have valid integrity?

The original issue #272 is targeted for developers to prevent broken integration between K8up & (w)restic. It was never about valid backups. Should we reword it so that it aims for users and valid integrity instead?

Can we get @Kidswiss involved here for a democratic/quorum decision and be done with it? :D

e2e/definitions/backup_repository/README.md
ccremer: I think so. I remember some discussion that we want to create a developer guide/section anyway.

Could we at least have a script that does the recreation for us?

cimnine:
I will go with a middle ground. The backup will not be loaded from a fixed snapshot, but will be created on the spot. But this will be done as part of the Restore test and not rely on the Backup test.

Like the K8up deployment is tested on it's own and both – the Restore test and the Backup test – rely on their own deployment of K8up.

cimnine:
I've pushed a new solution as described above. It doesn't currently show up as GitHub seems [to have some difficulties right now](https://www.githubstatus.com/incidents/wdhkj70kynwk).

e2e/lib/k8up.bash
```diff
@@ -20,18 +24,23 @@ setup() {
...
 	reset_debug
+	clear_pv_data
 }
 
 teardown() {
 	cp -r /tmp/detik debug || true
 }
 
+clear_pv_data() {
+	rm -rfv ./debug/data/*
```
ccremer, line 35: this doesn't work, because this also deletes `./debug/data/pvc-subject`, which is set up by `make` with our local user id. If we remove `pvc-subject`, then KIND will create this path when a PV is created, but with root permissions.

This cascades to the point where the subject deployment cannot create the test file with tee:
`tee: /data/old_file.txt: Permission denied`
and it crashloops.

you can easily reproduce this with `make clean e2e-test`

e2e/test-02-restore.bats
```diff
@@ -0,0 +1,37 @@
...
+	new_content="New content: $(timestamp)"
+	new_filename="new_file.txt"
+	given_a_clean_ns
+	given_a_subject "${new_filename}" "${new_content}"
+
+	# Restore
+	apply definitions/restore
+	try "at most 10 times every 1s to get restore named 'k8up-k8up-restore' and verify that '.status.started' is 'true'"
+	try "at most 10 times every 1s to get job named 'k8up-k8up-restore' and verify that '.status.active' is '1'"
+	wait_until restore/k8up-k8up-restore completed
```
ccremer, line 33: Hm, we might need to verify the condition with the `try` syntax. Because the condition can be "Completed" but it's the `.reason` field that says whether it was successful or not.

ccremer (changes requested):

e2e/lib/k8up.bash
```diff
@@ -129,10 +158,64 @@ given_a_running_operator() {
...
+
+	commands=(
+		"ls -la \"$(dirname "${expected_file}")\""
+		"test -f \"${expected_file}\""
+		"cat \"${expected_file}\""
+		"test \"${expected_content}\" \"=\" \"\$(cat \"${expected_file}\")\" "
+	)
+
+	echo "Testing if file '${expected_file}' contains '${expected_content}'"
+	echo "in container '${container}' of pod '${pod}'."
```
ccremer, line 208: I would advise to keep single log statement in single echo call. Our target screens aren't mobile phones ;)
```suggestion
	echo "Testing if file '${expected_file}' contains '${expected_content}' in container '${container}' of pod '${pod}'."
```

e2e/lib/k8up.bash
```diff
@@ -129,10 +158,64 @@ given_a_running_operator() {
...
+		errcho "$0 Expected 2 arguments, got ${#}."
+		exit 1
+	fi
+
+	backup_file_name=${1}
+	backup_file_content=${2}
+	given_a_subject "${backup_file_name}" "${backup_file_content}"
+
+	apply definitions/backup
+	wait_until backup/k8up-k8up-backup completed
```
ccremer, line 172: same here: the `.reason` field of the condition says something about its success. "completed" condition being true only means it is now finished with some result.

In hindsight, maybe it's best if we stay with the `try` syntax for consistency reasons?

e2e/test-01-backup.bats
```diff
@@ -12,10 +12,13 @@ DETIK_CLIENT_NAMESPACE="k8up-e2e-subject"
...
+	expected_filename="expected_filename.txt"
+
 	given_a_running_operator
 	given_a_clean_ns
 	given_s3_storage
-	given_a_subject
+	given_a_subject "${expected_filename}" "${expected_content}"
 
 	apply definitions/backup
 	try "at most 10 times every 1s to get backup named 'k8up-k8up-backup' and verify that '.status.started' is 'true'"
```
ccremer, line 24: Line 26 (can't select that in GH diff) with `try "at most 10 times every 2s to find 1 pod named 'k8up-k8up-backup-.....' with 'status' being 'Running'"` sometimes fails if the pod is `Succeeded` before the `try` is querying the status. I would just remove it and rely on the next assertion.

ccremer (approved):
thanks for the work! :tada:

PR #388, opened 2021-03-05
Refactor e2e-test utils

cimnine:
## Summary

Some changes in the support code of the e2e tests in preparation for https://github.com/vshn/k8up/pull/387.
and therefore related to #272.

## Checklist

- [x] Update tests.

### DIFF
--- a/e2e/Makefile
+++ b/e2e/Makefile
@@ -28,7 +28,7 @@ test: export KUBECONFIG = $(KIND_KUBECONFIG)
 test: export E2E_IMAGE = $(E2E_IMG)
 test: export WRESTIC_IMAGE = $(WRESTIC_IMG)
 test: setup kind-e2e-image ## Run the E2E tests
-	@$(bats) . $(bats_args)
+	@$(bats) $(bats_args) $(BATS_FILES)
 
 kind-e2e-image: kind-setup ## Load the e2e container image onto e2e cluster
 	@$(KIND) load docker-image --name $(KIND_CLUSTER) $(E2E_IMG)
--- a/e2e/lib/k8up.bash
+++ b/e2e/lib/k8up.bash
@@ -31,10 +31,11 @@ kustomize() {
 }
 
 restic() {
-	kubectl run wrestic \
+	kubectl run "wrestic" \
 		--rm \
 		--attach \
 		--restart Never \
+		--wait \
 		--namespace "${DETIK_CLIENT_NAMESPACE-"k8up-system"}" \
 		--image "${WRESTIC_IMAGE}" \
 		--env "AWS_ACCESS_KEY_ID=myaccesskey" \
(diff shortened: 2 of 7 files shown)


e2e/lib/k8up.bash
```diff
@@ -29,7 +36,7 @@ kustomize() {
 }
 
 restic() {
-	kubectl run wrestic \
+	kubectl run "wrestic-$(randstr 5)" \
```
ccremer, line 39: why the randomized suffix?

ccremer (commented):
the GH jobs seem to fail. Any leads?

e2e/lib/k8up.bash
cimnine: Because it seams that `kubectl` returns befor the Pod it creates is actually removed. So, when instantly after `kubectl` returns a new on-demand Pod with the same name is created, then it fails with a `Pod already exists` error.

cimnine:
> the GH jobs seem to fail. Any leads?

I'm looking into it right now. Locally, it still works 🤔

e2e/lib/k8up.bash
cimnine: Since `$(</dev/urandom)` seems to not work in GH Actions, I've had to change this slightly to rely on timestamps instead of a random string. I've also added some explanation to the commit that introduces these suffixes.

cimnine:
It's now green 🟢

e2e/lib/k8up.bash
ccremer: Have you tried the `--wait` flag?
> If true, wait for resources to be gone before returning. This waits for finalizers.

e2e/Makefile
```diff
@@ -54,5 +54,10 @@ help: ## Show this help
...
 ###
 
-$(bats):
+$(bats): package-lock.json
 	@npm install
+
+package-lock.json: package.json
+	@npm install
+
+package.json:
```
ccremer, line 63: This causes npm to be installed twice per `make e2e-test` run. Is that intended?
I would expect that it only installs packages when bats isn't installed yet.

```console
$ make e2e-test
make[1]: Entering directory '/home/chris/Documents/source/vshn/k8up/e2e'

up to date, audited 2 packages in 576ms

found 0 vulnerabilities
make[1]: Leaving directory '/home/chris/Documents/source/vshn/k8up/e2e'
go build -o /home/chris/Documents/source/vshn/k8up/./k8up main.go
...
make[1]: Entering directory '/home/chris/Documents/source/vshn/k8up/e2e'

up to date, audited 2 packages in 584ms

found 0 vulnerabilities
 ✓ lint assertions 
 ✓ verify the deployment 
 ✓ verify a backup 

3 tests, 0 failures
```

e2e/lib/k8up.bash
```diff
@@ -74,11 +97,16 @@ apply() {
 	kubectl apply -f "debug/${1}/main.yml"
 }
 
-given_a_subject() {
-	kubectl delete namespace k8up-e2e-subject --ignore-not-found
+given_a_clean_ns() {
+	kubectl delete namespace "${DETIK_CLIENT_NAMESPACE}" --ignore-not-found
 	kubectl delete pv subject-pv --ignore-not-found
-	kubectl create namespace k8up-e2e-subject || true
+	kubectl create namespace "${DETIK_CLIENT_NAMESPACE}" || true
```
ccremer, line 103: Probably a more reliable command, in terms of exit code.
```suggestion
	kubectl create namespace --save-config --dry-run=client -o yaml "${DETIK_CLIENT_NAMESPACE}" | kubectl apply -f -
```

ccremer: Actually, if we delete the namespace before, why is there a `|| true` ?

e2e/Makefile
cimnine: 🤔

I've deleted it for now

e2e/lib/k8up.bash
cimnine: I think so as well.

ccremer (approved):

PR #400, opened 2021-03-16
Actually Perform the Cleanup of Checks

### LINKED ISSUE #393: K8up does not clean up check jobs
## Describe the bug

I set the env variable `BACKUP_GLOBALKEEPJOBS`.
K8up cleans then `prune` and `backup` jobs, but not `check` jobs.

## Expected behavior

If `BACKUP_GLOBALKEEPJOBS` is set, k8up will cleanup `backup`, `check` and `prune` jobs.

- Image Version: e.g. v1.0.2
- K8s Version: v1.20.4+k3s1
- K8s Distribution: k3s

cimnine:
## Summary

It was discovered that the cleanup of Checks was not implemented. This PR adds this missing functionality.

Closes #393

## Checklist

- [x] Update tests.

### DIFF
--- a/controllers/check_integration_test.go
+++ b/controllers/check_integration_test.go
@@ -1,0 +1,100 @@
+// +build integration
+
+package controllers_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/suite"
+	batchv1 "k8s.io/api/batch/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	k8upv1a1 "github.com/vshn/k8up/api/v1alpha1"
+	"github.com/vshn/k8up/controllers"
+)
+
+type CheckTestSuite struct {
+	EnvTestSuite
+
+	GivenCheck *k8upv1a1.Check
+	CheckName  string
+}
+
+func Test_Check(t *testing.T) {
+	suite.Run(t, new(CheckTestSuite))
+}
+
+func (c *CheckTestSuite) TestReconciliation() {
+	c.givenCheckResource()
+
+	result := c.whenReconcile()
+
+	c.Assert().GreaterOrEqual(result.RequeueAfter, 30*time.Second)
+	c.expectAJobEventually()
+}
+
+func (c *CheckTestSuite) BeforeTest(suiteName, testName string) {
+	c.CheckName = "check-integration-test"
+}
+
+func NewCheckResource(restoreName, namespace string) *k8upv1a1.Check {
+	keepJobs := 5
+	return &k8upv1a1.Check{
+		Spec: k8upv1a1.CheckSpec{
+			KeepJobs: &keepJobs,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      restoreName,
+			Namespace: namespace,
+		},
+	}
+}
+
+func (c *CheckTestSuite) givenCheckResource() {
+	c.GivenCheck = NewCheckResource(c.CheckName, c.NS)
+	c.EnsureResources(c.GivenCheck)
+}
+
+func (c *CheckTestSuite) whenReconcile() controllerruntime.Result {
+	controller := controllers.CheckReconciler{
+		Client: c.Client,
+		Log:    c.Logger,
+		Scheme: c.Scheme,
+	}
+
+	key := types.NamespacedName{
+		Namespace: c.NS,
+		Name:      c.CheckName,
+	}
+	request := controllerruntime.Request{
+		NamespacedName: key,
+	}
+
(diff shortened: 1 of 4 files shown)


cimnine:
@tobru I've requested your review as I updated the GoLand instructions in the docs.

cimnine:
@ccremer I think codeclimate's duplicate code issue is not worth the effort, because the core part of the deduplication logic is already extracted to a common place. The part which is supposedly duplicated is the code that prepares the data structures to leverage the deduplication function. Because of Go's type system, it looks like an exact copy, yet it isn't.

WDYT?

cimnine:
It strikes me as weird that the diff-coverage did not increase at all, since I've added tests that cover code not tested before 🤔  Is integrationtest coverage perhaps not reported correctly?

tobru (approved):

controllers/check_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		c.T().Logf("%d deleted Checks found", amountOfDeletedItems)
+		done = amountOfDeletedItems == expectedDeletes
+
+		return
+	})
+}
+
+func (c *CheckTestSuite) whenJobCallbackIsInvoked(checkName string) {
```
ccremer, line 102: I recommend to rename this to `ts` to be consistent with other integration tests and give same dev experience.
```suggestion
func (ts *CheckTestSuite) whenJobCallbackIsInvoked(checkName string) {
```

controllers/check_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+	checkNSName := types.NamespacedName{Name: checkName, Namespace: c.NS}
+
+	check := &batchv1.Job{}
+	err := c.Client.Get(c.Ctx, checkNSName, check)
+	c.Require().NoError(err)
```
ccremer, line 107: there should be a `fetchResource()` function from EnvTestSuite already, saving you from some DRY

executor/check.go
```diff
@@ -48,12 +55,13 @@ func (c *CheckExecutor) Execute() error {
 	}
 	checkJob.GetLabels()[job.K8upExclusive] = "true"
 
-	return c.startCheck(checkObject, checkJob)
+	return c.startCheck(checkJob, checkObject)
 }
 
-func (c *CheckExecutor) startCheck(checkObject *k8upv1alpha1.Check, checkJob *batchv1.Job) error {
+func (c *CheckExecutor) startCheck(checkJob *batchv1.Job, checkObject *k8upv1alpha1.Check) error {
 	c.check = checkObject
```
ccremer, line 62: Line 62 assigns does `c.check = checkObject`, why can't we do that within Execute() already? that would save us a parameter

ccremer:
> Because of Go's type system, it looks like an exact copy, yet it isn't.

In CodeClimate UI, you can mark this issue as invalid or wontfix with your reasoning. then it becomes green and should be saved for later PRs as well.

ccremer:
> It strikes me as weird that the diff-coverage did not increase at all, since I've added tests that cover code not tested before :thinking: Is integrationtest coverage perhaps not reported correctly?

I'd recommend to run `make integration-test` once in current master branch and then compare with your PR. you could also run all integration tests with GoLand with Coverage enabled to see which parts of the code gets executed and which not. It's possible that some other integration test is going through the same code locations already, but it still shouldn't be 0% I guess...

Be aware that code coverage in Goland is not calculated the same as gotest but it may give some clue.

cimnine:
> In CodeClimate UI, you can mark this issue as invalid or wontfix with your reasoning. then it becomes green and should be saved for later PRs as well.

I'm afraid, I don't have these privileges.

cimnine:
> I'd recommend to run make integration-test once in current master branch and then compare with your PR.

It increases form 30% to 38%.

> you could also run all integration tests with GoLand with Coverage enabled

For `controllers/check.go`, it increases from 0% to 60%.

I think that's a good-enough ratio :)

docs/modules/ROOT/pages/explanations/ide.adoc
```diff
@@ -9,18 +9,18 @@ You can basically just open the root directory in GoLand and you'll be mostly re
...
 
-Now try to re-run the test.
-It should now succeed.
-You will see test results in your IDE and be able to use the integrated debugger as well.
+You can use all the usual shortcuts to quickly run a test (e.g. kbd:[Ctrl+Shift+D] to debug a specific test method).
+If it doesn't work, check the following:
+
+. Check that the _Custom tags_ field in _Preferences_ → _Go_ → _Build Tags & Vendoring_ contains to term `integration`.
+. Make sure the _Working directory_ of your test's _Run Configuration_ points to the directory where your test lies,
+  e.g. `/home/your_user/projects/k8up/controllers`.
```
ccremer, line 22: We have a one-sentence-per-line styleguide, so this should be all one line
```suggestion
. Make sure the _Working directory_ of your test's _Run Configuration_ points to the directory where your test lies, e.g. `/home/your_user/projects/k8up/controllers`.
```

ccremer:
The coverage is indeed strange...
well, what matters is that we have a working test case :)

ccremer (approved):

PR #410, opened 2021-03-24
BackupCommand Annotation E2E Test

cimnine:
## Summary

This PR adds an E2E-test to verify the functionality of [the `backupcommand` annotation](https://k8up.io/k8up/1.0.0/references/annotations.html).

## Checklist

- [x] Update tests.


docs/modules/ROOT/pages/references/annotations.adoc
```diff
@@ -14,7 +14,7 @@ See xref:references/config-reference.adoc[Configuration reference]
 |`k8up.syn.tools/backup`
 |If defined, this influences whether this pod shall be backed up (`true`) or not (`false`)
 |Either `'true'` or `'false'`
-|`Pod`
+|`Pod`, `PersistenVolumeClaim`
```
ccremer, line 17:
```suggestion
|`Pod`, `PersistentVolumeClaim`
```

e2e/definitions/annotated-subject/pv.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: subject-pv
```
ccremer, line 4: The PV and PVC look very similar to the other subject.
I think it makes sense to create separate definition dir for them? and include via kustomization.yaml?

e2e/definitions/subject/deployment.yaml
```diff
@@ -22,6 +22,7 @@ spec:
         - -c
         - |
           printf '$BACKUP_FILE_CONTENT' | tee '/data/$BACKUP_FILE_NAME' && \
+          echo && \
```
ccremer, line 25: would adding newline to printf also work?
```suggestion
          printf '$BACKUP_FILE_CONTENT' \n | tee '/data/$BACKUP_FILE_NAME' && \
```

e2e/lib/k8up.bash
```diff
@@ -120,6 +120,19 @@ given_a_subject() {
 	echo "✅  The subject is ready"
 }
 
+given_an_annotated_subject() {
+	if [ "${#}" != 2 ]; then
+		errcho "$0 Expected 2 arguments, got ${#}."
+		exit 1
+	fi
```
ccremer, line 127: Not sure that if-clause is worth maintaining. We're writing tests here, not really a library. I would just assume that the tester supplies the required arguments by reading the code.

e2e/definitions/subject/deployment.yaml
ccremer: scratch that, not worth changing :)

cimnine: It would also change the behaviour. All of a sudden the `${BACKUP_FILE_CONTENT}` that is written to the file is actually `${BACKUP_FILE_CONTENT}\n`, which has to be compensated for in other places, like when comparing for equality. That was the point of using `printf '$BACKUP_FILE_CONTENT' in the first place. Otherwise I could have just used `echo '$BACKUP_FILE_CONTENT'`.

ccremer (approved):

PR #431, opened 2021-05-21
New configuration variable to pass options to restic

cimnine:
## Summary

On a global level, this PR allows to pass options down to restic.

Closes https://github.com/vshn/k8up/issues/424

## Checklist

- [ ] Update tests.

### DIFF
--- a/executor/generic.go
+++ b/executor/generic.go
@@ -204,6 +204,10 @@ func DefaultEnv(namespace string) EnvVarConverter {
 	defaults.SetString(cfg.AwsSecretAccessKeyEnvName, cfg.Config.GlobalSecretAccessKey)
 	defaults.SetString("HOSTNAME", namespace)
 
+	if cfg.Config.ResticOptions != "" {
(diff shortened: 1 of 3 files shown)


executor/generic.go
```diff
@@ -204,6 +204,10 @@ func DefaultEnv(namespace string) EnvVarConverter {
 	defaults.SetString(cfg.AwsSecretAccessKeyEnvName, cfg.Config.GlobalSecretAccessKey)
 	defaults.SetString("HOSTNAME", namespace)
 
+	if cfg.Config.ResticOptions != "" {
+		defaults.SetString(cfg.ResticOptionsEnvName, cfg.Config.ResticOptions)
+	}
```
ccremer, line 209: Is there a harm if the environment variable is defined but empty? We could skip the conditional

ccremer (approved):

ccremer:
It tested it locally with e2e tests. I can confirm that the RESTIC_OPTIONS variable is passed to pods spawned by K8up. I cannot confirm whether they actually work since a new release of wrestic image is needed

PR #434, opened 2021-05-26
Pin quay.io/vshn/wrestic to latest

ccremer:
## Summary

We should always test the newest version.
Will also reduce future version updates

## Checklist


Kidswiss (approved):
LGTM

PR #452, opened 2021-07-05
Remove duplicate keys initialDelaySeconds, timeoutSeconds

psy-q:
## Summary

This PR removes two duplicate keys from the MariaDB tutorial deployment's `spec.template.spec.readinessProbe` section. This way, the k8up tutorial works again as documented.

Otherwise kubectl throws an error:

```
error: map[string]interface {}(nil): yaml: unmarshal errors:
  line 37: mapping key "timeoutSeconds" already defined at line 27
  line 36: mapping key "initialDelaySeconds" already defined at line 28
```

At least when using kubectl 1.21.1 against apiserver 1.20.2. I'm not sure which part (kubectl or the API server) does these validations.

## Checklist


ccremer:
(Please label documentation fixes with documentation :) )

PR #455, opened 2021-07-08
Validate and update tutorial

psy-q:
## Summary

The tutorial had some parts that didn't match reality anymore. There were also some defects (missing files, wrong paths) from when the separate tutorial repo was merged in with this main repo. I tried running every step of the tutorial and fixed as I went along.

There is a pretty signifcant change to the way WordPress is restored: I couldn't get it to work as previously documented, nothing would be restored to the existing and bound PVC. A solution that worked was to bind another empty PVC to `/restore` and restore to this, then rsync the files using a shell in the container. The scripts have been adapted for this.

## Checklist

- [x] Keep pull requests small so they can be easily reviewed.  (Hmmm :worried:)


docs/modules/ROOT/pages/tutorials/tutorial.adoc
```diff
@@ -310,20 +315,38 @@ $ cp -R /data/wordpress-pvc/* .
...
+[source,yaml]
+....
+spec:
+  template:
+    metadata:
+      annotations:
+        k8up.syn.tools/backupcommand: appuio.ch/backupcommand: /bin/bash -c 'mysqldump -uroot -p"${MARIADB_ROOT_PASSWORD}" --all-databases'
+....
+
+Thus the restoration procedure is different. Instead of restoring files to a PVC, we can instead connect to the running database in the pod and restore directly from the dump.
```
ccremer, line 331: Following the [style guide](https://syn.tools/syn/references/style-guide.html), you should make 1 line per sentence
```suggestion
Thus the restoration procedure is different. 
Instead of restoring files to a PVC, we can instead connect to the running database in the pod and restore directly from the dump.
```

docs/modules/ROOT/pages/tutorials/tutorial.adoc
```diff
@@ -266,17 +269,19 @@ spec:
   backend:
 ....
 
+Don't forget to fill in your actual `backend:` information as documented above. There is a predefined restore YAML file in `../examples/restore/wordpress.yaml` where you only have to replace `SNAPSHOT_ID`.
```
ccremer, line 272: also here
```suggestion
Don't forget to fill in your actual `backend:` information as documented above. 
There is a predefined restore YAML file in `../examples/restore/wordpress.yaml` where you only have to replace `SNAPSHOT_ID`.
```

docs/modules/ROOT/pages/tutorials/tutorial.adoc
```diff
@@ -310,20 +316,39 @@ $ cp -R /data/wordpress-pvc/* .
...
 
 [start=5]
-. Delete the MariaDB pod and a new one will be immediately created due to the deployment policy (see `mariadb/deployment.yml`): `kubectl delete pod mariadb-XXXXXXXXXX-xxxx`
+. Now connect to the MariaDB pod, copy the dump and a simple restore script to the pod and execute it there:
+
+....
+$ export MARIADB_POD=$(kubectl get pods | grep mariadb | awk '{print $1}')
+$ kubectl cp backup.sql "$MARIADB_POD":/
+$ kubectl cp scripts/db_restore.sh "$MARIADB_POD":/
+$ kubectl exec "$MARIADB_POD" -- /db_restore.sh
```
ccremer, line 350: Isn't there are more direct way to get the pods? e.g.
`kubectl get pods -o custom-columns="NAME:.metadata.name" --no-headers -l app.kubernetes.io/name=mariadb`

or probably even `kubectl exec deploy/mariadb -- /db_restore.sh` (s/deploy/sts/ if statefulset) so you won't have to select a specific pod first.

psy-q: I took those lines pretty much verbatim from `scripts/4_restore.sh` but I see what you mean. I've adapted to: `kubectl get pods -o custom-columns="NAME:.metadata.name" --no-headers -l "app=wordpress,tier=mariadb"` since the  `-l app.kubernetes.io/name=mariadb` didn't seem to hit that pod (?)

`kubectl exec deploy/mariadb -- /db_restore.sh` does work but now that `$MARIADB_POD` is set, wouldn't it be OK to use it in all three lines?

ccremer: yeah this `-l app.kubernetes.io/name=mariadb` was just prosa guessing in the blue. I didn't actually know which labels ;)
ah I overlooked the kubectl cp part. `kubectl cp` doesn't seem to support the deploy/ syntax, at least not in `--help`.
sure, you could keep `kubectl exec $MARIADB_POD` for consistency, no hard feeling here.

docs/modules/ROOT/pages/tutorials/tutorial.adoc
```diff
@@ -298,32 +306,49 @@ Let's start with the WordPress pod:
 TIP: In K9s you can open a shell in a pod by selecting that pod and then hitting the kbd:[S] key.
 
 [start=2]
-. Replace the files inside the pod with the ones from the restore:
+. Replace the files inside the pod with the ones from the restore. Using rsync here simplifies things by only copying files that are different from the target and also deleting any additional malicious files an attacker may have added:
```
ccremer, line 309: should work as well when indented with 2 spaces
```suggestion
. Replace the files inside the pod with the ones from the restore. 
  Using rsync here simplifies things by only copying files that are different from the target and also deleting any additional malicious files an attacker may have added:
```

docs/modules/ROOT/pages/tutorials/tutorial.adoc
```diff
@@ -298,32 +306,49 @@ Let's start with the WordPress pod:
...
-$ cp -R /data/mariadb-pvc/* .
+$ export SNAPSHOT_ID=$(restic snapshots --json --last --path /default-mariadb | jq -r '.[0].id')
+$ restic dump "${SNAPSHOT_ID}" /default-mariadb > backup.sql
 ....
 
 [start=5]
-. Delete the MariaDB pod and a new one will be immediately created due to the deployment policy (see `mariadb/deployment.yml`): `kubectl delete pod mariadb-XXXXXXXXXX-xxxx`
+. Now connect to the MariaDB pod, copy the dump and a simple restore script to the pod and execute it there:
+
+....
```
ccremer, line 346: You can indent the code block and make it part of the numbering by adding the + between. Maybe you can then avoid having to specifically set the numbering via `[start=x]` in the next item
```suggestion
. Now connect to the MariaDB pod, copy the dump and a simple restore script to the pod and execute it there:
+
....
```

docs/modules/ROOT/pages/tutorials/tutorial.adoc
```diff
@@ -298,32 +306,49 @@ Let's start with the WordPress pod:
...
+
+Thus the restoration procedure is different. 
+Instead of restoring files to a PVC, we can instead connect to the running database in the pod and restore directly from the dump.
 
 [start=4]
-. Follow the same logic as previously:
+. Get the snapshot ID for the database dump and save it to your local machine as `backup.sql`:
 
 ....
```
ccremer, line 338: also here
```suggestion
. Get the snapshot ID for the database dump and save it to your local machine as `backup.sql`:
+
....
```

docs/modules/ROOT/pages/tutorials/tutorial.adoc
psy-q: Ah, alright :)

docs/modules/ROOT/pages/tutorials/tutorial.adoc
psy-q: Thanks, I have to get reacquainted with AsciiDoc.

docs/modules/ROOT/pages/tutorials/tutorial.adoc
```diff
@@ -298,32 +306,50 @@ Let's start with the WordPress pod:
...
-$ cd /var/lib/mysql # Attention! Different path!
-$ mv data /
-$ rm -r *
-$ cp -R /data/mariadb-pvc/* .
+$ export SNAPSHOT_ID=$(restic snapshots --json --last --path /default-mariadb | jq -r '.[0].id')
+$ restic dump "${SNAPSHOT_ID}" /default-mariadb > backup.sql
```
Kidswiss, line 341: We could probably pipe this directly to a `restic dump... | kubectl exec...` thus making this a oneliner and remove the whole copy stuff below.

Kidswiss:
I only had one remark so far. It's not really important, but if you have time/wish to experiment around feel free to try it out.

docs/modules/ROOT/pages/tutorials/tutorial.adoc
```diff
@@ -298,32 +306,50 @@ Let's start with the WordPress pod:
...
+. Replace the files inside the pod with the ones from the restore. 
+  Using rsync here simplifies things by only copying files that are different from the target and also deleting any additional malicious files an attacker may have added:
 
 ....
-$ cd /var/www/html # or whatever value has `mountPath`
-$ mv data /
-$ rm -r *
-$ cp -R /data/wordpress-pvc/* .
+# apt update && apt install rsync
+# rsync -a --delete /restore/ /var/www/html/
```
ccremer, line 314: Why using the # here?

Also, one thing to note here is that root user has to run in the container in order to install rsync. On OpenShift, you cannot install additional packages since we're running with a non-privileged user. Maybe add a note about this?
Or maybe simply something like `rm -r /var/www/html/* && cp -r -p /restore/* /var/www/html/` is sufficient ? (not tested command) 
```suggestion
  Using rsync here simplifies things by only copying files that are different from the target and also deleting any additional malicious files an attacker may have added:
+
[source,bash]
----
apt update && apt install rsync
rsync -a --delete /restore/ /var/www/html/ # or replaced with rm/cp
----
```

Kidswiss: This gets me thinking... Why the rsync stuff anyways?

Wouldn't it be easier to simply mount the PVC with the restored data to the deployment? The method here could be used for more granular restore (f.e. only one file). But I think it's a bit overkill for a tutorial?

ccremer: It could work for this particular deployment. But if it were a statefulset (mariadb?), then it depends how the STS is configured. If it's a volume template, then you'd have to use cp/rsync, or recreate the PVC before starting up the STS again. Did I get this right?

Kidswiss: Yeah, but in this tutorial it's not a statefulset? I feel like this overcomplicates things in a tutorial that should just show "hey look it could look like this to use it".

psy-q: I had the `#` because I got a root user when exec'ing /bin/bash in the container. But this is all going to change after my new discovery that the restore *does* work, it just doesn't delete files that don't exist in the snapshot... So I will replace the whole rsync section with just a sentence mentioning this.

That means that in the case of a defaced WordPress site where an attacker has placed nasty things into `wp-content` to redistribute them, these would not be deleted by a restore. So it's not really a restore to a pristine state like the trick with `rsync --delete` from a separate PVC would be. But it's probably better/less complicated for a tutorial.

Kidswiss: No biggie :)

In a case of a nasty attack nuking the PVCs and hosts would be advisable anyways. You can't really be sure where the attacker placed backdoors and other nasty things.

docs/modules/ROOT/pages/tutorials/tutorial.adoc
psy-q: I need to expand `$MARIADB_ROOT_PASSWORD` from the env inside the container for that. But just appending it to kubectl will expand it on the host system. I couldn't figure out if it's possible to escape it so it gets expanded inside?

What I came up with is this little adventure:

```bash
restic dump "${SNAPSHOT_ID}" /default-mariadb | kubectl exec -i "$MARIADB_POD" -- mysql -uroot --password=`kubectl exec "$MARIADB_POD" -- printenv "MARIADB_ROOT_PASSWORD"`
```

Not elegant but it works. Is there a better way?

ccremer: You could wrap it with `sh -c ...`
There's an example here: https://k8up.io/k8up/1.1.0/how-tos/prebackuppod.html

psy-q: Ah, alright! Thanks, changed it to that.

ccremer (approved):
LGTM. Didn't verify whether the tutorial works though.

Kidswiss:
LGTM

Maybe in the future we could migrate it away from minikube? Kind would be perfectly fine here.

Kidswiss (approved):

psy-q:
I spotted a wrong path from when I moved the backup example YAML but I think now I'm really really done with this :)

ccremer:
> but I think now I'm really really done with this :)

feel free to merge then :)

PR #462, opened 2021-07-19
Clean histories for successful and failed jobs separately

### LINKED ISSUE #322: Keep the Job Pods of Failed Jobs
## Summary

**As** user of K8up
**I want** to have failed K8up jobs available
**So that** I can investigate why a K8up job failed

## Context

The K8up backup operator will remove jobs (and their associated pods) according to a defined retention. It does that no matter the actual state of the job. So it cleans up for successfully finished jobs as well as for failed ones.

It doesn't really make sense to keep the successful ones around, so they should be cleaned anyways. But the operator should keep the failed jobs/pods around for easier log analysis.

There may be customers with smaller envs that also want to keep the successful jobs, so it would make sense to make this behaviour configurable.

## Out of Scope

* Do not update the helm chart and commodore component to make use of the new fields.

## Further links

* https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#jobs-history-limits
* https://github.com/vshn/k8up/blob/03bd378e769913a9337afa5b642dfda7f4ab5225/executor/backup.go#L154

## Acceptance criteria

*Given* a K8up job
*When* the job fails
*Then* the job is kept for further investigation.

* Helm chart and commodore component have the version bumped.

## Implementation Ideas

Could be done similar to how Kubernetes CronJobs does it: `.spec.successfulJobsHistoryLimit` and `.spec.failedJobsHistoryLimit`.

Add two new fields to the CRD and keep the existing as deprecated. The new ones have precedence whyle the existing one is used as a fallback. The change in behaviour is that the existing keep jobs is used for both successfull and failed thus doubling the amound of kept ones.

bastjan:
## Summary

This PR allows cleaning histories for successful and failed jobs separately.

It introduces two new fields to the job CRDs `.spec.successfulJobsHistoryLimit` and `.spec.failedJobsHistoryLimit`. Those fields override the now deprecated `.spec.KeepJobs`.

It fixes a bug where most job cleanups would result in always cleaning up the last object, due to the misuse of a pointer in a for range loop:

https://github.com/vshn/k8up/blob/df7435b28e9292ef88c582bce9aa64bb69185168/executor/check.go#L115-L117

Fixes #322.

## Checklist

- [ ] Keep pull requests small so they can be easily reviewed. Sadly no 🙈
- [x] Update tests.


ccremer (commented):
This is a first review and I'm not through yet. Overall good, but I don't see much changes in the documentation. Could you double check whether the new variables and behaviour is updated in the docs?

api/v1alpha1/backup_types.go
```diff
@@ -95,6 +102,31 @@ func (b *Backup) GetResources() corev1.ResourceRequirements {
 	return b.Spec.Resources
 }
 
+// GetFailedJobsHistoryLimit returns failed jobs history limit
```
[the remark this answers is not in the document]
bastjan: Yes. Way better this way 👍

api/v1alpha1/history_limits_test.go
```diff
@@ -0,0 +1,71 @@
...
+				return &v1alpha1.Prune{Spec: v1alpha1.PruneSpec{SuccessfulJobsHistoryLimit: successful, FailedJobsHistoryLimit: failed, KeepJobs: deprecatedKeep}}
+			},
+		},
+		{
+			desc: "Restore",
+			createSpec: func(successful, failed, deprecatedKeep *int) limiter {
+				return &v1alpha1.Restore{Spec: v1alpha1.RestoreSpec{SuccessfulJobsHistoryLimit: successful, FailedJobsHistoryLimit: failed, KeepJobs: deprecatedKeep}}
+			},
+		},
+	}
```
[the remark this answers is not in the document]
bastjan: I don't think so is used differing values to ensure failed jobs and successful jobs correspond to the set values. I added a no fallback value test though.

executor/generic.go
```diff
@@ -137,6 +123,7 @@ func (g *generic) RegisterJobSucceededConditionCallback() {
 				k8upv1alpha1.ReasonSucceeded,
 				"the job '%v/%v' ended successfully",
 				event.Job.Namespace, event.Job.Name)
+			g.Log.Info("c", "condis", g.Obj.GetStatus().Conditions)
```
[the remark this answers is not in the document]
bastjan: That was a debug statement 🙈

executor/cleaner/cleaner_test.go
```diff
@@ -0,0 +1,113 @@
...
+func newMockClient(jobs *k8upv1a1.RestoreList) client.Client {
+	objs := make([]client.Object, len(jobs.Items))
+	for i := range jobs.Items {
+		objs[i] = &jobs.Items[i]
+	}
+
+	scheme := runtime.NewScheme()
+	k8upv1a1.AddToScheme(scheme)
+	return fake.NewClientBuilder().WithScheme(scheme).WithObjects(objs...).Build()
+}
```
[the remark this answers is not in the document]
bastjan: The cleaner is integration tested in the `CheckTestSuite`. 

The problem with integration tests are that they are usually harder to debug. e.g. There must be a correct environment around which can be complicated with editor integrations. And they are slower. I prefer unit tests as long as possible and create more high level integration tests.

bastjan:
> This is a first review and I'm not through yet. Overall good, but I don't see much changes in the documentation. Could you double check whether the new variables and behaviour is updated in the docs?

Thanks for the review 😊

The documentation is auto generated here: `docs/modules/ROOT/pages/references/api-reference.adoc`. Any other place needed? 😊

ccremer:
> The documentation is auto generated here: `docs/modules/ROOT/pages/references/api-reference.adoc`. Any other place needed? blush

Not everything is automated :)
Please do a local preview and go through all pages where documenting the new behaviour could make sense. I recall there's a config reference page where the env variables are documented. There definitely BACKUP_KEEPJOBS needs to be more thoroughly documented/deprecated. You might also find code examples and add the new variables there and then.

api/v1alpha1/history_limits_test.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+func TestHistoryLimits(t *testing.T) {
+
+	type limiter interface {
+		GetSuccessfulJobsHistoryLimit() *int
+		GetFailedJobsHistoryLimit() *int
+	}
+
+	testCases := map[string]struct {
```
corvus-ch, line 18: Could be simplified to `map[string]func`.

Also, I prefer to put test tables outside of test functions. I makes it more readable (subjective to me) and allows to reuse one test table in multiple test functions (yes that is actually a thing).

executor/cleaner/cleaner_test.go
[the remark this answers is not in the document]
ccremer: > correct environment around which can be complicated with editor integrations

We have automated this and thanks to plain testing.T any IDE should be able to integrate nicely. 
Did you look at the env test suite here: https://github.com/vshn/k8up/blob/master/controllers/envsuite_test.go ?
The debugging is now the same as when using the fake client (you get a plain k8s client object).

It just feels kind of wrong to start using the fake client here and become inconsistent with other tests that we have just because it feels "slow"...

cfg/config.go
```diff
@@ -84,19 +86,21 @@ var (
...
-		MetricsBindAddress:      ":8080",
-		PodFilter:               "backupPod=true",
-		EnableLeaderElection:    true,
+		MountPath:                        "/data",
+		BackupAnnotation:                 "k8up.syn.tools/backup",
+		BackupCommandAnnotation:          "k8up.syn.tools/backupcommand",
+		FileExtensionAnnotation:          "k8up.syn.tools/file-extension",
+		ServiceAccount:                   "pod-executor",
+		BackupCheckSchedule:              "0 0 * * 0",
+		GlobalKeepJobs:                   6,
```
corvus-ch, line 95: Why still use the deprecated value? My expectation was to have `GlobalKeepJobs` unset and have the others set instead.

cfg/config.go
```diff
@@ -156,3 +160,21 @@ func (c Configuration) GetGlobalDefaultResources() (res corev1.ResourceRequireme
 func GetGlobalRepository() string {
 	return fmt.Sprintf("s3:%s/%s", Config.GlobalS3Endpoint, Config.GlobalS3Bucket)
 }
+
+// GetGlobalFailedJobsHistoryLimit returns the global failed jobs history limit.
+// Returns global KeepJobs if unspecified.
+func (c Configuration) GetGlobalFailedJobsHistoryLimit() int {
+	if c.GlobalFailedJobsHistoryLimit != -1 {
```
corvus-ch, line 167: If the input is less that `-1` then we get a negative number.

```suggestion
	if c.GlobalFailedJobsHistoryLimit > -1 {
```

cfg/config.go
```diff
@@ -156,3 +160,21 @@ func (c Configuration) GetGlobalDefaultResources() (res corev1.ResourceRequireme
...
+		return c.GlobalFailedJobsHistoryLimit
+	}
+	return c.GlobalKeepJobs
+}
+
+// GetGlobalSuccessfulJobsHistoryLimit returns the global successful jobs history limit.
+// Returns global KeepJobs if unspecified.
+func (c Configuration) GetGlobalSuccessfulJobsHistoryLimit() int {
+	if c.GlobalSuccessfulJobsHistoryLimit != -1 {
```
corvus-ch, line 176:
```suggestion
	if c.GlobalSuccessfulJobsHistoryLimit > -1 {
```

cfg/config_test.go
```diff
@@ -52,6 +53,32 @@ func Test_Configuration_DefaultConfig(t *testing.T) {
 	assert.Contains(t, err.Error(), "operator namespace")
 }
 
+func Test_Configuration_GetGlobalFailedJobsHistoryLimit(t *testing.T) {
```
corvus-ch, line 56: Add test case for when the imput is less than `-1`.

job/status.go
```diff
@@ -91,3 +91,34 @@ func (c *Config) SetFinished(namespace, name string) {
 		c.Log.Error(err, "could not patch status")
 	}
 }
+
+// GroupJobsByStatus groups jobs by the running state
+func GroupJobsByStatus(jobs []k8upv1alpha1.JobObject) (running []k8upv1alpha1.JobObject, failed []k8upv1alpha1.JobObject, successful []k8upv1alpha1.JobObject) {
```
corvus-ch, line 96: Try to avoid duplicate names. On that caller side this would read `job.GroupJobsByStatus`. I would prefer `job.GroupByStatus`. Less words and more on the natural language side.

corvus-ch (changes requested):

api/v1alpha1/schedule_types.go
```diff
@@ -123,6 +133,23 @@ func (s *Schedule) GetResources() corev1.ResourceRequirements {
...
+// Returns KeepJobs if unspecified.
+func (s *Schedule) GetFailedJobsHistoryLimit() *int {
+	if s.Spec.FailedJobsHistoryLimit != nil {
+		return s.Spec.FailedJobsHistoryLimit
+	}
+	return s.Spec.KeepJobs
+}
+
+// GetSuccessfulJobsHistoryLimit returns successful jobs history limit
+func (s *Schedule) GetSuccessfulJobsHistoryLimit() *int {
```
ccremer, line 146: Also here. Probably best to use the "Find usages" function of your IDE.

job/status.go
```diff
@@ -91,3 +91,34 @@ func (c *Config) SetFinished(namespace, name string) {
...
+}
+
+func isJobFinished(job k8upv1alpha1.JobObject) (bool, k8upv1alpha1.ConditionReason) {
+	for _, c := range job.GetStatus().Conditions {
+		if (k8upv1alpha1.ConditionType(c.Type) == k8upv1alpha1.ConditionCompleted) && c.Status == metav1.ConditionTrue {
+			return true, k8upv1alpha1.ConditionReason(c.Reason)
+		}
+	}
+	return false, ""
+}
```
ccremer, line 124: You could attach this helper function to the Status object maybe?
Like `job.GetStatus().IsFinished()`. Would make this helper accessible everywhere.

I think there are already some helpers if I'm not mistaken...?

api/v1alpha1/backup_types.go
```diff
@@ -12,8 +12,15 @@ import (
 type BackupSpec struct {
 	RunnableSpec `json:",inline"`
 
-	// KeepJobs amount of jobs to keep for later analysis
+	// DEPRECATED KeepJobs amount of jobs to keep for later analysis. Use FailedJobsHistoryLimit and SuccessfulJobsHistoryLimit respectively.
+	// +optional
 	KeepJobs *int `json:"keepJobs,omitempty"`
```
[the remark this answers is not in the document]
bastjan: good one 👍 didn't know that

api/v1alpha1/archive_types.go
```diff
@@ -76,6 +76,32 @@ func (a *Archive) GetResources() corev1.ResourceRequirements {
...
+// Returns KeepJobs if unspecified.
+func (a *Archive) GetFailedJobsHistoryLimit() *int {
+	if a.Spec.FailedJobsHistoryLimit != nil {
+		return a.Spec.FailedJobsHistoryLimit
+	}
+	return a.Spec.KeepJobs
+}
+
+// GetSuccessfulJobsHistoryLimit returns successful jobs history limit
+func (a *Archive) GetSuccessfulJobsHistoryLimit() *int {
```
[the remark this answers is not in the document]
bastjan: seems like my regex skills are bad :(

cfg/config.go
bastjan: Limited to zero and added a test 👍

api/v1alpha1/object.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// +k8s:deepcopy-gen=false
+
+// JobObject is an interface that must be implemented by all CRDs that implement a
+// job.
```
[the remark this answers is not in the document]
bastjan: fully agree 😊

api/v1alpha1/archive_types.go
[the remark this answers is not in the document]
bastjan: sorry about that. got a coffee and fixed it 😅

job/status.go
bastjan: Thanks for the tip, found some other issues with my status implementation too

Kidswiss:
@bastjan 

Sorry to barge in like that. But this caught my eye:

> It fixes a bug where most job cleanups would result in always cleaning up the last object, due to the misuse of a pointer in a for range loop:

It would fix this case, too: https://github.com/vshn/k8up/issues/465

Just opened that one a few hours ago :D

bastjan:
> @bastjan
>
> It would fix this case, too: #465, right?
> 
> Just opened that one a few hours ago :D

I can't yet prove it, but I do think so yes 😊

ccremer:
I ran the integration tests locally and get some error output here. It doesn't fail the tests (exit code 0) but I'm not sure if that's something as expected:
```console
=== CONT  Test_Check/TestJobCleanup
    check_integration_test.go:105: 2 deleted Checks found
=== RUN   Test_Check/TestReconciliation
    envsuite_test.go:147: creating namespace 'x7vs977v'
    envsuite_test.go:154: creating resource 'x7vs977v/check-integration-test0'
=== CONT  Test_Check
    logger.go:130: 2021-07-26T18:53:20.936+0200	DEBUG	adding job to the queue	{"check": "x7vs977v/check-integration-test0"}
    logger.go:130: 2021-07-26T18:53:20.948+0200	DEBUG	adding job to the queue	{"check": "x7vs977v/check-integration-test0"}
=== CONT  Test_Check/TestReconciliation
    check_integration_test.go:169: 1 Jobs found
=== CONT  Test_Check
    logger.go:130: 2021-07-26T18:53:22.349+0200	ERROR	could not patch status conditions	{"check": "x7vs977v/check-integration-test0", "error": "Patch \"http://127.0.0.1:36099/apis/backup.appuio.ch/v1alpha1/namespaces/x7vs977v/checks/check-integration-test0/status\": dial tcp 127.0.0.1:36099: connect: connection refused"}
    logger.go:130: 2021-07-26T18:53:22.349+0200	ERROR	cannot create job	{"check": "x7vs977v/check-integration-test0", "repository": "s3:/", "error": "Post \"http://127.0.0.1:36099/apis/batch/v1/namespaces/x7vs977v/jobs\": dial tcp 127.0.0.1:36099: connect: connection refused"}
--- PASS: Test_Check (14.69s)
    --- PASS: Test_Check/TestJobCleanup (7.17s)
    --- PASS: Test_Check/TestReconciliation (1.03s)
```

ccremer:
Please have another look at the documentation

* https://github.com/vshn/k8up/blob/7b47bcd9cc4a81576f8b0f95f0b3f4594b457681/docs/modules/ROOT/pages/references/object-specifications.adoc: We make still use of the deprecated KeepJobs field. Probably should remove those and replace with newer example.
* https://github.com/vshn/k8up/blob/7b47bcd9cc4a81576f8b0f95f0b3f4594b457681/docs/modules/ROOT/pages/references/status.adoc: Please check if the status and conditions are up-to-date
* https://github.com/vshn/k8up/blob/improve-monitoring/config/samples/k8up_v1alpha1_schedule.yaml#L41: Update the sample

Basically, a `grep -r -i "keepjobs" .` and see where it needs updating

bastjan:
> Please have another look at the documentation
> 
> * https://github.com/vshn/k8up/blob/7b47bcd9cc4a81576f8b0f95f0b3f4594b457681/docs/modules/ROOT/pages/references/object-specifications.adoc: We make still use of the deprecated KeepJobs field. Probably should remove those and replace with newer example.
> * https://github.com/vshn/k8up/blob/7b47bcd9cc4a81576f8b0f95f0b3f4594b457681/docs/modules/ROOT/pages/references/status.adoc: Please check if the status and conditions are up-to-date
> * https://github.com/vshn/k8up/blob/improve-monitoring/config/samples/k8up_v1alpha1_schedule.yaml#L41: Update the sample
> 
> Basically, a `grep -r -i "keepjobs" .` and see where it needs updating

Thanks 😊 I'm still working on the documentation👍

bastjan:
> I ran the integration tests locally and get some error output here. It doesn't fail the tests (exit code 0) but I'm not sure if that's something as expected:

I have the same issues on master. Did not yet take a look at the root cause.

executor/generic.go
```diff
@@ -176,6 +162,28 @@ func (g *generic) listOldResources(namespace string, objList client.ObjectList)
...
+		return
+	}
+
+	cl := cleaner.ObjectCleaner{Client: g.Client, Limits: limits, Log: g.Logger()}
+	err = cl.CleanOldObjects(g.CTX, typ.GetJobObjects())
+	if err != nil {
+		g.SetConditionFalseWithMessage(k8upv1alpha1.ConditionScrubbed, k8upv1alpha1.ReasonDeletionFailed, "could not cleanup old resources: %v", err)
+		return
+	}
+	g.SetConditionTrueWithMessage(k8upv1alpha1.ConditionScrubbed, k8upv1alpha1.ReasonSucceeded, "Deleted resources")
```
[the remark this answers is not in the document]
bastjan: done 👍

cfg/config.go
```diff
@@ -42,6 +42,8 @@ type Configuration struct {
 	BackupCheckSchedule              string `koanf:"checkschedule"`
 	GlobalAccessKey                  string `koanf:"globalaccesskeyid"`
 	GlobalKeepJobs                   int    `koanf:"globalkeepjobs"`
+	GlobalFailedJobsHistoryLimit     int    `koanf:"globalfailedjobshistorylimit"`
+	GlobalSuccessfulJobsHistoryLimit int    `koanf:"globalsuccessfuljobshistorylimit"`
```
ccremer, line 46:
```suggestion
	GlobalFailedJobsHistoryLimit     int    `koanf:"global-failed-jobs-history-limit"`
	GlobalSuccessfulJobsHistoryLimit int    `koanf:"global-successful-jobs-history-limit"`
```

docs/modules/ROOT/pages/references/config-reference.adoc
```diff
@@ -24,6 +24,8 @@ You need to define `BACKUP_OPERATOR_NAMESPACE`, but everything else can be left
 `BACKUP_GLOBAL_CPU_LIMIT`:: set the CPU limit for scheduled jobs
 `BACKUP_GLOBAL_CPU_REQUEST`:: set the CPU request for scheduled jobs
 `BACKUP_GLOBALKEEPJOBS`:: set the number of old jobs to keep when cleaning up, applies to all job types
+`BACKUP_GLOBALFAILEDJOBSHISTORYLIMIT`:: set the number of old, failed jobs to keep when cleaning up, applies to all job types
+`BACKUP_GLOBALSUCCESSFULJOBSHISTORYLIMIT`:: set the number of old, successful jobs to keep when cleaning up, applies to all job types
```
ccremer, line 28:
```suggestion
`BACKUP_GLOBAL_FAILED_JOBS_HISTORY_LIMIT`:: set the number of old, failed jobs to keep when cleaning up, applies to all job types
`BACKUP_GLOBAL_SUCCESSFUL_JOBS_HISTORY_LIMIT`:: set the number of old, successful jobs to keep when cleaning up, applies to all job types
```

docs/modules/ROOT/pages/references/object-specifications.adoc
```diff
@@ -224,7 +226,9 @@ spec:
 
 * `statsURL`: will send a JSON webhook containing check information information to this endpoint.
 * `backend`: see <<Backend, backend>>
-* `keepJobs`: amount of jobs that should be left after cleanup, for example how many job/pod objects should be left after they finished. Defaults to 6. Only applicable when used within a <<Schedule, schedule>>.
+* `keepJobs`: amount of jobs that should be left after cleanup, for example how many job/pod objects should be left after they finished. Deprecated, use `failedJobsHistoryLimit` and `successfulJobsHistoryLimit` instead. Only applicable when used within a <<Schedule, schedule>>.
+* `failedJobsHistoryLimit`: amount of failed jobs that should be left after cleanup, for example how many job/pod objects should be left after they finished. Defaults to 3. Only applicable when used within a <<Schedule, schedule>>.
+* `successfulJobsHistoryLimit`: amount of successful jobs that should be left after cleanup, for example how many job/pod objects should be left after they finished. Defaults to 3. Only applicable when used within a <<Schedule, schedule>>.
```
ccremer, line 231:
```suggestion
* `keepJobs`: amount of jobs that should be left after cleanup, for example how many job/pod objects should be left after they finished. 
  Deprecated, use `failedJobsHistoryLimit` and `successfulJobsHistoryLimit` instead. 
  Only applicable when used within a <<Schedule, schedule>>.
* `failedJobsHistoryLimit`: amount of failed jobs that should be left after cleanup, for example how many job/pod objects should be left after they finished. 
  Defaults to 3. Only applicable when used within a <<Schedule, schedule>>. 
  This setting takes precedence over `keepJops`.
* `successfulJobsHistoryLimit`: amount of successful jobs that should be left after cleanup, for example how many job/pod objects should be left after they finished. 
  Defaults to 3. Only applicable when used within a <<Schedule, schedule>>. 
  This setting takes precedence over `keepJops`.
```

docs/modules/ROOT/pages/references/object-specifications.adoc
```diff
@@ -264,7 +268,9 @@ spec:
 
 * `retention`: see <<Retention, retention>>
 * `backend`: see <<Backend, backend>>
-* `keepJobs`: amount of jobs that should be left after cleanup, for example how many job/pod objects should be left after they finished. Defaults to 6. Only applicable when used within a <<Schedule, schedule>>.
+* `keepJobs`: amount of jobs that should be left after cleanup, for example how many job/pod objects should be left after they finished. Deprecated, use `failedJobsHistoryLimit` and `successfulJobsHistoryLimit` instead. Only applicable when used within a <<Schedule, schedule>>.
+* `failedJobsHistoryLimit`: amount of failed jobs that should be left after cleanup, for example how many job/pod objects should be left after they finished. Defaults to 3. Only applicable when used within a <<Schedule, schedule>>.
+* `successfulJobsHistoryLimit`: amount of successful jobs that should be left after cleanup, for example how many job/pod objects should be left after they finished. Defaults to 3. Only applicable when used within a <<Schedule, schedule>>.
```
ccremer, line 273: Previous suggestion and one-line-per-sentence applies also here

executor/cleaner/cleaner_test.go
```diff
@@ -3,37 +3,31 @@
...
-	assert.NoError(t, client.List(context.TODO(), afterClean))
-	runningJobs, failedJobs, successfulJobs := job.GroupByStatus(afterClean.GetJobObjects())
-	assert.Equal(t, 3, len(runningJobs))
-	assert.Equal(t, 1, len(failedJobs))
-	assert.Equal(t, 1, len(successfulJobs))
+func (ts *CleanerTestSuite) TestCleanup() {
+	ts.withJobs()
+	ts.runCleanup()
+	ts.checkIfJobsDeleted()
 }
```
ccremer: Nice! Clean and readable. Just a small suggestion
```suggestion
func (ts *CleanerTestSuite) TestCleanup() {
	ts.withJobs()
	ts.runCleanup()
	ts.assertJobsDeleted()
}
```

executor/cleaner/cleaner_test.go
```diff
@@ -95,9 +111,10 @@ func markJobSuccessful(job k8upv1a1.JobObject) {
...
-				Type:   k8upv1a1.ConditionCompleted.String(),
-				Status: metav1.ConditionTrue,
-				Reason: k8upv1a1.ReasonSucceeded.String(),
+				Type:               k8upv1a1.ConditionCompleted.String(),
+				Status:             metav1.ConditionTrue,
+				Reason:             k8upv1a1.ReasonSucceeded.String(),
+				LastTransitionTime: metav1.Now(),
 			},
 		},
 	})
```
ccremer: There's also a ts.SetCondition() helper, in case that is more useful

executor/cleaner/cleaner_test.go
bastjan: sounds better 👍

executor/cleaner/cleaner_test.go
bastjan: `ts.SetCondition()` would need an object already present in the test suite. I prefer the flow of first creating a list of objects and then saving them altogether to the suite.

ccremer (approved):

cfg/config_test.go
```diff
@@ -52,6 +53,46 @@ func Test_Configuration_DefaultConfig(t *testing.T) {
...
+	t.Run("GlobalKeepJobsIfNotSet", func(t *testing.T) {
+		c := NewDefaultConfig()
+		c.GlobalFailedJobsHistoryLimit = -1
+		c.GlobalKeepJobs = rand.Int()
```
corvus-ch, line 60: Be cautious with random values in tests. They tend to result in flappy tests. One failed a few passed and some fails and now good again. Use explicit test values that covers known edge cases (for example lower one out of bound, lower in bound, somewhere whithin the middle, upper in bound and upper out of bound).

There is fuzzy testing but that is a subject on its own and I would keep them in a testsuite on there own that is run separatley.

However, it should not be an issue here.

corvus-ch (approved):
Looking good now. I left some remarks. Those are general thoughts and I do not see a need for a change.

Issue #427, opened 2021-05-10
k8up_jobs_failed_counter should be decrease when Failed Job is deleted


ZuGao:
## Summary

**As** "VSHN Engineer"\
**I want** "Prometheus value of k8up_jobs_failed_counter to reflect the current state of failed jobs in a cluster"\
**So that** "I would not get notified once I fix the backup or start a new one"

## Context

We have our Prometheus alert  with expression `rate(k8up_jobs_failed_counter[1d]) > 0` that will always fire an alert in Icinga even when we clean up the job and start a new one. We want the [value](https://github.com/vshn/k8up/blob/f03063dd53975b599b984963992a690fed5daf13/observer/observer.go#L36) of `k8up_jobs_failed_counter` to be decreased when a job in _Failed_ state is deleted so that the result of **rate** function would go back to 0, changing the alert to green again.

## Further links

* The exact file to be updated - > https://github.com/vshn/k8up/blob/master/observer/observer.go

## Acceptance criteria

* When a failed job is deleted manually, decrease the counter

## Implementation Ideas

* You are already tracking [deleted](https://github.com/vshn/k8up/blob/f03063dd53975b599b984963992a690fed5daf13/observer/observer.go#L135) Jobs. Try to figure out, maybe via the Status field of _batchv1.Job_, if the Job being deleted was previously in Failed state.

## Alternative Solution

* Try to implement another metric that would provide the same logic as above and leave the existing metric `k8up_jobs_failed_counter` as is.

ccremer:
Note: Counters cannot be decreased by design: https://prometheus.io/docs/concepts/metric_types/#counter

ZuGao:
We have to find a solution to this problem as it cripples our alert system in Icinga. Indeed the counter object cannot decrease its value but it can be substituted with another object class if necessary. In case this solution is unacceptable, as I already wrote in the main paragraph, a new metric should be introduced to reflect the state of the cluster in real time so that when issues are fixed the metrics should reflect that swiftly.

ccremer:
I like the idea to use k8s native metrics!
e.g. Openshift Logging (OCP4) also brings some alerts that fire when jobs fail based on this metric.

Adding more labels is often a good idea.

> we can't configure the operator to keep 0 old jobs as that would reset the metrics on k8s

[...]

PR #463, opened 2021-07-23
Consistent job naming

bastjan:
## Summary

This PR:
- Prefixes the job with the backup type 
- Sets the type as the container name

Part of #427.

## Checklist

- [x] Update the documentation. (Looked through the docs, the generated batch/v1.Job name is referenced nowhere)
- [x] Update tests.

### DIFF
--- a/api/v1alpha1/job_name_test.go
+++ b/api/v1alpha1/job_name_test.go
@@ -1,0 +1,26 @@
+package v1alpha1_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/vshn/k8up/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestJobName(t *testing.T) {
+	subjects := map[string]v1alpha1.JobObject{
+		"archive-job-name":  &v1alpha1.Archive{ObjectMeta: metav1.ObjectMeta{Name: "job-name"}},
+		"backup-job-name":   &v1alpha1.Backup{ObjectMeta: metav1.ObjectMeta{Name: "job-name"}},
+		"check-job-name":    &v1alpha1.Check{ObjectMeta: metav1.ObjectMeta{Name: "job-name"}},
+		"prune-job-name":    &v1alpha1.Prune{ObjectMeta: metav1.ObjectMeta{Name: "job-name"}},
+		"restore-job-name":  &v1alpha1.Restore{ObjectMeta: metav1.ObjectMeta{Name: "job-name"}},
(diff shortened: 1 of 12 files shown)


job/job.go
```diff
@@ -65,7 +65,7 @@ func GenerateGenericJob(obj k8upv1alpha1.JobObject, config Config) (*batchv1.Job
 					RestartPolicy: corev1.RestartPolicyOnFailure,
 					Containers: []corev1.Container{
 						{
-							Name:      obj.GetMetaObject().GetName(),
+							Name:      string(obj.GetType()),
```
ccremer, line 68: Is there no String() method available?
```suggestion
							Name:      obj.GetType().String(),
```

api/v1alpha1/schedule_types.go
```diff
@@ -119,6 +119,11 @@ func (s *Schedule) GetMetaObject() metav1.Object {
 	return s
 }
 
+// GetJobName returns the name of the underlying batch/v1 job.
+func (s *Schedule) GetJobName() string {
+	return s.GetType().String() + "-" + s.Name
+}
```
ccremer, line 125: Godoc comment doesn't match here. There's no batch/v1 job

api/v1alpha1/job_name_test.go
```diff
@@ -0,0 +1,25 @@
...
+		&v1alpha1.Archive{ObjectMeta: metav1.ObjectMeta{Name: name}},
+		&v1alpha1.Backup{ObjectMeta: metav1.ObjectMeta{Name: name}},
+		&v1alpha1.Check{ObjectMeta: metav1.ObjectMeta{Name: name}},
+		&v1alpha1.Prune{ObjectMeta: metav1.ObjectMeta{Name: name}},
+		&v1alpha1.Restore{ObjectMeta: metav1.ObjectMeta{Name: name}},
+		&v1alpha1.Schedule{ObjectMeta: metav1.ObjectMeta{Name: name}},
+	}
+
+	for _, subject := range subjects {
+		assert.Equal(t, subject.GetType().String()+"-"+name, subject.GetJobName())
```
corvus-ch, line 23: I would prefer to not compute the expected value but have it hard coded. For that I would make `subjects` a dictionary containing both the expected job name and the subject (`map[string]v1alpha1.JobObject`). While at it, I also would use nested test functions. This makes it esier to identify the exact case when there is a test failure.

corvus-ch (approved):
Whith the comments by @ccremer addressed, this is LGTM.

job/job.go
bastjan: there is one 👍 looks like my editor thought otherwise, thanks 😊

ccremer (approved):

PR #491, opened 2021-08-31
Regression: Exclude goland/vscode from git

ccremer:
## Summary

This is probably a regression from a recent wrestic integration, so I'd not add it to changelog.

## Checklist

- [x] Update tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+# IDEs
+.idea
+.vscode
+
 # Binaries for programs and plugins
 *.exe
 *.dll


cimnine (approved):
TBH, I believe these are entries that belong into each developers global gitignore file. Because we do not want to add all the IDEs out there. Otherwise you'd have to add ignores for Eclipse, Sublime Text, vim Swap Files, emacs, etc.
The same goes for Windows' `Thumbs.db` and macOS' `.DS_Store`.

I've approved nonetheless, since this is not my call to make.

ccremer:
That's a very valid point. But that requires a separate discussion in the team. There are countless repositories which have IDEs in their gitignore.

PR #495, opened 2021-09-03
Unified Command Line Interface

### LINKED ISSUE #477: Unifi CLI of backup and restic modules of k8up
## Summary

**As** user of k8up\
**I want** a stringent usage of the tool\
**So that** I can reduce learning to a minimum.

## Context

#359 moved the wrestic code into the k8up repository, but did try to changes as little as possible of the architecture.
The idea of this issue is to unify the command line interface of the two modules, operator and restic.

Currently, the two modules have different concepts to deal with arguments and variables. While the _operator_ module is mostly relying on environment variables, the _restic_ module mostly relies on command line arguments.

One unified approach could be that for every argument, both are possible: CLI argument and environment variable.

## Out of Scope

* Changes to the functionality.

## Further links

* https://github.com/vshn/k8up/issues/359

## Acceptance criteria

* Given `k8up help`, when looking through the available commands and arguments, the two modules seem to have been developed with the same usage concepts in mind.

## Implementation Ideas

The [cli library](https://github.com/urfave/cli) in use already supports environment variables for CLI arguments. This could be leveraged to unifi the command line interface of the two modules.

If this change is not backwards compatible, make sure to adjust the relevant documentation, code parts, configuration examples, default configuration values, etc.

cimnine:
## Summary

Both modules (_restic_ and _operator_) are now configured via `urfave/cli`'s Flags. All the flags' values can be still be defined via their respective ENV variables.

The advantage is that both modules build upon the same mechanism to configuration. Also `urface/cli` does some basic parsing and it has built-in support for required arguments / ENV vars.
Finally, it reduces one dependency.

Closes #477

## Checklist

- [x] Update tests.


restic/cfg/config.go
```diff
@@ -0,0 +1,91 @@
...
+	RestoreFilter      string
+	VerifyRestore      bool
+	RestoreTrimPath    bool
+
+	PruneKeepLast    *int
+	PruneKeepHourly  *int
+	PruneKeepDaily   *int
+	PruneKeepWeekly  *int
+	PruneKeepMonthly *int
+	PruneKeepYearly  *int
```
ccremer, line 59: Q: Does urfave/cli correctly set these values for pointers?

restic/cfg/config.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Validate ensures a consistent configuration and returns an error should that not be the case
+func (c *Configuration) Validate() error {
+	if c.DoRestore {
+		c.RestoreType = strings.ToLower(c.RestoreType)
+		switch c.RestoreType {
+		case RestoreTypeS3:
+			switch {
+			case "" == c.RestoreS3Endpoint:
+				return fmt.Errorf("if the restore type is set to '%s', then the restore s3 endpoint must be defined", RestoreTypeS3)
```
ccremer, line 73: This kinda reads strange, why not
```suggestion
			case c.RestoreS3Endpoint == "":
				return fmt.Errorf("if the restore type is set to '%s', then the restore s3 endpoint must be defined", RestoreTypeS3)
```

ccremer (commented):
:tada:

cmd/operator/main.go
```diff
@@ -27,10 +25,58 @@ import (
...
+			&cli.StringFlag{Destination: &cfg.Config.GlobalS3Endpoint, Name: "globals3endpoint", EnvVars: []string{"BACKUP_GLOBALS3ENDPOINT"}, Usage: "set the S3 endpoint to be used globally"},
+
+			&cli.StringFlag{Destination: &cfg.Config.GlobalCPUResourceRequest, Name: "global-cpu-request", EnvVars: []string{"BACKUP_GLOBAL_CPU_REQUEST"}, Usage: "set the CPU request for scheduled jobs"},
+			&cli.StringFlag{Destination: &cfg.Config.GlobalCPUResourceLimit, Name: "global-cpu-limit", EnvVars: []string{"BACKUP_GLOBAL_CPU_LIMIT"}, Usage: "set the CPU limit for scheduled jobs"},
+			&cli.StringFlag{Destination: &cfg.Config.GlobalMemoryResourceRequest, Name: "global-memory-request", EnvVars: []string{"BACKUP_GLOBAL_MEMORY_REQUEST"}, Usage: "set the memory request for scheduled jobs"},
+			&cli.StringFlag{Destination: &cfg.Config.GlobalMemoryResourceLimit, Name: "global-memory-limit", EnvVars: []string{"BACKUP_GLOBAL_MEMORY_LIMIT"}, Usage: "set the memory limit for scheduled jobs"},
+
+			&cli.StringFlag{Destination: &cfg.Config.BackupImage, Name: "image", EnvVars: []string{"BACKUP_IMAGE"}, Value: "quay.io/vshn/wrestic:latest", Usage: "URL of the restic image"},
```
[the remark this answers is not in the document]
cimnine: I try to focus on one thing at a time (and I regularly fail at it still). What you ask for will be handled in https://github.com/vshn/k8up/issues/476.

restic/cfg/config.go
cimnine: It comes from certain languages that require to call `equals` for String comparison, e.g. Java. There you want to do `"".equals(otherString)`. The reason is that if `otherString` is `null`, then `"".equals(otherString)` still works, whereas `otherString.equals("")` does not. This habbit sticks with me.

There are sometimes other benefits. E.g. here all the `""` align vertically, whereas they wouldn't if I'd write them the other way round. So it's easier to see that all three checks work similarly.

restic/cfg/config.go
cimnine: You raise a good point here and I will have to check whether these work as I expect.

cimnine: You were right. The values were not handled as I expected. I've updated the code to handle them.

docs/docs.mk
```diff
@@ -18,9 +18,19 @@ hunspell_cmd ?= $(docker_cmd) run $(docker_opts) --volume "$${PWD}":/spell vshn/
...
+docs-all: docs-update-usage docs-html docs-pdf ## Generate HTML and PDF docs
+
+.PHONY: docs-documents
+docs-documents: docs-update-usage docs-pdf docs-manpage docs-kindle docs-epub ## Generate downloadable docs
+
+.PHONY: docs-update-usage
+docs-update-usage: ## Generates dumps from `k8up --help`, which are then included as part of the docs
+	go run $(K8UP_MAIN_GO) --help > "$(docs_usage_dir)/k8up.txt"
+	go run $(K8UP_MAIN_GO) restic --help > "$(docs_usage_dir)/restic.txt"
+	go run $(K8UP_MAIN_GO) operator --help > "$(docs_usage_dir)/operator.txt"
```
ccremer, line 33: It probably makes sense to run this also in GitHub Actions, to ensure that the documentation is always up-to-date. Similar to `lint` target, we could check for changes with `git diff --exit-code`.

operator/cfg/config.go
```diff
@@ -32,98 +32,56 @@ const (
...
-	PromURL                          string `koanf:"promurl"`
-	RestartPolicy                    string `koanf:"restartpolicy"`
+	MountPath                        string
+	BackupAnnotation                 string
+	BackupCommandAnnotation          string
+	FileExtensionAnnotation          string
+	ServiceAccount                   string
+	BackupCheckSchedule              string
+	GlobalAccessKey                  string
+	GlobalKeepJobs                   *int
```
ccremer, line 50: Did you miss this one regarding pointers?

cimnine: 🤦

docs/docs.mk
cimnine: 👍  good idea

ccremer (approved):

bastjan (approved):
🚀

Issue #480, opened 2021-08-16
runAsUser, runAsGroup, fsGroup


cimnine:
## Summary

**As** an ops engineer setting up k8up\
**I want** control over the UID and GID with which (w)restic is started\
**So that** I can ensure tight security.

## Context

Currently restic is started as a specified user. This is either 1001 (65532 after #474) or root (maybe not anymore after #474), because these UIDs are burnt into the Docker image in the `Dockerfile`.

This could lead to problems with the mounts. I.e. if a Restore is performed, it will create the files with the permissions of the user it runs. When the container, who's files were restored, is restarted, it will see files with the UID 1001 (or 65532 or 0 respectively), which may prevent it's process to modify the files (edit, delete, maybe even read).

## Out of Scope

* Changes to other features

## Further links

* https://kubernetes.io/blog/2020/12/14/kubernetes-release-1.20-fsgroupchangepolicy-fsgrouppolicy/
* https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

## Acceptance criteria

* It is possible to set the UID/GID with which the restore Pod is run.
* The GID of the mount `/restore` matches the GID with which the restore Pod is run.

## Implementation Ideas

* A global flag (e.g. env var, CLI flag, both) to define the default UID and GID for restore Pods, plus
* a flag in the Restore CRD to define the UID and GID under which the restore Pod is created.

ccremer:
This is kinda a duplicate of #263 as they both aim to solve the same problem.
The current workaround is that there are image variants with root and non-root users.

However, since #497 removes the root user variants, we need an alternative to keep feature parity.
I'd suggest going forward with #263 instead of the idea in this issue for better flexibility (sometimes customizing UID and GID are not enough and I'd rather move this config to the CRD than CLI/ENV flags)

cimnine:
Closed in favor of #263

Issue #263, opened 2021-01-06
Ability to set SecurityContext for the various job types


Kidswiss:
## Summary

**As** a K8up user
**I want** to specify what user is used for the wrestic pods
**So that** it's able to backup all files

## Context

Wrestic 2.0 runs as root on purpose, there are some cases where PVCs can't be read otherwise.

But there also K8s distributions that do not want you to run the pods as root, like OpenShift. So we need a way to specify the `runAs` for the jobs so we can cover all the use cases.

Bonus: it would be great if the operator can determine this by itself. E.g. if the operator is running on OpenShift it will set a user and if running on another distro it will run as root.

## Out of Scope

* TBD

## Further links

* https://github.com/vshn/wrestic/pull/38
* https://github.com/vshn/wrestic/pull/19

## Acceptance criteria

*Given* a k8s cluster
*When* job is triggered
*Then* it will run with the right user

## Implementation Ideas

* none yet

ccremer:
`runAs` can often be insufficient when trying to access files from PVCs. `runAsGroup` or `fsGroup` are also relevant.

I propose to rather to set the [SecurityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) for either Pod or a container instead of just a single property

PR #497, opened 2021-09-10
Replace vshn/wrestic with unified vshn/k8up Docker image

cimnine:
## Summary

Replaces `vshn/wrestic` with the unified image `vshn/k8up`.

Closes https://github.com/vshn/k8up/issues/476

## Checklist

- [x] Update tests.


Dockerfile
```diff
@@ -14,24 +14,9 @@ RUN apk add --update --no-cache \
...
-## wrestic compatibility shim
-FROM k8up as wrestic
-
-WORKDIR /app
-ENTRYPOINT ["./wrestic"]
-
-## wrestic root compatibility shim
-FROM wrestic as wrestic-root
-
-USER 0
```
ccremer, line 34: This PR basically removes the possibility to have the image run as root, which is required in some scenarios: https://github.com/vshn/wrestic/pull/19

We should consider at least allowing to set the security context as an alternative, see https://github.com/vshn/k8up/issues/263

ccremer (changes requested):

Dockerfile
cimnine: I guess https://github.com/vshn/k8up/issues/480 and #263 are the same ^_^

TBH, I don't like to bring back the 'root' Docker image, as it's a work-around for not being able to influence the security context. But it is definitely the "quicker" way.

I suggest to give an implementation of #263 / #480 a time-boxed shot. What do you think @ccremer @tobru ?

ccremer: Yeah, #480 tries to solve the same problem, and yeah, I don't like the root image variants either, it was a workaround.
I prefer #263 though, as it gives more flexibility for the end users. Since we are aiming towards a breaking change, I see this as the only real and intended usage for backing up with root permissions.

cimnine: #480 would be an improvement and guarantee feature parity, as (previously) the restic image could also only be configured on operator level, if I'm not mistaken. Also I believe #480 would be the smaller effort to implement. Plus, once it is implemented, it also helps #263.

ccremer: I'm not sure, it looks about to be the same effort. What speaks against adding a field for the CRDs and take that security context "verbatim" to the generated jobs?

cimnine: I think you make a good argument there, Chris. I haven't thought about that solution, which might make this easier than I feared.

tobru: I'd say let's go for #263, I like that approach much more :+1:

cimnine: With #499 merged (and therefore #263 resolved) declare this discussion as resolved.

ccremer (approved):

PR #499, opened 2021-09-14
Add podSecurityContext to Schedule and schedulable actions

### LINKED ISSUE #263: Ability to set SecurityContext for the various job types

### LINKED ISSUE #480: runAsUser, runAsGroup, fsGroup

cimnine:
## Summary

This PR aims to implement #263. A security context is added to the `Schedule` CRD, as well as to the definitions of all the schedulable actions.

Closes #263 .

## Checklist

- [x] Update tests.


api/v1alpha1/job_object.go
```diff
@@ -17,7 +17,10 @@ type JobObject interface {
 	GetType() JobType
 	// GetJobName returns the name of the underlying batch/v1 job.
 	GetJobName() string
+	// GetResources returns the specified resource requirements
 	GetResources() corev1.ResourceRequirements
+	// GetSecurityContext returns the specified security context
+	GetSecurityContext() *corev1.PodSecurityContext
```
ccremer, line 23:
```suggestion
	// GetSecurityContext returns the specified pod security context
	GetSecurityContext() *corev1.PodSecurityContext
```

api/v1alpha1/check_types.go
```diff
@@ -90,6 +90,11 @@ func (c *Check) GetResources() corev1.ResourceRequirements {
 	return c.Spec.Resources
 }
 
+// GetSecurityContext returns the security context
+func (c *Check) GetSecurityContext() *corev1.PodSecurityContext {
```
ccremer, line 94: ...and all the other types that I'm too lazy to mark ;)
```suggestion
// GetSecurityContext returns the pod security context
func (c *Check) GetSecurityContext() *corev1.PodSecurityContext {
```

cimnine:
> Do you guys know if pod security context is enough to be able to do backups as root? Or should it rather be container security context?

```yaml
kind: Backup
metadata:
  name: demo-backup
spec:
  backend:
    ...
  podSecurityContext:
    runAsNonRoot: false
    runAsGroup: 0
    runAsUser: 0
```

The command runs as root on my testcluster:

```
2021-09-16T08:12:29.095Z	INFO	k8up	Starting k8up…	{"version": "snapshot", "date": "unknown", "commit": "unknown", "go_os": "linux", "go_arch": "amd64", "go_version": "go1.17", "uid": 0, "gid": 0}
```

The command does not run as root if `podSecurityContext.runAsNonRoot: true` or if no `podSecurityContext` is defined:

```
2021-09-16T08:45:56.903Z	INFO	k8up	Starting k8up…	{"version": "snapshot", "date": "unknown", "commit": "unknown", "go_os": "linux", "go_arch": "amd64", "go_version": "go1.17", "uid": 65532, "gid": 0}
```

Kidswiss (approved):
I believe that should be enough to backup the files as root.

Maybe it's possible to simulate that in some automated test? So we'd be sure if it works or not.

cimnine:
> Maybe it's possible to simulate that in some automated test? So we'd be sure if it works or not.

The security context is defined in the Backup and the Restore CRD used for the e2e test. Without it, the backup would still work, but the restore fail, as the directory permissions would be wrong and restic would not be able to write to the directory.

Issue #513, opened 2021-09-17
Automate Documentation Versioning


tobru:
## Summary

**As** K8up Maintainer
**I want** automated documentation versioning
**So that** on K8up release the documentation is also released.

## Context

Documentation versioning is independent of the K8up code. The version of the documentation is specified in `docs/antora.yml`.
For each Major and Minor version a `docs/vMajor.Minor` branch has to be created and the `docs/antora.yml` adapted accordingly.
When a Patch release is made, the corresponding documentation branch needs to be updated accordingly (no new branch needs to be created for Patch releases).

## Out of Scope

* Antora CI Pipeline overhaul

## Further links

* https://docs.antora.org/antora/2.3/content-source-versioning-methods/
* https://git.vshn.net/vshn/k8up_documentation/

## Acceptance criteria

* Given a new Major or Minor version release of K8up, when the release tag is pushed, a docs branch gets created and the `docs/antora.yml` file patched accordingly
* Given a new Patch version release of K8up, when the release tag is pushed, the docs branch gets updated accordingly.

## Implementation Ideas

* GitHub Action doing it automatically

cimnine:
It might also make sense to document how the documentation system now works and what conventions (like branch naming) it follows/expects.

PR #517, opened 2021-09-17
Automate the release of documentation for Antora

ccremer:
## Summary

This should now build documentation for tags using a "release" branch just for Antora

* Builds documentation with pushes to `master`. Master branch is considered Antora "prerelease"
  * Includes any branches matching "docs/v*" as releases.
  * Includes branches matching "docs/v*-rc* as prereleases.

* Builds documentation with `tag` pushes.
  * Creates a new branch with Major and Minor pattern, e.g. `v1.2.3` will create branch `docs/v1.2`
  * Patch versions will cause the branch to be rebased onto the tag
  * Prereleases will get their own branch per minor version, e.g. `v1.2.3-rc1` will create branch `docs/v1.2-rc`
  * All Prereleases belonging to the same minor version will cause the branch to be rebased onto latest matching rc tag.

## Checklist

- [x] Update tests.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,58 @@
+name: Docs
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - "*"
+
+jobs:
+  antora:
+    runs-on: ubuntu-latest
+    if: ${{ contains(github.ref, 'tags') }}
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Configure Git
+      run: |
+        git config user.name "Antora via GitHub Actions"
+        git config user.email "actions@github.com"
+    - name: Parse semver string
+      id: semver
+      uses: booxmedialtd/ws-action-parse-semver@v1
+      with:
+        input_string: ${{ github.ref }}
+        version_extractor_regex: '\/v(.*)$'
+    - name: Set variables
+      run: |
+        echo "MINOR_VERSION=${{ steps.semver.outputs.major }}.${{ steps.semver.outputs.minor }}" >> $GITHUB_ENV
+        echo "BRANCH_NAME=docs/v${{ steps.semver.outputs.major }}.${{ steps.semver.outputs.minor }}" >> $GITHUB_ENV
+    - name: Set branch name for Prerelease
+      if: ${{ steps.semver.outputs.prerelease != '' }}
+      # Cover all prereleases in one single rc branch. Administrators should manually delete the docs/vX.Y-rc branches from time to time for cleanup rc releases.
+      run: echo "BRANCH_NAME=docs/v${{ env.MINOR_VERSION}}-rc" >> $GITHUB_ENV
+
+    - name: Checkout remote branch if exists
+      run: git checkout ${{ env.BRANCH_NAME }}
+      continue-on-error: true
+    - name: Rebase if possible
+      run: git rebase ${GITHUB_REF##*/} ${{ env.BRANCH_NAME }}
+      continue-on-error: true
+    - name: Create new branch if not existing
+      run: git switch --create ${{ env.BRANCH_NAME }}
+      continue-on-error: true
+
+    - name: Patch Antora file for Release
+      run: yq eval 'del(.prerelease) | del (.display_version) | .version = "${{ env.MINOR_VERSION }}"' -i docs/antora.yml
+      if: ${{ steps.semver.outputs.prerelease == '' }}
+    - name: Patch Antora file for Prerelease
+      run: yq eval 'del (.display_version) | .version = "${{ env.MINOR_VERSION }}", .prerelease = "-${{ steps.semver.outputs.prerelease }}"' -i docs/antora.yml
+      if: ${{ steps.semver.outputs.prerelease != '' }}
+
+    - name: Commit
+      run: git commit -all --message "Update version for Antora" --author "Antora via GitHub Actions <actions@github.com>"
+      continue-on-error: true
+    - name: Push
+      run: git push --atomic --force --set-upstream origin ${{ env.BRANCH_NAME }}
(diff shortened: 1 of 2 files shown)


ccremer:
So I did continue this in my fork: https://github.com/ccremer/k8up/pull/10
Later finished in greposync: https://github.com/ccremer/greposync/pull/86

Lemme know if this goes into the right direction

tobru:
I think that's a good direction and seems to do what was asked in #513 :+1: 

What I discovered is that https://github.com/k8up-io/k8up/blob/master/docs/modules/ROOT/nav.adoc?plain=1#L2 doesn't work anymore. So I opted to add a changelog link manually for the time being in the doc release branches. If you have an idea how to automate that as well, it would be cool. Otherwise we can do that in another issue.

ccremer:
tbh I would just link to  https://github.com/k8up-io/k8up/releases without direct link magic.

tobru:
> tbh I would just link to https://github.com/k8up-io/k8up/releases without direct link magic.

Fine be me as well :+1:

ccremer:
@tobru This PR currently creates branch with `docs/v1.2` and not `docs/v1.2.x`. Is this somehow a problem somewhere else?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,58 @@
...
+      with:
+        input_string: ${{ github.ref }}
+        version_extractor_regex: '\/v(.*)$'
+    - name: Set variables
+      run: |
+        echo "MINOR_VERSION=${{ steps.semver.outputs.major }}.${{ steps.semver.outputs.minor }}" >> $GITHUB_ENV
+        echo "BRANCH_NAME=docs/v${{ steps.semver.outputs.major }}.${{ steps.semver.outputs.minor }}" >> $GITHUB_ENV
+    - name: Set branch name for Prerelease
+      if: ${{ steps.semver.outputs.prerelease != '' }}
+      # Cover all prereleases in one single rc branch. Administrators should manually delete the docs/vX.Y-rc branches from time to time for cleanup rc releases.
```
ccremer, line 34: I guess this could also be automated.
We could delete the `-rc` branch if the actual matching non-rc version has been released.
This could also be done in a later step after we've gathered some experience with it.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,58 @@
+name: Docs
+
+on:
+  push:
+    branches:
+      - master
```
ccremer, line 6: Actually we don't need to run this workflow in master, since the Antora builder runs outside of this repository.
Is there some kind of trigger that we need to execute on master? Or does it get updated regularly on a schedule without triggers?

ccremer (commented):
@tobru this might be for you to answer

.github/workflows/docs.yml
cimnine: Afaik Antora runs based on a webhook that is configured in the k8up repo.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,58 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ contains(github.ref, 'tags') }}
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Configure Git
+      run: |
+        git config user.name "GitHub Actions"
+        git config user.email "actions@github.com"
```
cimnine, line 21: Maybe this could be "Antora via GitHub Action" or alike? I would put at least some information here that makes the commits that result of this action relatable to this action.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,58 @@
...
+
+    - name: Patch Antora file for Release
+      run: yq eval 'del(.prerelease) | del (.display_version) | .version = "${{ env.MINOR_VERSION }}"' -i docs/antora.yml
+      if: ${{ steps.semver.outputs.prerelease == '' }}
+    - name: Patch Antora file for Prerelease
+      run: yq eval 'del (.display_version) | .version = "${{ env.MINOR_VERSION }}", .prerelease = "-${{ steps.semver.outputs.prerelease }}"' -i docs/antora.yml
+      if: ${{ steps.semver.outputs.prerelease != '' }}
+
+    - name: Commit
+      run: git commit -am "Update version for Antora" --author "GitHub Actions <actions@github.com>"
```
cimnine, line 55: I guess author should no longer be necessary here, since it was configured above?

cimnine: I further suggest to use long versions of the parameters:

```suggestion
      run: git commit --all --message "Update version for Antora"
```

.github/workflows/docs.yml
tobru: Good idea, I'd do it in a later step.

.github/workflows/docs.yml
tobru: It currently runs in a webhook. Could be improved by moving the CI trigger into a GitHub Action, like done for example here: https://github.com/appuio/appuio-io-docs/blob/master/.github/workflows/triggerci.yml. This gives better control.

But I don't think that it's urgent to change that now.

tobru:
> This PR currently creates branch with docs/v1.2 and not docs/v1.2.x. Is this somehow a problem somewhere else?

Nope, that's perfect :+1:

tobru (approved):

.github/workflows/docs.yml
ccremer: I'm not quite sure. In one of my tests it turned out that there was a different author than committer even with the git config from above... So I kept it explicitly

ccremer: I might try again in my fork or other private repo, but for now I'm fine with this

.github/workflows/docs.yml
tobru: Absolutely :+1:

.github/workflows/docs.yml
ccremer: actually this was pretty easy to implement, so I have this covered already.

PR #534, opened 2021-10-05
Only publish arch-agnostic Docker image tags

cimnine:
## Summary

Currently, goreleaser pushes intermediate images such as [`v2.0.0-rc3-arm64`](https://github.com/k8up-io/k8up/pkgs/container/k8up/8063600?tag=v2.0.0-rc3-arm64). Users should actually not need to be interested in the tag specific to the CPU architecture of the Docker image, but only in the agnostic tag, i.e. `v2.0.0-rc3`, and Docker will resolve the correct underlying images.

## Checklist

- [ ] Update tests.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -21,6 +21,7 @@ dockers:
   - "quay.io/k8up-io/k8up:v{{ .Version }}-amd64"
   goarch: amd64
   use: buildx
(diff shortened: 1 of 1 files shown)


ccremer (approved):
I'm curious, does this actually work? I always thought they should be pushed separately in addition...
But let's try for the next release, we can still revert it if it doesn't :)

PR #570, opened 2021-11-25
feat: added label `K8uplabel` to the `Job`

rparree:
## Summary

Currently the pods created by the Jobs only have the generated `job-name` label. For scheduled jobs this is based on the `Job` or the generated name of the `CronJob`. The latter makes the label indeterministic and hence unusable for references.

In our case we need to add  `NetworkPolicy` to add an egress to s3 and the k8s API server.

## Checklist

- [x ] Keep pull requests small so they can be easily reviewed.


ccremer:
uuh, this is interesting. The e2e test fails because Minio helm chart at https://github.com/minio/charts completely disappeared?

ccremer (approved):
@Kidswiss I do not see a blocker or issue with this change. Do you see any?

Kidswiss:
@ccremer LGTM

We used to have such labels in v1.x but it looks like they didn't survive the rewrite :)

ccremer:
> We used to have such labels in v1.x but it looks like they didn't survive the rewrite :)

I couldn't quite remember it, but in that case it's a regression bug

PR #640, opened 2022-04-12
Add kind-minio target to makefile

Kidswiss:
## Summary
This target will deploy a minio instance into the kind cluster. Prior to
this the minio instance had to be running seperately.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,11 +159,18 @@ kind-clean: ## Removes the kind instance if it exists.
 
 .PHONY: kind-run
 kind-run: export KUBECONFIG = $(KIND_KUBECONFIG)
-kind-run: kind-setup install run-operator ## Runs the operator on the local host but configured for the kind cluster
+kind-run: kind-setup kind-minio install run-operator ## Runs the operator on the local host but configured for the kind cluster
 
 kind-e2e-image: docker-build
 	$(e2e_make) kind-e2e-image
 
+.PHONY: kind-minio
+kind-minio: $(minio_sentinel)
+
+$(minio_sentinel): export KUBECONFIG = $(KIND_KUBECONFIG)
+$(minio_sentinel): kind-setup
+	kubectl apply -f $(SAMPLES_ROOT_DIR)/deployments/minio.yaml
+	@touch $@
 ###
 ### E2E Test
 ###
--- a/Makefile.vars.mk
+++ b/Makefile.vars.mk
@@ -16,6 +16,9 @@ CRD_FILE ?= k8up-crd.yaml
 CRD_ROOT_DIR ?= config/crd/apiextensions.k8s.io
 CRD_DOCS_REF_PATH ?= docs/modules/ROOT/pages/references/api-reference.adoc
 
+SAMPLES_ROOT_DIR ?= config/samples
+minio_sentinel = $(e2etest_dir)/minio_sentinel
+
 KIND_NODE_VERSION ?= v1.23.1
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -159,11 +159,17 @@ kind-clean: ## Removes the kind instance if it exists.
...
 
 kind-e2e-image: docker-build
 	$(e2e_make) kind-e2e-image
 
+.PHONY: kind-minio
+kind-minio: export KUBECONFIG = $(KIND_KUBECONFIG)
+kind-minio: kind-setup
+kind-minio:
+	kubectl apply -f $(SAMPLES_ROOT_DIR)/deployments/minio.yaml
+
```
ccremer, line 172: I think it's safe to assume that the minio only really needs to be applied once. I propose to introduce a "sentinel file" that prevents subsequent `make kind-run` from applying the same manifests over again, thus speeding up local dev.
```suggestion
.PHONY: kind-minio
kind-minio: $(minio_sentinel)

$(minio_sentinel): export KUBECONFIG = $(KIND_KUBECONFIG)
$(minio_sentinel): kind-setup
	kubectl apply -f $(SAMPLES_ROOT_DIR)/deployments/minio.yaml
	@touch $@
```

Makefile.vars.mk
```diff
@@ -16,6 +16,8 @@ CRD_FILE ?= k8up-crd.yaml
 CRD_ROOT_DIR ?= config/crd/apiextensions.k8s.io
 CRD_DOCS_REF_PATH ?= docs/modules/ROOT/pages/references/api-reference.adoc
 
+SAMPLES_ROOT_DIR ?= config/samples
+
```
ccremer, line 20:
```suggestion
SAMPLES_ROOT_DIR ?= config/samples
minio_sentinel = .kind/minio_sentinel

```

Makefile
Kidswiss: Yeah I agree, the minio instance will probably not really change when developing. Thanks for the suggestion.

Makefile.vars.mk
```diff
@@ -16,6 +16,9 @@ CRD_FILE ?= k8up-crd.yaml
 CRD_ROOT_DIR ?= config/crd/apiextensions.k8s.io
 CRD_DOCS_REF_PATH ?= docs/modules/ROOT/pages/references/api-reference.adoc
 
+SAMPLES_ROOT_DIR ?= config/samples
+minio_sentinel = $(e2etest_dir)/minio_sentinel
```
ccremer, line 20: are you sure that the e2etest_dir gets created in a new, cleaned environment, without calling `make e2e-test` first? aka is make actually able to create the file without the dir? I'm not sure about that.

Kidswiss: The `kind-setup` target uses that folder, too: https://github.com/k8up-io/k8up/blob/master/e2e/kind.mk#L1

The `.kind` folder doesn't exist, so your first suggestion wasn't working.

ccremer: ah, I mixed that up. I thought kind stuff would go into `.kind` and e2e stuff into `.e2e`. I must have remembered it wrongly then. All good

ccremer (approved):

Makefile.vars.mk
Kidswiss: yeah I was a bit confused, too. The makefiles baesd on `go-bootstrap` use `.kind`.

PR #657, opened 2022-05-09
Add 'envFrom' generic configuration for backend spec

sashokbg:
## Summary

Added envFrom configuration for backends
EnvFrom can be used to add arbitrary env variables 'in bulk' to the job container template.
This feature can be useful in implementing some generic authentication

Signed-off-by: Alexander KIRILOV <aleksandar.kirilov@proxym.fr>

## Checklist
- [x] Update tests.

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/generic-env-vars.adoc
+++ b/docs/modules/ROOT/pages/how-tos/generic-env-vars.adoc
@@ -1,0 +1,25 @@
+= Generic Restic Environment Variables
+
+Most of the supported backup backends allow you to specify authentication details as secret values.
+
+One such example is the S3 backend spec; it has `accessKeyIDSecretRef` and `secretAccessKeySecretRef` fields.
+
+It is however sometimes useful to pass additional environment variables to the container that runs the backup.
+This can be achieved with the help of the `envFrom` field that references a config map or a secret at the backend level.
+
+Please be aware that you could potentially add conflicting or duplicate environment variables.
+
+Check the api reference for the xref:references/api-reference.adoc#{anchor_prefix}-github-com-k8up-io-k8up-api-v1-backend[Backend].
+
+You can read link:https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables[the kubernetes documentation] for more information on how to use `envFrom`.
+
+== Open Stack Swift Authentication Example
+
+The following example shows you how to configure Open Stack Swift authentication using a Kubernetes secret and `envFrom` configuration for your backend.
+
+You can read more on link:https://wiki.openstack.org/wiki/OpenStackClient/Authentication[Open Stack Client Authentication].
+
+[source,yaml]
+----
+include::example$open-stack-swift-auth.yaml[]
+----
--- a/operator/executor/archive.go
+++ b/operator/executor/archive.go
@@ -61,6 +61,7 @@ func (a *ArchiveExecutor) startArchive(archiveJob *batchv1.Job, archive *k8upv1.
 	a.RegisterJobSucceededConditionCallback()
 
 	archiveJob.Spec.Template.Spec.Containers[0].Env = a.setupEnvVars(archive)
+	archive.Spec.AppendEnvFromToContainer(&archiveJob.Spec.Template.Spec.Containers[0])
 	archiveJob.Spec.Template.Spec.Containers[0].Args = a.setupArgs(archive)
 
 	err := a.Client.Create(a.CTX, archiveJob)
--- a/operator/executor/backup.go
+++ b/operator/executor/backup.go
@@ -123,6 +123,7 @@ func (b *BackupExecutor) startBackup(backupJob *batchv1.Job) error {
 	}
 
 	backupJob.Spec.Template.Spec.Containers[0].Env = b.setupEnvVars()
+	b.backup.Spec.AppendEnvFromToContainer(&backupJob.Spec.Template.Spec.Containers[0])
 	backupJob.Spec.Template.Spec.Volumes = volumes
 	backupJob.Spec.Template.Spec.ServiceAccountName = cfg.Config.ServiceAccount
 	backupJob.Spec.Template.Spec.Containers[0].VolumeMounts = b.newVolumeMounts(volumes)
--- a/operator/executor/prune.go
+++ b/operator/executor/prune.go
@@ -66,6 +66,7 @@ func (p *PruneExecutor) startPrune(pruneJob *batchv1.Job, prune *k8upv1.Prune) {
 	p.RegisterJobSucceededConditionCallback()
 
 	pruneJob.Spec.Template.Spec.Containers[0].Env = p.setupEnvVars(prune)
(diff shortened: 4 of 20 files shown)


Kidswiss:
Hi @sashokbg 

Thanks for your contribution! I'm having a look at this right now.

operator/executor/backup.go
```diff
@@ -123,6 +123,7 @@ func (b *BackupExecutor) startBackup(backupJob *batchv1.Job) error {
 	}
 
 	backupJob.Spec.Template.Spec.Containers[0].Env = b.setupEnvVars()
+	backupJob.Spec.Template.Spec.Containers[0].EnvFrom = b.backup.Spec.Backend.EnvFrom
```
Kidswiss, line 126: You probably want to add this to the other jobs as well, or only backups will be able to use those variables.

Kidswiss (changes requested):
Overall LGTM.

Just one minor thing, so that it works as intended.

sashokbg:
@Kidswiss I am really sorry for the question but since it is my first time writing Go code - how should I implement the null check ?
A simple will work ?

	if restore.Spec.Backend.EnvFrom != nil {
		j.Spec.Template.Spec.Containers[0].EnvFrom = restore.Spec.Backend.EnvFrom
	}

sashokbg:
@Kidswiss thank you for your feedback - I have updated the PR to reflect the requested changes. I have tested the integration tests locally and everything passes

docs/modules/ROOT/pages/how-tos/generic-authentication.adoc
```diff
@@ -0,0 +1,23 @@
...
+It is however sometimes useful to pass additional environment variables to the container that runs the backup. +
+This can be achieved with the help of the "EnvFrom" value that references a config map or a secret at the backend level.
+
+Check the api reference for the xref:references/api-reference.adoc#{anchor_prefix}-github-com-k8up-io-k8up-api-v1-backend[Backend]
+
+You can read link:https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables[the kubernetes documentation] for more information on how to use "envFrom"
+
+== Open Stack Swift Authentication Example
+
+The following example shows you how to configure Open Stack Swift authentication using a kubernetes secret and "EnvFrom" configuration for your backend.
```
ccremer, line 16:
```suggestion
The following example shows you how to configure Open Stack Swift authentication using a Kubernetes secret and `envFrom` configuration for your backend.
```

operator/executor/archive.go
```diff
@@ -61,6 +61,9 @@ func (a *ArchiveExecutor) startArchive(archiveJob *batchv1.Job, archive *k8upv1.
 	a.RegisterJobSucceededConditionCallback()
 
 	archiveJob.Spec.Template.Spec.Containers[0].Env = a.setupEnvVars(archive)
+	if archive.Spec.Backend.EnvFrom != nil {
+		archiveJob.Spec.Template.Spec.Containers[0].EnvFrom = archive.Spec.Backend.EnvFrom
+	}
```
ccremer, line 66:
```suggestion
	if archive.Spec.Backend != nil {
		archiveJob.Spec.Template.Spec.Containers[0].EnvFrom = archive.Spec.Backend.EnvFrom
	}
```

operator/executor/prune.go
```diff
@@ -66,6 +66,9 @@ func (p *PruneExecutor) startPrune(pruneJob *batchv1.Job, prune *k8upv1.Prune) {
 	p.RegisterJobSucceededConditionCallback()
 
 	pruneJob.Spec.Template.Spec.Containers[0].Env = p.setupEnvVars(prune)
+	if prune.Spec.Backend.EnvFrom != nil {
+		pruneJob.Spec.Template.Spec.Containers[0].EnvFrom = prune.Spec.Backend.EnvFrom
+	}
```
ccremer, line 71:
```suggestion
	if prune.Spec.Backend != nil {
		pruneJob.Spec.Template.Spec.Containers[0].EnvFrom = prune.Spec.Backend.EnvFrom
	}
```

sashokbg:
Hello @ccremer I will try to add a couple of unit tests to cover the null exception. Being a total novice in go this will probably take me a some time :)

docs/modules/ROOT/pages/how-tos/generic-authentication.adoc
```diff
@@ -0,0 +1,25 @@
+= Generic Restic Environment Variables
```
ccremer, line 1: Please rename the file so it matches the title somewhat, e.g. `generic-env-vars.adoc` (don't forget to update the nav link as well, see https://github.com/k8up-io/k8up/blob/master/docs/modules/ROOT/nav.adoc, I can't suggest edits in unchanged files)

ccremer (commented):
I had a look at it again and I figured we could optimize the code a bit.
Instead of doing a nil check in every occurrence, we add a function to the `spec` struct that accepts a container spec and adds the env vars. That way this should be easier to unit test and the logic is in a single place only.

https://github.com/k8up-io/k8up/blob/a92ff237eacaf47951f30cef499fd399a6355f7f/api/v1/runnable_types.go#L8
```go
func (in *RunnableSpec) AppendEnvFromToContainer(containerSpec *corev1.ContainerSpec) {
  if in.Backend != nil {
    containerSpec.EnvFrom = append([]corev1.EnvFromSource, in.Backend.EnvFrom...)
  }
}
```
then in each executor instead of the nil check call the function like this:
```go
  b.backup.Spec.AppendEnvFromToContainer(&backupJob.Spec.Template.Spec.Containers[0])
```

WDYT?

sashokbg:
Hello @ccremer thank you for your time and helpful feedback. I think what you proposed is a good idea and went ahead and implemented it as you specified. I have also included a unit test that tests the new function

ccremer (approved):

docs/modules/ROOT/pages/how-tos/generic-env-vars.adoc
```diff
@@ -0,0 +1,25 @@
+= Generic Restic Environment Variables
```
ccremer, line 1: This file still doesn't show up in the menu. You have to actively link the page. Please add it to `docs/modules/ROOT/nav.adoc` (see e.g. https://github.com/k8up-io/k8up/blob/master/docs/modules/ROOT/nav.adoc)

you can view a preview of the documentation by running `make docs-preview` on `http://localhost:2020`.

ccremer (changes requested):
Actually I've found something to be missing

docs/modules/ROOT/pages/how-tos/generic-env-vars.adoc
sashokbg: Ah true, I forgot it was an antora page :) I will add it

sashokbg: Done

ccremer (approved):
I think you need to amend your commits for the DCO checks, other than that it's LGTM

Kidswiss (approved):
I like the idea of moving it to it's own method.

LGTM

ccremer:
@sashokbg thank you for your contribution :tada:

Issue #638, opened 2022-04-08
K8up operator crashes in panic because of missing API implementation


ThomasSteinbach:
### Description

As for Backups I tried to set the `podSecurityContext` within the `backup` definition of a `Schedule`:
https://k8up.io/k8up/2.1/references/api-reference.html#k8s-api-github-com-k8up-io-k8up-api-v1-runnablespec

As described in the API documentation this should be valid, as the `Schedule` specification takes the same `backup` specification as the `Backup` itself.

However setting the `podSecurityContext` leads to a `panic` crash of the K8up operator pod (see Logs section).

### Logs

```shell
logs from the k8up operator pod:


 panic: reflect: call of reflect.Value.Type on zero Value                                                                                                            │
                                                                                                                                                                     │
 goroutine 470 [running]:                                                                                                                                            │
 reflect.Value.Type({0x0, 0x0, 0x16e1060})                                                                                                                           │
     /opt/hostedtoolcache/go/1.17.7/x64/src/reflect/value.go:2262 +0x12e                                                                                             │
 github.com/imdario/mergo.merge({0x16e1060, 0xc0004b4690}, {0x16e1060, 0x0}, {0x0, 0x0, 0x1782a6c})                                                                  │
     /home/runner/go/pkg/mod/github.com/imdario/mergo@v0.3.12/merge.go:363 +0x1c9                                                                                    │
 github.com/imdario/mergo.Merge(...)                                                                                                                                 │
     /home/runner/go/pkg/mod/github.com/imdario/mergo@v0.3.12/merge.go:296                                                                                           │
 github.com/k8up-io/k8up/operator/handler.(*ScheduleHandler).mergeSecurityContextWithDefaults(0xc0009879d8, 0xc00077ac00)                                            │
     /home/runner/work/k8up/k8up/operator/handler/schedule.go:139 +0xce                                                                                              │
 github.com/k8up-io/k8up/operator/handler.(*ScheduleHandler).mergeWithDefaults(0xc00077ab00, 0xc00077ab00)                                                           │
     /home/runner/work/k8up/k8up/operator/handler/schedule.go:108 +0x45                                                                                              │
 github.com/k8up-io/k8up/operator/handler.(*ScheduleHandler).createJobList(0xc0009879d8)                                                                             │
     /home/runner/work/k8up/k8up/operator/handler/schedule.go:93 +0x465                                                                                              │
[...]

PR #676, opened 2022-05-24
Fix panic when merging PodSecuritySontext from Schedule #638

ZuGao:
## Summary

Closes https://github.com/k8up-io/k8up/issues/638

## Checklist

- [x] Update tests.

### DIFF
--- a/operator/handler/schedule_test.go
+++ b/operator/handler/schedule_test.go
@@ -184,6 +184,42 @@ func TestScheduleHandler_mergeBackendWithDefaults(t *testing.T) {
 	}
 }
 
+func TestScheduleHandler_mergePodSecurityContextWithDefaults(t *testing.T) {
+	tests := map[string]struct {
+		givenSchedulePodSecurityContext *corev1.PodSecurityContext
+		givenResourcePodSecurityContext *corev1.PodSecurityContext
+		expectedPodSecurityContext      *corev1.PodSecurityContext
+	}{
+		"GivenScheduleContextIsNil_WhenResourceContextIsGiven_ThenUseResourcePSC": {
+			givenSchedulePodSecurityContext: nil,
+			givenResourcePodSecurityContext: newBuilderPSC().runAsUser(0).runAsGroup(1).seLinuxOptions("s0:c123,c456").get(),
+			expectedPodSecurityContext:      newBuilderPSC().runAsUser(0).runAsGroup(1).seLinuxOptions("s0:c123,c456").get(),
+		},
+		"GivenScheduleContextIsPopulated_WhenResourceContextIsGiven_ThenMergeWithResourcePSC": {
+			givenSchedulePodSecurityContext: newBuilderPSC().runAsGroup(2).seLinuxOptions("s1:c123,c456").get(),
+			givenResourcePodSecurityContext: newBuilderPSC().runAsUser(1).runAsGroup(1).get(),
+			expectedPodSecurityContext:      newBuilderPSC().runAsUser(1).runAsGroup(1).seLinuxOptions("s1:c123,c456").get(),
+		},
+		"GivenScheduleContextIsPopulated_WhenResourceContextIsNil_ThenUseResourcePSC": {
+			givenSchedulePodSecurityContext: newBuilderPSC().runAsGroup(1).seLinuxOptions("s1:c123,c456").get(),
+			givenResourcePodSecurityContext: nil,
+			expectedPodSecurityContext:      newBuilderPSC().runAsGroup(1).seLinuxOptions("s1:c123,c456").get(),
+		},
+	}
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			schedule := ScheduleHandler{schedule: &k8upv1.Schedule{Spec: k8upv1.ScheduleSpec{
+				PodSecurityContext: tt.givenSchedulePodSecurityContext,
+			}}}
+			res := &k8upv1.RunnableSpec{
+				PodSecurityContext: tt.givenResourcePodSecurityContext,
+			}
+			schedule.mergeSecurityContextWithDefaults(res)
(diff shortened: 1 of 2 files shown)


operator/handler/schedule_test.go
```diff
@@ -184,6 +184,33 @@ func TestScheduleHandler_mergeBackendWithDefaults(t *testing.T) {
...
 
+func TestScheduleHandler_mergePodSecurityContextWithDefaults(t *testing.T) {
+	var nilPodSecurityContextPointer *corev1.PodSecurityContext
+	tests := map[string]struct {
+		givenSchedulePodSecurityContext *corev1.PodSecurityContext
+		givenResourcePodSecurityContext *corev1.PodSecurityContext
+		expectedPodSecurityContext      *corev1.PodSecurityContext
+	}{
+		"Given_ResourcePSC_When_SpecPSC_Is_Nil_Then_UseResourcePSC": {
+			givenSchedulePodSecurityContext: nilPodSecurityContextPointer,
```
ccremer, line 195:
```suggestion
			givenSchedulePodSecurityContext: nil,
```

operator/handler/schedule_test.go
```diff
@@ -184,6 +184,33 @@ func TestScheduleHandler_mergeBackendWithDefaults(t *testing.T) {
 	}
 }
 
+func TestScheduleHandler_mergePodSecurityContextWithDefaults(t *testing.T) {
+	var nilPodSecurityContextPointer *corev1.PodSecurityContext
```
ccremer, line 188: why having `nil` as a variable?
```suggestion
```

operator/handler/schedule_test.go
```diff
@@ -184,6 +184,33 @@ func TestScheduleHandler_mergeBackendWithDefaults(t *testing.T) {
...
+	tests := map[string]struct {
+		givenSchedulePodSecurityContext *corev1.PodSecurityContext
+		givenResourcePodSecurityContext *corev1.PodSecurityContext
+		expectedPodSecurityContext      *corev1.PodSecurityContext
+	}{
+		"Given_ResourcePSC_When_SpecPSC_Is_Nil_Then_UseResourcePSC": {
+			givenSchedulePodSecurityContext: nilPodSecurityContextPointer,
+			givenResourcePodSecurityContext: newPodSecurityContext(0, 1, corev1.SELinuxOptions{Level: "s0:c123,c456"}),
+			expectedPodSecurityContext:      newPodSecurityContext(0, 1, corev1.SELinuxOptions{Level: "s0:c123,c456"}),
+		},
```
ccremer, line 198: If you've already prepare test cases, please add cases that verifies the existing behaviour

operator/handler/schedule_test.go
```diff
@@ -184,6 +184,33 @@ func TestScheduleHandler_mergeBackendWithDefaults(t *testing.T) {
...
 }
 
+func TestScheduleHandler_mergePodSecurityContextWithDefaults(t *testing.T) {
+	var nilPodSecurityContextPointer *corev1.PodSecurityContext
+	tests := map[string]struct {
+		givenSchedulePodSecurityContext *corev1.PodSecurityContext
+		givenResourcePodSecurityContext *corev1.PodSecurityContext
+		expectedPodSecurityContext      *corev1.PodSecurityContext
+	}{
+		"Given_ResourcePSC_When_SpecPSC_Is_Nil_Then_UseResourcePSC": {
```
ccremer, line 194:
```suggestion
		"GivenScheduleContextIsNil_WhenResourceContextIsGiven_ThenUseResourcePSC": {
```

operator/handler/schedule_test.go
```diff
@@ -184,6 +184,33 @@ func TestScheduleHandler_mergeBackendWithDefaults(t *testing.T) {
...
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			schedule := ScheduleHandler{schedule: &k8upv1.Schedule{Spec: k8upv1.ScheduleSpec{
+				PodSecurityContext: tt.givenSchedulePodSecurityContext,
+			}}}
+			res := &k8upv1.RunnableSpec{
+				PodSecurityContext: tt.givenResourcePodSecurityContext,
+			}
+			schedule.mergeSecurityContextWithDefaults(res)
+			assert.ObjectsAreEqual(*tt.expectedPodSecurityContext, *tt.givenResourcePodSecurityContext)
```
ccremer, line 209: that's wrong. There's no `t` passed through and the return value of `ObjectsAreEqual` is a boolean that you don't consume. This results in the test passing even if the objects aren't equal, meaning the test would be a worthless no-op ;)
```suggestion
			assert.Equal(t, *tt.expectedPodSecurityContext, *tt.givenResourcePodSecurityContext)
```

ccremer (changes requested):

operator/handler/schedule_test.go
ZuGao: Isn't this the value which triggers the panic error?

operator/handler/schedule_test.go
ZuGao: Hmm interesting, I expected that it would test the equality. Anyway that's ok, I will change it

operator/handler/schedule_test.go
ccremer: See my suggestion "above", just use `nil`, since `givenSchedulePodSecurityContext` is a pointer type `*corev1.PodSecurityContext`

operator/handler/schedule_test.go
ccremer: Well it does test the equality, but it doesn't fail the test because `t` is used to tell the test environment that something failed. And that isn't part of the method signature in `ObjectsAreEqual`

ccremer (approved):
one slight name suggestion, otherwise LGTM

Kidswiss (approved):
LGTM

Issue #690, opened 2022-06-16
Backups to B2 fail after the first backup


l0wl3vel:
### Description

When creating a schedule with the B2 backend instead of the S3 backend the first backup runs fine and ends sucessfully. But every subsequent backup fails with the attached logs.

### Additional Context

Backups via the S3 provider run fine all the time, but we would like to use the path parameter that is only offered by the B2 provider.

### Logs

```shell
Common labels: {"pod":"backup-jira-backup-vrkmz-ttlf8","stream":"stderr","app":"backup-jira-backup-vrkmz","container":"backup","job":"apps-jira/backup-jira-backup-vrkmz","namespace":"apps-jira","node_name":"hub-default-worker-78db654874-stvq4"}
Line limit: 1000
Total bytes processed: "5.29  kB"


2022-06-16T14:11:32+02:00	2022/06/16 12:11:32 unable to start k8up: failed to initialise the restic repository: cmd.Wait() err: 1
2022-06-16T14:11:32+02:00	1.6553814927560086e+09	INFO	k8up.restic.restic.RepoInit.restic.stderr	
2022-06-16T14:11:32+02:00	1.6553814927558627e+09	INFO	k8up.restic.restic.RepoInit.restic.stderr	Fatal: create repository at b2:<redacted>: failed: config already exists
2022-06-16T14:10:41+02:00	1.6553814411108267e+09	INFO	k8up.restic.restic.RepoInit.command	Defining RESTIC_PROGRESS_FPS	{"frequency": 0.016666666666666666}
2022-06-16T14:10:41+02:00	1.6553814411107955e+09	INFO	k8up.restic.restic.RepoInit.command	restic command	{"path": "/usr/local/bin/restic", "args": ["init", "--option", ""]}
2022-06-16T14:10:41+02:00	1.655381441110681e+09	INFO	k8up.restic.restic	using the following restic options	{"options": [""]}
2022-06-16T14:10:41+02:00	1.6553814411104522e+09	INFO	k8up.restic	setting up a signal handler
2022-06-16T14:10:41+02:00	1.6553814411104424e+09	INFO	k8up.restic	initializing
[...]

PR #696, opened 2022-06-23
change to array of str for acceptable errors on repo init

jessebot:
## Summary
This is to fix this issue: https://github.com/k8up-io/k8up/issues/690

We currently only check for "already intitialized", but in the restic backend for backblaze b2, they also throw "config already exists" as a possible error for the same thing. Because both errors are possible and probably fine, I created an array, in case more show up in the future, for easier fixing.

## Checklist

- [ ] Update tests.

### DIFF
--- a/restic/cli/init.go
+++ b/restic/cli/init.go
@@ -46,14 +46,16 @@ func (i *initStdErrWrapper) Write(p []byte) (n int, err error) {
 	scanner := bufio.NewScanner(bytes.NewReader(p))
 
 	// array of acceptable errors to attempt to continue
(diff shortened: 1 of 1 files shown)


Kidswiss:
hi @jessebot 

I've triggered the pipeline here and the E2E tests seem to succeed.

Were you able to verify, that this works with B2? If so this looks good to me.

ccremer (approved):

Kidswiss (approved):
LGTM

Kidswiss:
@jessebot thanks for your contribution! 🎉

PR #738, opened 2022-09-16
Fix operator not marking completed jobs in CRs

ccremer:
## Summary

* Fixes an issue where `Jobs` with `Completed` condition would not cause `Backups`, `Prune` or `Check` to be marked as completed as well. This can be caused when K8up is not running while a `Job` completes (e.g. restart).

### For Code changes


Kidswiss:
I think we need to adjust the condition for adding the callback a bit. This implementation adds the callback each time a started job is reconciled, even if it's already finished. This could potentially pile up more than one of these callbacks.

I don't think we need to adjust the metrics. They are handled by the observer independent of the registered callbacks.

Kidswiss (approved):
After a small discussion it makes sense to merge this as a workaround and then start working on removing the observer as part of a tech debt.

ccremer:
I have summarized the technical debt in https://github.com/k8up-io/k8up/issues/743

PR #746, opened 2022-10-06
Add step required to clear K8upJobStuck

psy-q:
Mention that restarting the k8up operator process is necessary to clear K8upJobStuck

## Summary

* The K8upJobStuck alert does not clear on its own even when the stuck jobs have been removed
* Killing the operator process or deleting the pod will clear the counters and the alert
* Thus, mention this in the documentation

### For Code changes


psy-q:
I was told https://github.com/k8up-io/k8up/releases/tag/v2.5.1 would handle some of this, so closing.

PR #752, opened 2022-10-18
Decrease the queue counter unconditionally

### LINKED ISSUE #751: Wrongly Decreasing the queued metric
### Description

During a debug session where a seemingly wrong `K8upJobStuck` alert kept appearing on a cluster, I found that if there are many backup jobs to the same repo, the likelyhood of wrong stuck alerts is a lot higher.

### Additional Context

Currently, the decrease is only called if there's no other job in the queue pointing to the same repository: https://github.com/k8up-io/k8up/blob/master/operator/queue/execution.go#L90

The `K8upJobStuck` check sees if there is a non-zero amount of jobs in the queue for more than 24h. This is to ensure that the operator is actually able to schedule the given jobs.

The way the logic is intended is, when a job is added to the queue https://github.com/k8up-io/k8up/blob/master/operator/queue/execution.go#L79, it increases the counter. When it leaves the queue it should decrease the counter https://github.com/k8up-io/k8up/blob/master/operator/queue/execution.go#L90. However the decrease is only triggered if there are no other jobs queued for the same repository.

The fix is to just move the decrease outside of that if case.

### Logs

```shell
Not applicable.
```

### Expected Behavior

The counter is always decreased when a job is fetched from the queue.

### Steps To Reproduce

1. Have schedules point to the same repository
2. The schedules trigger at the same time
3. Get the alert

### Version of K8up

v2.5.1

### Version of Kubernetes

any

any

Kidswiss:
## Summary

Fixes #751

### For Code changes

### DIFF
--- a/operator/queue/execution.go
+++ b/operator/queue/execution.go
@@ -87,8 +87,8 @@ func (eq *ExecutionQueue) Get(repository string) Executor {
 	entry := eq.queues[repository].get()
 	if eq.queues[repository].Len() == 0 {
 		delete(eq.queues, repository)
-		eq.decQueueGauge(entry.GetJobNamespace(), entry.GetJobType())
 	}
+	eq.decQueueGauge(entry.GetJobNamespace(), entry.GetJobType())
 	return entry
 }
 


ccremer (approved):

bastjan (approved):

Kidswiss:
I've created a manual image for this branch and asked @haasad to test it on a cluster where we know that this issue happens regularly. Once I've got feedback I'll merge here.

haasad:
I have this version running on one of the problematic clusters now, will observe for a few days and give feedback

ccremer:
Merging to make a prerelease

haasad:
@Kidswiss problem hasn't occurred anymore for a week now on a cluster running with the `ghcr.io/k8up-io/k8up:fix_decrease_queue` image, whereas it appeared every 2-3 days before

Issue #612, opened 2022-02-17
Dissolve EffectiveSchedules CRD


ccremer:
### Summary

**As** developer\
**I want** the generated schedules created by the randomized schedule to be in the status spec of the same resource\
**So that** I can inspect the generated schedule in the status and not via another CRD

### Context

In #333 a new CRD `EffectiveSchedule` was introduced as a first part of trying to "deduplicate" prune jobs. However, the 2nd part never got finished and today, also doesn't need further consideration.

Thus, let's get rid of this CRD and move the `effectiveSchedule` spec property to the status field of the resource. This should reduce the overall complexity of K8up.

Since the effective cron schedule is generated with a stable seed, the schedule should result in the same cron syntax, ensuring that the backups still happen at the same time as before. Thus this feature isn't a breaking change.

### Out of Scope

- Each `EffectiveSchedule` CR is meant to be closed for customization by humans and are thus also in another namespace. If there were modifications done manually, we don't need to ensure them that they are kept and transferred to the new mechanism. Though we should inform/document it accordingly for those that have modified them.

### Further links

- #333 
- #581

### Acceptance Criteria

- Given the existing list of `EffectiveSchedule`, when updating to the next K8up version, then remove `EffectiveSchedule` CRD and all associated CR.
- Given schedules with randomized cron schedules, when updating to the next K8up version, regenerate the effective cron schedule and put the value in the status field (or alternatively, directly into the K8s CronJob object).

PR #774, opened 2022-12-06
Move randomized schedules to Status and remove EffectiveSchedules CRD

ccremer:
## Summary

This PR replaces the `EffectiveSchedule` CRD with a new status field `.status.effectiveSchedules` in the `Schedule` CRD.
See #612 for more context and reasoning.

An `EffectiveSchedule` resource is generally inaccessible to normal users. It was used to store the generated schedules from schedule definitions like `@daily-random` as a preparation for another feature. However, this follow-up feature never landed and now some complexity can be reduced with this PR.

### Upgrade instructions for administrators
The `EffectiveSchedule` is now obsolete and can be removed. Removing it is optional though, they cause no harm if K8up is updated as usually provided the new CRD is still installed though.

For a clean upgrade, proceed as following:
1. Apply the CRDs of K8up v2.6 in order to get the new status field definition in the `Schedule` CRD.
2. Upgrade the deployment of K8up to v2.6
3. Remove the obsolete CRD with `kubectl delete crd effectiveschedules.k8up.io`

Users don't need to change anything, but they will notice new properties in `.status` of `Schedules`.

Labelled as "breaking change" since a new CRD version needs to be applied. Not bumping the major version since it doesn't generally break K8up backups as the generated schedule should stay the same thanks to a deterministic seed.

## Checklist


api/v1/schedule_types.go
```diff
@@ -83,6 +83,13 @@ type ScheduleStatus struct {
 	// They are an extension mechanism which allows tools and other controllers to collect summary information about
 	// resources without needing to understand resource-specific status details.
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	// EffectiveSchedules contains a list of schedules generated from randomizing schedules.
+	EffectiveSchedules []EffectiveSchedule `json:"effectiveSchedules,omitempty"`
```
Kidswiss, line 87: I know that for specs maps are an antipattern for K8s.

But I feel here for the status it would simplify quite a few things?

e.g. map[JobType]EffectiveSchedule?

Kidswiss (approved):
Just one small thought, the rest LGTM

api/v1/schedule_types.go
ccremer: It's not that bad, there's just about 30 LoC that deal with setting/removing effective schedules, not worth it IMO. I wouldn't recommend using an antipattern just to get rid of 30 lines.

Kidswiss: Fine by me.

ccremer:
Merging. Tests fail with an obscure error regarding envtest.
It is mentioned and tracked here: https://github.com/kubernetes-sigs/controller-runtime/issues/2086

PR #785, opened 2022-12-19
Setup golangci-lint (replace CodeClimate)

ccremer:
## Summary

Sets up golangci-lint instead of codeclimate. Codeclimate seems to no longer support Go language.

Note: We do lose the code coverage information in CI, but it never had real consequences anyway

## Checklist


zugao (approved):
Seems there are no issues when running linting and everything works as expected. There's no other way to install the binary than to run a script? Maybe use of docker?

ccremer:
> There's no other way to install the binary than to run a script? Maybe use of docker?

We already install a bunch of binaries through script (e.g. integration test binaries, minio). In fact we're running more with direct binaries than with docker :)
Plus, it's the same setup as https://github.com/vshn/go-bootstrap, so I didn't invest too much in new alternatives.

PR #786, opened 2022-12-20
Pin K8up image tag in chart

ccremer:
## Summary

For Chart v3 we set the image to `v2` with `Always` pull policy. In hindsight, this was a mistake.
We now have sometimes new K8up releases that introduce new non-breaking features also in CRDs, which make an upgrade with CRDs necessary. With `Always` pull policy, there can be new K8up Pod started with old CRDs, causing issues and errors.

This change now pins the tag again.
As a Bonus, specifying `k8up.backupImage` isn't necessary anymore, it defaults to values defined in `image`.

## Checklist

- [x] PR contains the label `area:chart`
- [x] PR contains the chart label, e.g. `chart:k8up`
- [x] Variables are documented in the values.yaml using the format required by [Helm-Docs](https://github.com/norwoodj/helm-docs#valuesyaml-metadata).
- [x] Chart Version bumped if immediate release after merging is planned
- [x] I have run `make chart-docs`
- [x] Link this PR to related code release or other issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,8 +74,6 @@ deploy: kind-load-image install ## Deploy controller in the configured Kubernete
 		--set image.registry=$(E2E_REGISTRY) \
 		--set image.repository=$(E2E_REPO) \
 		--set image.tag=$(E2E_TAG) \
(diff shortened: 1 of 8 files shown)


zugao (approved):
LGTM

Issue #743, opened 2022-09-23
Get rid of observer


ccremer:
### Summary

**As** maintainer of K8up\
**I want** to refactor the operator to remove the internal "observer" module\
**So that** I can reduce complexity and sources of bugs

### Context

The internal observer package is mainly used as a trigger for certain events.
For example, when a backup job is completed, callback functions are called to execute the next step of a backup.
This part is in the code for historic reasons, but today we have better mechanisms: status conditions

To get rid of the observer, we need to achieve the following:
- batch/v1/Job controller that gets notified when jobs change state. We have it already, but we're using it to notify the observer. Here, we need to change code so that the Job's status conditions get translated to conditons that the `Backup` resource understands. We know what the backup object is, since each Job has a controller reference. So the job controller can directly update the Backup's status condition.
- Backup controller: instead of registering callbacks, executes different logic based on the status conditions present.

The same logic mentioned in the backup controller is also applicable to prune, checks etc.

### Out of Scope

Changing CRDs, e.g. preparing for K8up v3

### Further links

- #workar

PR #789, opened 2022-12-21
Free Check and Prune from internal Observer, rework exclusive job queueing

ccremer:
## Summary

Part 3 of a multi-PR code refactoring.

This PR frees up the Check and Prune controllers from the stateful "observer" code. That way, the operator is a bit more stateless, resp. Kubernetes handles the state for us. That should make the operator more resilient for restarts.
For the user, the change shouldn't be noticeable.

This PR finally gets rid of the observer. The observer contained logic to determine whether an "exclusive" or priority job is running already and if so, holds back all other jobs. This logic has been replaced by a simple listing and filtering all currently running jobs that target a specific Restic repository.

Other Refactorings:

* Make dedicated packages for the k8upv1/{Check, Prune} controller and combine with handler.

Known Issues:

* Testing is still hard, due to the job runner executor that is still stateful. In a later PR, we should get rid of the job queue and find a different mechanism for priority queueing exclusive jobs. Once we have this, we can reconcile K8up jobs much more conveniently.
* Cleanup still doesn't work. I may have broken something but with getting rid of the executor we should be able to make the cleanup as part of the normal reconciliation process.

Part of #743 

## Checklist


operator/executor/worker.go
```diff
@@ -55,13 +54,22 @@ func (qe *QueueWorker) loopRepositoryJobs(repository string) {
 		if job.Exclusive() {
 			// TODO: discard an exclusive job if there's any other exclusive job running
```
Kidswiss, line 55: We could probably get rid of this todo, doesn't seem relevant anymore.

Kidswiss (approved):
LGTM

PR #801, opened 2023-01-23
feat: add support for RWO PVCs backups

### LINKED ISSUE #319: Integrated Support for RWO PVCs File Backups
## Summary

**As** user of K8up
**I want** to backup PVCs with RWO access mode
**So that** my precious files stored in RWO volumes are backed up as well.

## Context

Backing up RWO volumes is currently not directly supported by K8up and is only possible via workarounds.

## Out of Scope

* -

## Further links

* https://kubernetes.io/docs/concepts/storage/persistent-volumes/#node-affinity

## Acceptance criteria

```gherkin
Given a PVC with RWO access mode in Bound phase
And PVC has `k8up.io/backup=false` annotation
When a backup is scheduled
Then the backup for this PVC is skipped
```

```gherkin
Given a PVC with RWO access mode _not_ in Bound phase
When a backup is scheduled
Then the backup for this PVC is skipped
```

```gherkin
Given a PVC with RWO access mode in Bound phase
And an application Pod has mounted that PVC
When a backup runs
Then the backup Pod runs on the same node as the app Pod.
And the files are backed up.
```

```gherkin
Given a PVC with RWO access mode in Bound phase
And an application Pod has mounted that PVC
And the app Pod has the `k8up.io/backupcommand` annotation
When a backup runs
Then the backup is done via existing backup command
```

```gherkin
Given a PVC with RWO access mode in Bound phase
And no application Pod has mounted that PVC
When a backup runs
Then K8up will fail a backup since it can't reliably determine the target node where the backup Pod should run.
```

```gherkin
Given a PVC with RWO access mode in Bound phase
And no application Pod has mounted that PVC
And the linked PV has a NodeAffinity rule defined
When a backup runs
Then the backup Pod runs on the same node as configured via NodeAffinity spec.
And the files are backed up.
```

```gherkin
Given a PVC with RWO access mode in Bound phase
And no application Pod has mounted that PVC
And the PVC has the `k8up.io/hostname` annotation
When a backup runs
Then the backup Pod runs on the node as configured via annotation spec.
And the files are backed up.
```

```gherkin
Given a PVC with RWO access mode in Bound phase
And no application Pod has mounted that PVC
And the PVC has the `k8up.io/hostname` annotation
And the linked PV has a NodeAffinity rule defined
When a backup runs
Then the backup Pod runs on the same node as configured via NodeAffinity spec since that takes precedence in Kubernetes
And the files are backed up.
```

## Implementation Ideas

During the implementation of #12 we found that with the correct Pod affinity a RWO PV can be mounted to a second Pod, given it's on the same host.

With that information it should be possible to support file backups on RWO PVCs with the following process:

* one backup job for all RWX and BackupCommands as already implemented
* for each RWO marked for backup we'll spawn another job with correct Pod affinity, so it will be scheduled on the node where the RWO PVC is mounted

File backups are much more reliable and faster. Until now we use the BackupCommands to stream the backups via stdin/stdout to wrestic for RWO PVCs.

Of course if any application aware things (DB dumps, etc.) should still be done via BackupCommands.

Unfortunately unlike the backup commands the Pod affinity needs to be set in the operator, which has its own set of problems:

* the Pod affinity is fixed at the time of creation, so if something changes on the hosts (reboots, etc.) the Pod affinity may not match correctly anymore -> may be solvable via clever podaffinity and retries
* a lot more Pods may get created during backups (performance issues?)
* but things can be run parallel

mweibel:
## Summary

* Support RWO PVCs backups
* Adjust local e2e testing & cluster to be able to test it

fixes #319

### For Code changes


mweibel:
While implementing this I noticed (a bit late unfortunately), that the job controller marks a parent as finished (https://github.com/k8up-io/k8up/blob/master/operator/jobcontroller/controller.go#L112). E.g. when a Backup creates a Job, the job controller watches the job and once it is done, marks the owner (the backup in this case) as finished. 
This does not work when there are more than one jobs created per Backup.

I'll refactor this in a separate PR. Unfortunately this might delay the implementation of this feature.

mweibel:
current status: 
because we now spawn multiple jobs per backup, this requires a refactoring (see #806, #807).
Otherwise it *seems* to be working - I just couldn't verify it all yet, unfortunately.

mweibel:
FYI changed base from master to `owner-reconcile` since those changes are needed for this.

mweibel:
Open issue: if restic hasn't initialized the repository yet and more than one job concurrently tries to create the repo, all subsequent attempts to use such repo fail since restic initializes the repo twice with different keys and then complains with the following error:
```
INFO    k8up.restic.restic.unlock.restic.stderr Fatal: config or key <REDACTED> is damaged: ciphertext verification failed
```

mweibel:
> Open issue: if restic hasn't initialized the repository yet and more than one job concurrently tries to create the repo, all subsequent attempts to use such repo fail since restic initializes the repo twice with different keys and then complains with the following error:
> 
> ```
> INFO    k8up.restic.restic.unlock.restic.stderr Fatal: config or key <REDACTED> is damaged: ciphertext verification failed
> ```

this should be fixed now.

the e2e test works but it's late in the day and I'll need to verify everything again next working day.

operator/backupcontroller/executor.go
```diff
@@ -42,54 +46,122 @@ func (b *BackupExecutor) Execute(ctx context.Context) error {
...
+		if !containsAccessMode(pvc.Spec.AccessModes, corev1.ReadWriteMany) && !isRWO && !hasBackupAnnotation {
+			log.Info("PVC is neither RWX nor RWO and has no backup annotation", "pvc", pvc.GetName())
 			continue
 		}
 
-		if !ok {
-			log.Info("PVC doesn't have annotation, adding to list", "pvc", item.GetName())
-		} else if anno, _ := strconv.ParseBool(tmpAnnotation); !anno {
-			log.Info("PVC skipped due to annotation", "pvc", item.GetName(), "annotation", tmpAnnotation)
+		if !hasBackupAnnotation {
```
Kidswiss, line 97: Do I see that right, that this basically inverts the functionality of the annotation?

E.g. Instead of annotate to include, we should annotate to exclude?

If so we should probably treat this as a breaking change.

operator/backupcontroller/executor.go
```diff
@@ -42,54 +46,122 @@ func (b *BackupExecutor) Execute(ctx context.Context) error {
...
+
+		backupItems = append(backupItems, bi)
 	}
 
-	return volumes, nil
+	return backupItems, nil
 }
 
-func (b *BackupExecutor) startBackup(ctx context.Context) error {
+// findNode tries to find a PVs NodeAffinity for a specific hostname. If found will return that.
```
Kidswiss, line 146: Reminds me of sleep sort:

```go
for i := range numbers {
    go func(){
        time.Sleep(numbers[i]*time.Second)
        fmt.Println(numbers[i])
    }
}
```

Kidswiss (commented):
So far the code LGTM.

operator/backupcontroller/executor.go
mweibel: haha yeah :smile:

operator/backupcontroller/executor.go
Kidswiss: Yeah, you're right, I didn't see the negation in the old version :)

mweibel:
regarding testing: I did local tests but not in real clusters. I'd advise to release a beta/RC first and test it before making a stable release. It should not fail but it's rather complex so you never know :slightly_smiling_face:

e2e/lib/k8up.bash
```diff
@@ -126,12 +129,34 @@ given_an_annotated_subject() {
...
+	echo "✅  The pvc rwo worker subject is ready"
+}
+
+given_a_rwo_pvc_subject_in_controlplane_node() {
+	require_args 2 ${#}
+
+	export BACKUP_FILE_NAME=${1}
+	export BACKUP_FILE_CONTENT=${2}
+
+	yq e 'with(select(document_index == 2) .spec.template.spec; .containers[0].securityContext.runAsUser='$(id -u)' | .containers[0].env[0].value=strenv(BACKUP_FILE_CONTENT) | .containers[0].env[1].value=strenv(BACKUP_FILE_NAME))' definitions/pvc-rwo-subject/controlplane.yaml | kubectl apply -f -
```
Kidswiss, line 155: Is there a reason you run the control plane with `runAsUser` and the worker one not? This leads to a permission denied on my mac...

e2e/definitions/kind/config.yaml
```diff
@@ -5,3 +5,9 @@ nodes:
   extraMounts:
   - hostPath: ./e2e/debug/data
     containerPath: /tmp/e2e
+- role: worker
+  labels:
+    worker: true
+  extraMounts:
```
Kidswiss, line 11: Are the extraMounts actually necessary? They break the e2e-tests on my mac. Without them, they run mostly through. Except for the remark below.

Kidswiss:
Just two remarks about the e2e-tests. Once they're fixed you can merge.

The rest LGTM.

Kidswiss (changes requested):
Just two remarks about the e2e-tests. Once they're fixed you can merge.

The rest LGTM.

e2e/lib/k8up.bash
mweibel: good catch - not intentional, was an oversight. I'll double check everything on macOS again. :+1:

e2e/definitions/kind/config.yaml
mweibel: I forgot to update the worker/controlplane PVC. That's fixed now and works on my macOS. Can you check again? make sure to `make kind-clean` first.

Kidswiss: Werid, it still fails for me. But it's okay if it's working for you, I'll have a look at this another time.

Kidswiss (approved):

PR #813, opened 2023-02-24
Save available snapshots into CRs

Kidswiss:
## Summary

After a backup and prune job, K8up will now write all the snapshots back to the k8s cluster.

There are a few edge cases that are handled as well:

* if the repository contains backups from multiple namespaces it will filter them accordingly
* if the namespace has backups to multiple repositories they will also be filtered correctly

The result should be a list of snapshots in each namespace that mirrors what K8up is managing.

### For Code changes


Kidswiss:
To test it locally you can run `gmake e2e-test -e BATS_FILES=test-07-snapshots.bats`.

This will run both edge cases mentioned in the summary. Once it's finished you can connect to the kind cluster and manually inspect the snapshots and create new backups if you wish.

restic/kubernetes/snapshots.go
```diff
@@ -0,0 +1,129 @@
...
+// diff will execute a given function for each element that is in a but not b
+// it assumes that both lists belong to the same namespace and repository
+func diff(a, b *k8upv1.SnapshotList, diffFunc func(snap k8upv1.Snapshot) error) error {
+
+	snapMap := listToMap(b)
+
+	for _, snapshot := range a.Items {
+		// Avoid pointer bug
+		snapshot := &snapshot
+		if _, ok := snapMap[*snapshot.Spec.ID]; !ok {
```
zugao, line 62: Nitpick. I would have implemented a contain method for the list instead of using a map. Though I know that go optimised very well maps for object search so I'm ok with that. It just feels it's more of an internal thing to list instead of putting logic in `snapshot.go`

restic/kubernetes/snapshots.go
```diff
@@ -0,0 +1,129 @@
...
+	for _, snapshot := range list {
+		// Avoid pointer bug
+		snapshot := snapshot
+
+		// we don't want snapshots that belong to another namespace in here.
+		if snapshot.Hostname != namespace {
+			continue
+		}
+
+		finalList.Items = append(finalList.Items, k8upv1.Snapshot{
```
zugao, line 100: Any reason why you don't reuse the existing snapshot?

zugao (approved):
Added some small things, overall great. Haven't tested it.

restic/kubernetes/snapshots.go
Kidswiss: What do you mean by existing? I'm converting the snapshtos from the `dto.Snapshot` to `k8upv1.Snapshot` here. One format is what restic gives me an the other one is what the CRD defines.

restic/cli/restore.go
```diff
@@ -138,7 +139,7 @@ func (r *Restic) getLatestSnapshot(snapshotID string, log logr.Logger) (Snapshot
 	return snapshot, err
 }
 
-func (r *Restic) folderRestore(restoreDir string, snapshot Snapshot, restoreFilter string, verify bool, log logr.Logger) error {
+func (r *Restic) folderRestore(restoreDir string, snapshot dto.Snapshot, restoreFilter string, verify bool, log logr.Logger) error {
```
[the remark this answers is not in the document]
Kidswiss: Yeah that's something we should look at in another context. We could probably optimize a lot in the restic packages.

docs/modules/ROOT/pages/how-tos/restore.adoc
```diff
@@ -2,6 +2,53 @@
...
+  namespace: default
+spec:
+  date: "2023-03-03T07:34:42Z"
+  id: 162e7a85acbc14de93dad31a3699331cb32187ff0d7bd2227b7c4362a1d13a42
+  paths:
+  - /data/subject-pvc
+  repository: s3:http://minio.minio.svc.cluster.local:9000/backup
+----
+
+The paths are in the format of `/data/$PVCNAME`.
```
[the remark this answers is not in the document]
Kidswiss: Yeah I thought about this, too. But there are also stdin backups that don't have a PVC associated. I found that leaving it empty in such cases doesn't seem like a good solution...

restic/kubernetes/snapshots.go
zugao: I would have internalised this in the snapshot with a map

restic/kubernetes/snapshots.go
zugao: Ah, I missed that, they are 2 different objects, it's okey then :).

Kidswiss:
Thanks for the review.

PR #814, opened 2023-02-26
fix: exclude non-running pods from podlist

huettner94:
## Summary

previously the executor would happily set `TARGET_PODS` for the backup job to the pod of a previous running backup instead of a actually running pod.
This prevents `k8up.io/backupcommand` from working, as then the backup job would actually check the old incarnation of itself instead of the real target pod.

### For Code changes

- [X] Link this PR to related issues
- [X] I have not made _any_ changes in the `charts/` directory.

### DIFF
--- a/operator/backupcontroller/executor.go
+++ b/operator/backupcontroller/executor.go
@@ -13,6 +13,7 @@ import (
 	"github.com/k8up-io/k8up/v2/operator/job"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -60,7 +61,8 @@ func (b *BackupExecutor) listAndFilterPVCs(ctx context.Context, annotation strin
 
 	pods := &corev1.PodList{}
 	pvcPodMap := make(map[string]corev1.Pod)
-	if err := b.Config.Client.List(ctx, pods, client.InNamespace(b.backup.Namespace)); err != nil {
+	labelselector, _ := labels.Parse("!" + job.K8uplabel)
+	if err := b.Config.Client.List(ctx, pods, client.InNamespace(b.backup.Namespace), client.MatchingLabelsSelector{Selector: labelselector}); err != nil {
 		return nil, fmt.Errorf("list pods: %w", err)
 	}
 	for _, pod := range pods.Items {
--- a/operator/backupcontroller/controller_integration_test.go
+++ b/operator/backupcontroller/controller_integration_test.go
@@ -15,6 +15,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


huettner94:
Unfortunately it does not seem like i can set labels on this MR

huettner94:
Thanks @Kidswiss for the feedback.

During fixing the issue i found that we already set a label on all pods used for backups. I then decided that using this label would be more reliable than filtering for running pods.

(also a integrationtest is now added)

Kidswiss:
> Thanks @Kidswiss for the feedback.
> 
> During fixing the issue i found that we already set a label on all pods used for backups. I then decided that using this label would be more reliable than filtering for running pods.
> 
> (also a integrationtest is now added)

Yeah, you're right, that makes more sense. I've triggered the tests, let's see how they go.

operator/backupcontroller/executor.go
```diff
@@ -60,7 +61,8 @@ func (b *BackupExecutor) listAndFilterPVCs(ctx context.Context, annotation strin
 
 	pods := &corev1.PodList{}
 	pvcPodMap := make(map[string]corev1.Pod)
-	if err := b.Config.Client.List(ctx, pods, client.InNamespace(b.backup.Namespace)); err != nil {
+	labelselector, _ := labels.Parse("!" + job.K8uplabel);
+	if err := b.Config.Client.List(ctx, pods, client.InNamespace(b.backup.Namespace), client.MatchingLabelsSelector{labelselector}); err != nil {
```
Kidswiss, line 65: The lint job is complaining here to not use an unkeyed field in the `client.MatchingLabelsSelector` literal.

Kidswiss (changes requested):
Just one small linting issue and then it should be ready for merging.

operator/backupcontroller/executor.go
huettner94: fixed

huettner94:
thanks fixed

Kidswiss:
@huettner94 unfortunately the linter found something new to complain 😅 

You can run the linter locally by using `make lint`. Then you won't have to wait for me to trigger the CI/CD every time.

huettner94:
Just pushed a fix for the lint. I could not reproduce the issue with the e2e tests locally. they worked fine for me

Kidswiss:
Hi @huettner94 

Sorry for the late answer was on a vacation.

The e2e looks good now. Sometime they can be a bit fiddly....

Can you please run `make docs-all` and then commit the changes? That should shut up the linter for good.

After that I think we're ready to merge here.

huettner94:
> Hi @huettner94
> 
> Sorry for the late answer, was on a vacation.
> 
> The e2e looks good now. Sometime they can be a bit fiddly....
> 
> Can you please run `make docs-all` and then commit the changes? That should shut up the linter for good.
> 
> After that I think we're ready to merge here.

No problem, vacation is always great.

There are some issues with the `make docs-all`:
1. It references some `playbook.yml` in `$(out_dir)/index.html: playbook.yml $(pages)`, but this file does no longer exist. It was removed in https://github.com/k8up-io/k8up/pull/446
2. The used docker image `vshn/antora:2.3.3` seems to no longer exist on dockerhub (same goes for `vshn/asciidoctor-pdf:1.8.1`)

If i take the `playbook.yml` from the old commit i get to the needed docker images, but there i cant really fix something

Kidswiss:
Huh, weird that the image is no longer there. I feel like this needs more investigation with our docs guy and is a separate issue...

I'm going ahead and merge it anyway and fix the linting/docs in another PR.

Thanks for the contribution! 🎉

PR #831, opened 2023-03-24
Cli snapshot restore

wejdross:
## Summary

* adding ability to restore objects using CLI command line, possible are restores to PVC as well as S3 to S3
### For Code changes

NOTE:
These things are not required to open a PR and can be done afterwards,
while the PR is open.
-->


cmd/cli/main.go
```diff
@@ -0,0 +1,361 @@
...
+		Subcommands: []*cli.Command{
+			{
+				Name:   "restore",
+				Action: RunRestore,
+				Flags: []cli.Flag{
+					&cli.StringFlag{
+						Destination: &restore.Cfg.Snapshot,
+						Required:    true,
+						Name:        "snapshot",
+						Usage:       "Required ; ID of the snapshot `kubectl get snapshots`, set, via cli or via env: ",
```
Kidswiss, line 34: Setting this to `latest` by default could be an UX improvement.

cmd/cli/main.go
```diff
@@ -0,0 +1,361 @@
...
+						Usage:       "Optional ; Set RestoreToS3SecretPasswordKey, key inside secret, under which Restic repo password is stored, via cli or via env: ",
+						EnvVars: []string{
+							"RESTORETOS3SECRETPASSWORDKEY",
+						},
+					},
+					&cli.StringFlag{
+						Destination: &restore.Cfg.Namespace,
+						Required:    false,
+						DefaultText: "default",
+						Name:        "namespace",
```
Kidswiss, line 184: Nit: I'd add a short form `-n` because it's what most k8s users would be familiar with.

See Aliases docs: https://github.com/urfave/cli/blob/main/docs/v2/examples/flags.md#alternate-names

cmd/cli/main.go
```diff
@@ -0,0 +1,361 @@
...
+					&cli.StringFlag{
+						Destination: &restore.Cfg.RestoreName,
+						Required:    false,
+						Name:        "restoreName",
+						Usage:       "Optional ; Set restoreName - metadata.Name field, if empty, k8up will generate name, via cli or via env: ",
+						EnvVars: []string{
+							"RESTORENAME",
+						},
+					},
+					&cli.StringFlag{
```
Kidswiss, line 123: Why not use the `&cli.IntFlag` here? Then you don't have to parse it yourself.

cmd/cli/main.go
```diff
@@ -0,0 +1,361 @@
...
+						EnvVars: []string{
+							"S3SECRETREF",
+						},
+					},
+					&cli.StringFlag{
+						Destination: &restore.Cfg.RestoreMethod,
+						Required:    true,
+						Name:        "restoreMethod",
+						DefaultText: "pvc",
+						Usage:       "Required ; Set restore method [ pvc|s3 ], via cli or via env: ",
```
Kidswiss, line 80: Nit: You can use flag actions to enforce this enum: https://github.com/urfave/cli/blob/main/docs/v2/examples/flags.md#flag-actions

cmd/cli/main.go
```diff
@@ -0,0 +1,361 @@
...
+			snapshot = (*snap.Spec.ID)
+			break
+		}
+	}
+	// nothing to do if there are no snapshots
+	if len(snapshot) == 0 {
+		logger.Error(err, "Snapshot ID wasn't found")
+		return err
+	}
+	logger.Info("Found correct snapshot", "SNAPSHOT", snapshot)
```
Kidswiss, line 263: Nit: maybe make this a debug message.
```suggestion
	logger.V(1).Info("Found correct snapshot", "SNAPSHOT", snapshot)
```

docs/modules/ROOT/examples/tutorial/backup.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   failedJobsHistoryLimit: 2
   successfulJobsHistoryLimit: 2
-  promURL: http://minio:9000
+#  promURL: http://minio:9000
```
Kidswiss, line 8: Nit: We could probably remove this completely from the samples? It doesn't work anyway out of the box without a prometheus pushgw.

docs/modules/ROOT/pages/how-tos/restore.adoc
```diff
@@ -230,3 +230,106 @@ dr-xr-xr-x  1 simonbeck  staff    0 Jun  8 09:21 tags
 ----
 
 Here you can browse all backups by host, ids, snapshots or tags.
+
+=== Restore via CLI
+
+K8up supports restoring via CLI. It's fast and simple way to create restore.k8up.io/v2 CRD.
```
Kidswiss, line 236:
```suggestion
K8up supports restoring via CLI. It's a fast and simple way to create restore.k8up.io/v2 CRD. 
```

docs/modules/ROOT/pages/how-tos/restore.adoc
```diff
@@ -230,3 +230,106 @@ dr-xr-xr-x  1 simonbeck  staff    0 Jun  8 09:21 tags
...
+  --s3secretRef minio-credentials                \
+  --snapshot f0dd5684                            \
+  --claimName wordpress-pvc                      \
+  --runAsUser 0                                  \
+  --restoreToS3Bucket backup2                    \
+  --restoreToS3Secret minio-credentials          \
+  --restoreToS3Endpoint http://minio:9000
+----
+
+It can be a little bit confusing to use all those flags, but it's pretty simple
```
Kidswiss, line 311: I wouldn't write this here. Just go straight to a simple example.

docs/modules/ROOT/pages/how-tos/restore.adoc
```diff
@@ -230,3 +230,106 @@ dr-xr-xr-x  1 simonbeck  staff    0 Jun  8 09:21 tags
...
+
+[source]
+----
+# PVC restore
+  (where backup is)
```
Kidswiss, line 316:
```suggestion
  (where the backup is)             
```

docs/modules/ROOT/pages/how-tos/restore.adoc
```diff
@@ -230,3 +230,106 @@ dr-xr-xr-x  1 simonbeck  staff    0 Jun  8 09:21 tags
...
+  --s3endpoint              --runAsUser          --claimName
+  --s3bucket                --kubeconfig                      
+  --s3secretRef             --snapshot
+                            --namespace
+                            --restoreMethod
+                            --secretRef
+                          
+# s3 to s3 restore
+
+  (where backup is)                             (different s3 where we want to copy our backup)
```
Kidswiss, line 327:
```suggestion
  (where the backup is)                             (different s3 where we want to copy our backup)
```

docs/modules/ROOT/pages/how-tos/restore.adoc
```diff
@@ -230,3 +230,106 @@ dr-xr-xr-x  1 simonbeck  staff    0 Jun  8 09:21 tags
...
+  --claimName wordpress-pvc                      \
+  --runAsUser 0                                  \
+  --restoreToS3Bucket backup2                    \
+  --restoreToS3Secret minio-credentials          \
+  --restoreToS3Endpoint http://minio:9000
+----
+
+It can be a little bit confusing to use all those flags, but it's pretty simple
+
+[source]
```
Kidswiss, line 313: I like this snipped, it really makes it simple to understand what flag is for what. 👍

Kidswiss (changes requested):
I really like this feature! Looks very good!

This is a first stepping stone towards some major redesigns we have planned for k8up v3 (or whatever version we'll end up with).

I found some small nitpicks, but all in all LGTM.

cmd/cli/main.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+
+	kconfig, err := clientcmd.BuildConfigFromFlags("", restore.Cfg.Kubeconfig)
+	if err != nil {
+		logger.Error(err, "Failed to create KubeConfig")
+		return err
+	}
+	logger.Info("Found Kubeconfig", "KUBECONFIG", restore.Cfg.Kubeconfig)
+
+	client, err := kubernetes.NewForConfig(kconfig)
```
Kidswiss, line 231: Nit: You could use the typed client here. It's a bit more comfortable to use than the rest client. See here for an example: https://github.com/k8up-io/k8up/blob/master/restic/kubernetes/config.go#L45

cmd/cli/main.go
wejdross: I disagree - it can be nice feature but ONLY if users are 100% sure what they do, otherwise they can mess with their apps while just playing arround with cli

cmd/cli/main.go
wejdross: we can make a story for that later, at this moment I don't see any advantage to use it

cmd/cli/main.go
```diff
@@ -0,0 +1,375 @@
...
+			{
+				Name:   "restore",
+				Action: RunRestore,
+				Flags: []cli.Flag{
+					&cli.StringFlag{
+						Destination: &restore.Cfg.Snapshot,
+						Required:    false,
+						Name:        "snapshot",
+						Value:       "latest",
+						Usage:       "Optional ; ID of the snapshot `kubectl get snapshots`, if left empty 'latest' will be used, set itvia cli or via env: ",
```
zugao, line 34:
```suggestion
						Usage:       "Optional ; ID of the snapshot `kubectl get snapshots`, if left empty 'latest' will be used, set it via cli or via env: ",
```
typo

cmd/cli/main.go
```diff
@@ -0,0 +1,375 @@
...
+
+func RunRestore(ctx *cli.Context) error {
+	var restoreName, snapshot string
+	var s3, pvc v1.RestoreMethod
+	logger := cmd.AppLogger(ctx).WithName("cli-restore")
+
+	// avoid name crashes + generate name if naming is not important for user
+	if restore.Cfg.RestoreName == "" {
+		restoreName = "cli-restore-" + restore.RandomStringGenerator(5)
+		log.Println("Creating restore with name: ", restoreName, " in napespace: ", restore.Cfg.Namespace)
```
zugao, line 231: Is it ok to have two types of logs here? there's `log.` and `logger.`

cmd/cli/main.go
```diff
@@ -0,0 +1,375 @@
...
+		// lines below looks very weird, but it's necessary, as actual reason why request failed is hidden in status.Message variable
+		var out1 metav1.Status
+		err := json.Unmarshal(status, &out1)
+		if err != nil {
+			logger.Error(err, "Failed to unmarshal Status object")
+		}
+		logger.Error(err, out1.Message)
+		return err
+	}
+	logger.Info(fmt.Sprintf("Backup created successfully, You can find it running:\tkubectl -n %s get restores.k8up.io %s", restore.Cfg.Namespace, restoreName))
```
zugao, line 372: I would avoid adding suggestions in the log. We better add them in documentation or in the description when running the app with `--help` argument.

cmd/cli/main.go
```diff
@@ -0,0 +1,375 @@
...
+					&cli.StringFlag{
+						Destination: &restore.Cfg.Snapshot,
+						Required:    false,
+						Name:        "snapshot",
+						Value:       "latest",
+						Usage:       "Optional ; ID of the snapshot `kubectl get snapshots`, if left empty 'latest' will be used, set itvia cli or via env: ",
+						EnvVars: []string{
+							"SNAPSHOT",
+						},
+					},
```
zugao, line 38: Can we somehow reduce the boilerplate code by adding a function with global vars which will create the object for us? It will be more succinct, less boilerplate and better for readability.

zugao (dismissed):
I haven't tested the code. I would also add some unit tests for critical parts of the code.

cmd/cli/main.go
wejdross: yes - that's UX improvement

cmd/cli/main.go
wejdross: why? I really like when software do something like that :D

cmd/cli/main.go
wejdross: explain please

cmd/cli/main.go
Kidswiss: I don't see a problem here either.

zugao: All these kind of suggestions on how restore works or what are the next steps after the restore has been done is very much useful in the docs, either in the repo or in the cli itself. Logs are supposed to tell what steps have been done by the application not what the user is supposed to do. Duplicating this log information for each restore does not seem very wise though I can leave with it as long as this is also documented in cli help or in the repo.

cmd/cli/main.go
zugao: Suggestion `getFlag(type, globVar1, globVar2....) `or `getStringFlag(globVar1, globVar2....)`. Try this one and see if it helps with readability and with reduced boilerplate code.

wejdross: it's yet another place where something can fail, I don't agree it'll solve any problem, but only introduce new, this code must be somewhere and adding another layer of abstraction won't solve any issue

Kidswiss: I wouldn't add something like this here. All other commands in K8up use this pattern. Using a new pattern here breaks consistency.

We could reduce the amount of lines by making all the structs one-liner like in `restic/main.go` but I actually think this here is more readable.

cmd/cli/main.go
wejdross: I'll add it to docs additionally before sprint review

cmd/cli/main.go
zugao: Then lets keep it as is.

cmd/cli/main.go
zugao: If the docs will be updated then I'm fine with it.

cmd/cli/main.go
```diff
@@ -140,7 +140,7 @@ var (
 						Destination: &restore.Cfg.RunAsUser,
 						Required:    false,
 						Name:        "runAsUser",
-						Value:       0,
+						Value:       -1,
```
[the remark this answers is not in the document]
wejdross: solved it by using `ctx.IsSet()` method

Kidswiss (approved):

PR #852, opened 2023-05-05
Improve RBAC for the operator

Kidswiss:
## Summary

Previously the operator had the `pod/exe` permissions with the verb
`create`. Which is necessary for K8up to create and bind the namespaced
roles where the actual backups happen.
    
To harden the RBAC a bit we're now deploying a clusterRole which
contains the necessary RBAC rules. Also K8up is now allowed to bind
exactly this clusterRole.
    
While this doesn't mitigate all potential exploits in case the K8up
operator pod is accessed by a malicious party, it makes it harder for
the malicious party to abuse K8up's permissions.

### For Code changes

### DIFF
--- a/charts/k8up/templates/executor-clusterrole.yaml
+++ b/charts/k8up/templates/executor-clusterrole.yaml
@@ -1,0 +1,35 @@
+{{- if .Values.rbac.create -}}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: 'k8up-executor'
+  labels:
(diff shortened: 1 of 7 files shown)


Kidswiss:
Even though we should not change anything in the `/chart` folder, for this to work with the e2e I have to change the chart here as well. Let's hope the CI/CD process doesn't break too hard due to this...

zugao (approved):
LGTM. We might need a permission diagram for k8up in order to understand how permissions work in k8up.

glrf (approved):

Issue #803, opened 2023-01-24
K8up is unable to restore a single file snapshot


bfeher-kx:
### Description

I've created an instance from ghcr.io/k8up-io/k8up:v2 and set the followings:

```
  - args:
    - -restore
    - -restoreSnap
    - 207d6a74
    - -restoreType
    - folder
```
The Snapshot id points to a single .sql file which has been created via K8up application aware backup from a PostgreSQL instance. It fails with with the following error: panic: runtime error: slice bounds out of range [:3] with capacity 2.

The snapshot above can be restored manually via the Restic binary packed in the same image. 

### Additional Context

My goal was to restore to a RWO volume thus used podAffinity spec to get my pod scheduled beside the target pod.

Tried with the operator as well, pointing to a RWX volume, it fails too.

### Logs

```shell
1.6745719441801126e+09    INFO    k8up.restic    initializing                                                                                                                                  
1.674571944180123e+09    INFO    k8up.restic    setting up a signal handler                                                                                                                    
1.6745719441801894e+09    INFO    k8up.restic.restic    using the following restic options    {"options": [""]}                                                                                
1.6745719441802113e+09    INFO    k8up.restic.restic.RepoInit.command    restic command    {"path": "/usr/local/bin/restic", "args": ["init", "--option", ""]}                                 
1.6745719441802182e+09    INFO    k8up.restic.restic.RepoInit.command    Defining RESTIC_PROGRESS_FPS    {"frequency": 0.016666666666666666}                                                   
1.6745719442500741e+09    INFO    k8up.restic.restic.unlock    unlocking repository    {"all": false}                                                                                          
1.6745719442501287e+09    INFO    k8up.restic.restic.unlock.command    restic command    {"path": "/usr/local/bin/restic", "args": ["unlock", "--option", ""]}                                 
1.6745719442501488e+09    INFO    k8up.restic.restic.unlock.command    Defining RESTIC_PROGRESS_FPS    {"frequency": 0.016666666666666666}                                                     
1.6745719448515744e+09    INFO    k8up.restic.restic.unlock.restic.stderr    successfully removed locks                                                                                        
1.6745719448527825e+09    INFO    k8up.restic.restic.snapshots    getting list of snapshots                                                                                                    
1.6745719448528092e+09    INFO    k8up.restic.restic.snapshots.command    restic command    {"path": "/usr/local/bin/restic", "args": ["snapshots", "--option", "", "--json"]}                 
[...]

PR #911, opened 2023-11-25
Bug/fix restore single file backup command

poyaz:
## Summary

* This PR solves restoring a single snapshot file (when using `backupcommand` annotation). Issue: #803

I added a function to check if the snapshot contains one file, no need to create a folder and then symlinked this folder with a real restore path (In #803, when the restore snapshot makes panic because of the use of PVC path)

* Also I solved the problem in the make file when using another build of `yq` package (File: `e2e/kind.mk`)

### For Code changes

### DIFF
--- a/restic/cli/restore.go
+++ b/restic/cli/restore.go
@@ -242,12 +242,27 @@ func (r *Restic) isRestoreSingleFile(log logr.Logger, snapshot dto.Snapshot) (bo
 
 	stdOutLines := strings.Split(capturedStdOut, "\n")
 
+	if len(stdOutLines) == 0 {
+		err := fmt.Errorf("no list exist for snapshot %v", snapshot.ID)
+		log.Error(err, "the snapshot list is empty")
+		return false, err
+	}
+
+	err := json.Unmarshal([]byte(stdOutLines[0]), &dto.Snapshot{})
+	if err != nil {
+		return false, err
+	}
+
 	count := 0
-	for _, fileJSON := range stdOutLines {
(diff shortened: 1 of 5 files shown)


restic/cli/restore.go
```diff
@@ -221,6 +227,47 @@ func (r *Restic) linkRestorePaths(snapshot dto.Snapshot, restoreDir string) (str
...
+	capturedStdOut := buf.String()
+
+	stdOutLines := strings.Split(capturedStdOut, "\n")
+
+	count := 0
+	for _, fileJSON := range stdOutLines {
+		node := &fileNode{}
+		err := json.Unmarshal([]byte(fileJSON), node)
+		if err != nil {
+			continue
```
Kidswiss, line 250: I'd prefer to log the error and not just discard it here. If for some reason the JSON is garbled it would probably take a long time to figure that out.

Kidswiss (changes requested):
Hey!

Thanks for the contribution! I just have one small nitpick, the rest LGTM.

I've also triggered the e2e tests.

restic/cli/restore.go
poyaz: Yeah, you are right, it's not good to skip error. I've just wrote this part of code like `extractFileNodes` function

I change this part of code and handling error to seem more understanding

poyaz: I fix it

Kidswiss (approved):
LGTM now!

Thanks!

PR #925, opened 2024-02-19
Helm Chart update: make executor ClusterRole unique

bgounon:
## Summary

Executor ClusterRole name should be dynamic instead of hard-coded. Without it, you cannot have several namespaced releases of k8up, as it triggers a collision on the executor ClusterRole name and fails.

### For Helm Chart changes


Kidswiss:
Hi @bgounon 

Thank you for the PR!

I like the idea, but just making the name dynamically based on the fullname could also break single installations, if a non-default name is specified.

Instead, you could add a new value under `rbac` like `disableClusterRole`. Then the first k8up install can install the clusterrole and all subsequent ones don't.

Another approach could be to disable the rbac completely and deploy the `ClusterRoles` separately, via the value `rbac.create`.

arska:
Thanks for taking the time to contribute this! We really appreciate it.

Since there hasn't been any activity on this PR for about 2 years, we're going to close it for now to keep things tidy. If this is still relevant, feel free to reopen or submit a new PR — we'd be happy to review it.

Issue #792, opened 2022-12-29
QNAP QuObjects support


sebt3:
### Description

QNAP QuObjects is an "s3 compliant" service on QNAP nas machine.
It can be configured as using tls, but it use a self-signed certificate. Restic can use it but requires "--insecure-tls" argument to use it. Yet k8up doesn't allow to pass that argument to restic.
It can also but used as http, but then restic use a minio compatibility mode which doesnt suit QuObjects and restic fail with "The Content-MD5 you specified was invalid".

### Logs

```shell
using tls:
1.6723221876801217e+09  INFO    k8up.restic.restic.RepoInit.restic.stderr       Fatal: create repository at s3:https://s3.rennes.home:8443/solidite failed: client.BucketExists: Get "https://s3.rennes.home:8443/solidite/?location=": x509: certificate is not valid for any names, but wanted to match s3.rennes.home

using plain http:
1.6723222329661973e+09  INFO    k8up.restic.restic.RepoInit.restic.stderr       Fatal: create key in repository at s3:http://s3.rennes.home:80/solidite failed: client.PutObject: The Content-MD5 you specified was invalid
```

### Expected Behavior

Using restic command-line it works :

export RESTIC_REPOSITORY="s3:https://s3.rennes.home:8443/solidite"
restic init --insecure-tls
created restic repository 493e88aa9c at s3:https://s3.rennes.home:8443/solidite
....

Please provide a way to propagate that --insecure-tls to restic

### Version of K8up

v2.5.2

### Version of Kubernetes

v1.25.3+k3s1

### Distribution of Kubernetes

k3s

ccremer:
Hi @sebt3 
Thanks for opening this issue and the interest and sorry for the late response.
I've allowed myself to change the title to a generic feature request where one can specify to ignore TLS verification. However, TLS verification being enabled by default is by design and the skipping verification is not a bug.

As of now, there's no possibility to backup to a Restic repository with a self-signed certificate. You might be able to backup using an alternative backend, e.g. non-TLS backend (`http`) or using the Restic REST server without TLS.

ccremer:
> It can also but used as http, but then restic use a minio compatibility mode which doesnt suit QuObjects and restic fail with "The Content-MD5 you specified was invalid".

Just extending up this. I'm not sure what exactly you mean with minio compatibility mode.
[...]

Issue #882, opened 2023-08-15
How to add insecureSkipTLSVerify=true for S3 bucket connection in K8up backup.yaml resource?


vistasunil:
### Description

Hello Team,

Provided [backup.yaml](https://github.com/k8up-io/k8up/blob/master/docs/modules/ROOT/examples/tutorial/backup.yaml) example file including K8up documentation, doesn't mentions anything about enabling insecureSkipTLSVerify=true flag for connecting to S3 buckets:

I am blocked due to this. Is there any way we can achieve the same?
If anyone knows, then kindly share the details.

Thanks & Regards,
Sunil Kumar

### Expected Behavior

Enabling insecure TLS while accessing S3 backup buckets should be allowed through K8up Kubernetes backup object.

### Version of K8up

2.7.1

### Version of Kubernetes

v1.24

### Distribution of Kubernetes

Rancher

Kidswiss:
Hi @vistasunil 

I'm closing this as a duplicate of https://github.com/k8up-io/k8up/issues/792.

PR #926, opened 2024-02-21
Enhancement: added --insecure-tls

eunames:
## Summary

* Short summary of what's included in the PR
* Give special note to breaking changes: List the exact changes or provide links to documentation.

### For Code changes

Added the oportunity to set a flag --insecure-tls for the restic command.
If you want to set the flag --insecure-tls you should set env SET_INSECURE_TLS_FLAG to true in deployment of k8up operator.

Example:
```
  template:
    spec:
      containers:
      - args:
        - operator
        env:
        - name: BACKUP_ENABLE_LEADER_ELECTION
          value: "true"
        - name: BACKUP_OPERATOR_NAMESPACE
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
        - name: SET_INSECURE_TLS_FLAG
          value: "true"
```
fixed: 
https://github.com/k8up-io/k8up/issues/792
https://github.com/k8up-io/k8up/issues/882
https://github.com/k8up-io/k8up/discussions/881


Kidswiss (changes requested):
Hi @eunames 

Thank you for the PR!

I feel like adding this as an operator level configuration is not very flexible. With this each and every backup either has it enabled or disabled.

This might be fine for a k8s cluster with only one tenant, however K8up is frequently used in a multi-tenant environment. Setting a security critical configuration globally is not a good idea in such a scenario, as it could compromise the security of the tenants.

So instead, it would be the more flexible approach to add a new flag in the backup CRD and then set `SET_INSECURE_TLS_FLAG` accordingly on the restic jobs. This way each tenant can configure the setting for each backup individually.

Can you please have a look at my suggestion? From what you already have it shouldn't be too hard to add it to the CRD and pass it to the restic jobs.

Kidswiss:
> A value TRUE in the operator level has priority over a value FALSE in the backup card

This feels somewhat like the wrong way around. All other operator level options have less precedence than the options in the backup object.

Also, are you still testing, or is this ready for a re-review from my side?

arska:
Thanks for taking the time to contribute this! We really appreciate it.

Since there hasn't been any activity on this PR for about 2 years, we're going to close it for now to keep things tidy. If this is still relevant, feel free to reopen or submit a new PR — we'd be happy to review it.

PR #938, opened 2024-03-16
Update object-specifications.adoc to fix spelling mistake

meierra:
passwordSecretReg to passwordSecretRef

## Summary

Original documentation has a spelling mistake sercretPasswordReg

### DIFF
--- a/docs/modules/ROOT/pages/references/object-specifications.adoc
+++ b/docs/modules/ROOT/pages/references/object-specifications.adoc
@@ -428,7 +428,7 @@ Settings:
 
 * `url`: URL of the Rest server instance (include scheme like `https://` on your own)
 * `userSecretRef`: Kubernetes secret reference containing the basic auth user
-* `passwordSecretReg`: Kubernetes secret reference containing the basic auth password
+* `passwordSecretRef`: Kubernetes secret reference containing the basic auth password
 
 == PreBackupPod
 


Kidswiss:
Hi @meierra 

Thanks for the contribution, could you please fix the DCO on the commit? Please check here how to do that: https://github.com/k8up-io/k8up/pull/938/checks?check_run_id=22742886371

Once that's fixed I can merge the change.

Thanks!

meierra:
Hi @Kidswiss,

I think you need to reject the change, I do think there is a spelling mistake, but it seems to be in the code itself as well, it seems that the code is written with sercretPasswordReg.. I'm not quite up to speed enough with go to truly understand, but I was having a lot of issues with setting up a rest connection and I thought it might be related to this documentation error.. seems that it's deeper than that though..

I do think that overall this does need to be fixed, but it's a little more than I can do quickly, I'm going to spend some time applying my mind to this and seeing if I can ultimately figure out the issue and also fix the code, but it will take a while for me to (a) get up to speed enough and (b) get a test environment going.... for now maybe you can just reject the commit?

Thanks,
Robert

Kidswiss:
Oh god, you're right, this typo has been in the CRD definitions for 3 years now :D

So changing this in the code would actually warrant a breaking change, as it renames the field in the CRD. And as long as the docs and the CRD say the same it's not too tragic. But nonetheless, we should probably fix it some time. That typo could really throw off people.

Issue #550, opened 2021-11-01
Provide opt-in for PVC backup


mstarostik:
### Summary

**As** an application operator
**I want** an option
**So that** I can default-disable PVC backups 

### Context

I've got some namespaces containing  few (or often just one) PVC that shall be included in a backup.
There are also various PVCs that contain data valuable enough to persist over pod restarts and migrations but not required to be backed up. Those volume contain e.g. cached data that is better recreated from its original source than restored from a backup in case of failure. Others hold data that is nice-to-keep for some time but not at all critical when lost.
The point is that a minority of very specific PVCs should be backed up, so I'd like to have a way to opt in to backup instead of having to opt out for all other volumes. 

### Acceptance Criteria

- Given an according configuration, only PVCs with an annotation of k8up.io/backup=true shall be backed up

### Implementation Ideas

k8up's CRDs' spec and or the operator's global configuration could gain an additional option to specifiy a default value for missing backup annotations.

Kidswiss:
Hi @mstarostik 

I'm fleshing out the ticket a bit more and edit your initial text to add some more information.

9numbernine9:
Hello! Just wondering if there has been any movement on this feature request? 

I have a similar situation as @mstarostik : I have a namespace that contains multiple PVCs inside, but only a few of them need to be backed up. Most of the PVCs (and the pods that attach them) are managed by an operator that (unfortunately) doesn't give a nice way to add the `k8up.io/backup: "false"` annotation to the pods or PVCs, so I need to manually `kubectl patch` the PVCs to add the annotation. This works _okay_ but if any new ones get added or the existing ones are destroyed and re-created, I need to re-apply the annotations again. 

It would be _really_ nice to configure k8up globally (or on a per-namespace basis) behaviour to require explicit _opt-in_ behaviour rather than opt-out.

rparree:
@9numbernine9 I am in a similar boat with crunchy PGO. I use this command:

```shell
$ kubectl  annotate pvc -lpostgres-operator.crunchydata.com/role=pgdata k8up.io/backup=false -A 
```

aleskandro:
Hi, I'd like this feature, and similarly to @rparree case, it might be worth considering opt-in by labels selectors IMO.


Moreover, by reading the AC, this case should be considered too:

```
Given configuration `pvc-opt-in=true`,
And a PVC `pvc2` with annotation `k8up.io/backup=false`,
And a backup spec with `include-pvc: pvc2`,
[...]

PR #948, opened 2024-03-19
Add option to ignore PVCs without annotation

Stogas:
## Summary

* Add CLI flag / environment variable to allow overriding the default behavior for PVCs without `k8up.io/backup` annotation
* Allow setting this variable through Helm chart's values
* Add relevant notes in documentation, where seems applicable

Relates to https://github.com/k8up-io/k8up/issues/550

### For Code changes

- [ ] Link this PR to related issues

### For Helm Chart changes


Stogas:
btw, I am unable to set any labels on this issue

Stogas:
I can see some chart tests are failing, partially because I haven't updated the chart tests to handle the new variable.

I'd love to help, but would love some guidance, as I don't have much experience with those.

Kidswiss:
Hi @Stogas 

Thanks for the contribution! Looks like a useful feature!

> I can see some chart tests are failing, partially because I haven't updated the chart tests to handle the new variable.
> 
> I'd love to help, but would love some guidance, as I don't have much experience with those.

The lint job is failing because the docs haven't been updated. You can run `make chart-docs` to fix that.

As for the other test: I'd need to dig into that one a bit as well. I haven't had those fail on me before. I think we just need to specify the new env var somewhere, so the test is aware about it.

Stogas:
I've added a fix for both, so both should succeed. However, I'm not entirely sure how to add and check for the value of the envVar in the generated manifest, so I added a test to only look for the envVar name in the right location.

Kidswiss:
Hmmm... still isn't quite happy. Btw you can use `make chart-test` you can run that test locally.

I can have a look later how to properly add your new envVar.

Stogas:
@Kidswiss yeah, I'm not sure why the fix didn't help.

Let me know if I can help somehow!

charts/k8up/test/deployment_test.go
```diff
@@ -37,8 +37,9 @@ func Test_Deployment_ShouldRender_EnvironmentVariables(t *testing.T) {
...
 	assert.Equal(t, "TZ", envs[1].Name)
 	assert.Equal(t, wantTimezone, envs[1].Value)
-	assert.Equal(t, "BACKUP_OPERATOR_NAMESPACE", envs[3].Name)
-	assert.Equal(t, "metadata.namespace", envs[3].ValueFrom.FieldRef.FieldPath)
+	assert.Equal(t, "BACKUP_SKIP_WITHOUT_ANNOTATION", envs[3].Name)
+	assert.Equal(t, "BACKUP_OPERATOR_NAMESPACE", envs[4].Name)
+	assert.Equal(t, "metadata.namespace", envs[4].ValueFrom.FieldRef.FieldPath)
 	assert.Equal(t, "BACKUP_GLOBALCPU_REQUEST", envs[4].Name, "Deployment does not use configured Env Name")
 	assert.Equal(t, wantCpuRequest, envs[4].Value, "Deployment does not use configured Env Value")
 	assert.Equal(t, "VARIABLE", envs[5].Name, "Deployment does not use configured Env Name")
```
Kidswiss, line 45: You forgot to move these values as well. The whole array shifts by one because of the new variable:
```suggestion
	assert.Equal(t, "BACKUP_GLOBALCPU_REQUEST", envs[5].Name, "Deployment does not use configured Env Name")
	assert.Equal(t, wantCpuRequest, envs[5].Value, "Deployment does not use configured Env Value")
	assert.Equal(t, "VARIABLE", envs[6].Name, "Deployment does not use configured Env Name")
	assert.Equal(t, "VALUE", envs[6].Value, "Deployment does not use configured Env Value")
```

Then the `make chart-test` runs through for me:
```

make chart-test
cd charts && go test ./...
ok      charts/k8up/test        1.554s
ok      charts/k8up/test/prometheus     (cached)
```

Stogas:
@Kidswiss fixed, requested re-review

Issue #900, opened 2023-10-18
Allow using custom CA certificates 


mpepping:
### Summary

**As** "an operator"\
**I want** "to use my custom CA bundle with K8up"\
**So that** "I can use a trusted, verified TLS connection to my backup backend"

### Context

To handle TLS connections that use a custom/self-signed CA as certificate issuer, we now create our own K8up container images that includes our own certificates. This allows us to have a verified connection to our S3 compatible backup backend.

We would prefer to have an option in the K8up container image itself to refer a custom CA bundle. This is supported in Restic via the `--cacert` flag. 

Effectively, we want to appoint a CA bundle file via an env.var or config file for the K8s Pod.

### Out of Scope

- skip TLS or insecure TLS (#792)

### Further links

- #568 comparable but stale 

### Implementation Ideas

- Referring a CA cert via the Object specification
- Golang's `SSL_CERT_FILE` variable

mpepping:
Just FYI; This is roughly how we build our a custom image now.

```
FROM registry.example.com/k8up-io/k8up:v2.7.1

USER root

RUN \
  apk add --no-cache openssl &&\
  /usr/bin/curl "http://pki.example.com/RootCACert.crt" |\
    /usr/bin/openssl x509 -inform DER -outform PEM -out /usr/local/share/ca-certificates/root.crt; \
  /usr/bin/curl "http://pki.example.com/intermediate.crt" |\
    /usr/bin/openssl x509 -inform DER -outform PEM -out /usr/local/share/ca-certificates/intermediate.crt; \
  /usr/bin/curl "http://pki.example.com/myissuer.crt" |\
    /usr/bin/openssl x509 -inform DER -outform PEM -out /usr/local/share/ca-certificates/myissuer.crt; \
  /usr/sbin/update-ca-certificates

USER 6553
```

poyaz:
Hi

We have the same problem with k8up and our s3 server runs as TLS with a self-sign certificate. We would like to use the feature for mounting certificate files. 

I am interested in developing this feature. I'd like to add these features:

1. First, add options for using `extraMount` options to add certificate files
2. Second, add env (or s3 parse URI) for TLS insecure

Reference:
#792

---

@Kidswiss @tobru

arska:
Custom CA certificate support is now available via the `tlsOptions.caCert` field in the backend spec, and is documented in the backup, restore, and archive how-to guides with examples for both self-signed and mTLS configurations.

See:
[...]

Issue #568, opened 2021-11-23
allow for a secret to be mounted as a TLS truststore


rparree:
### Context

I would be great to be able to inject TLS trust certificates  into the pod for the `Job ` template. In my situation the connection to s3/minio requires a trust certificate to be added to the container. For now i am going to create my own image based for the `backupImage` based on `ghcr.io/k8up-io/k8up`

ccremer:
Hi.
Thanks for you suggestion.
For the next time, please fill out the "Feature request" form on GH to the best of your knowledge.
I've edited the minimal scaffold for you now.

mpepping:
Quite relevant for my use case as well! Since this is an old issue @ccremer, is it feasible to open a new issue using the Feature request form?

ccremer:
Hi @mpepping 
I'm no longer associated with K8up. Maybe @Kidswiss can you help out here.

PR #949, opened 2024-03-23
[enhancement] Adding new feature for supporting self-signed certificate

poyaz:
## Summary

- This feature provides a supporting self-signed certificate and Mutual TLS (Related to #568, #900). These changes contain adding volume, volumeMount, and options section in the API resource. Also, you can mount volume and use environment variables to fill the required certificate.
- Adding option in operator and restic for using a dir for store temporary files (like generate client pem from **client.key** and **client.crt** according to `--tls-client-cert` option)
- Adding more e2e tests to make sure all things are okay (for adding self-signed feature)
- Fixing bug about checking errors in the s3 channel when an error is emitted in the channel (File: [restic/cli/restore.go](https://github.com/k8up-io/k8up/blob/v2.8.0/restic/cli/restore.go#L302-L320)). In this scenario after the error happens in s3 connection, the restore pod is frozen because the go routing never realize  about this error
- Complete document about how to use self-signed issue
- Adding variable `GO_EXEC` in Makefile to choose different versions of Golang (If the user installs multiply versions of Golang) 

### For Code changes

### For Helm Chart changes


e2e/lib/k8up.bash
```diff
@@ -69,13 +70,31 @@ restic() {
...
-		--repo "s3:http://minio.minio.svc.cluster.local:9000/backup" \
+		--repo "s3:http://minio.${MINIO_NAMESPACE}.svc.cluster.local:9000/backup" \
 		"${@}" \
 		--json > /dev/null
 	kubectl wait --for jsonpath='{.status.phase}'=Succeeded pod "$podname" -n "${DETIK_CLIENT_NAMESPACE-"k8up-system"}" --timeout=2m > /dev/null
 	kubectl -n "${DETIK_CLIENT_NAMESPACE-"k8up-system"}" logs "$podname"
 }
 
+mc() {
+	sleep 3
```
Kidswiss, line 81: Same here.

operator/archivecontroller/executor.go
```diff
@@ -67,19 +86,20 @@ func (a *ArchiveExecutor) jobName() string {
...
 }
 
-func (a *ArchiveExecutor) setupArgs(archive *k8upv1.Archive) []string {
-	args := []string{"-archive", "-restoreType", "s3"}
-
-	if archive.Spec.RestoreSpec != nil {
-		if len(archive.Spec.RestoreSpec.Tags) > 0 {
-			args = append(args, executor.BuildTagArgs(archive.Spec.RestoreSpec.Tags)...)
-		}
+func (a *ArchiveExecutor) setupArgs() ([]string, error) {
```
Kidswiss, line 89: The only return in this function returns a nil error. So it doesn't really need to return an error here.

cmd/restic/integration_test.go
```diff
@@ -59,9 +59,11 @@ func newTestErrorChannel() chan error {
 
 func (w *webhookserver) runWebServer(t *testing.T) {
 	mux := http.NewServeMux()
-	mux.HandleFunc("/", func(wr http.ResponseWriter, r *http.Request) {
-		w.jsonData, _ = io.ReadAll(r.Body)
-	})
+	mux.HandleFunc(
```
Kidswiss, line 62: Are you using some custom formatter? Is it possible to revert these formatting changes?

These format changes are a bit irritating and create a ton of changes in the diff. Sorry.

poyaz:
> Hi @poyaz
> 
> Thanks for this huge feature! Looks cool.
> 
> I haven't thoroughly reviewed it yet. Because unfortunately I have a problem with this PR: Your formatter did a lot of changes to the code. This makes reviewing this a lot harder. Also, your formatter style clashes with the formatting for the rest of the project.
> 
> Is it possible to adjust your formatter, so it conserves the old style?

Yeah, sorry for that
I will fix my formatter and revert to old style

e2e/lib/k8up.bash
```diff
@@ -56,6 +56,7 @@ clear_pv_data() {
 # It's very unreliable unfortunately. So running the pod, waiting and getting the
 # log output is a lot less prone for race conditions.
 restic() {
+	sleep 3
```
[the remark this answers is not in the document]
poyaz: I will remove it. It's because of failaier in getting log of container and you fix it in #944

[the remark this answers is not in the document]
poyaz: I forgot to remove sleep command after that change

api/v1/backend.go
```diff
@@ -279,3 +286,9 @@ func (in *RestServerSpec) String() string {
 	protocol, url, _ := strings.Cut(in.URL, "://")
 	return fmt.Sprintf("rest:%s://%s:%s@%s", protocol, "$(USER)", "$(PASSWORD)", url)
 }
+
+type BackendOpts struct {
```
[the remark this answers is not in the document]
poyaz: Actually it's maybe over engineering, I prefer use options property cuz maybe in feature k8up need some options for manage restic or anything.

It's like command optionsis  transfred to yaml. In feature we be able add more options for manage every thing

If you think it's unnecessary, I can change this property and rename to the TLSOptions or use ENV instead of yaml property

operator/archivecontroller/executor.go
poyaz: Actually I am using code style as like https://github.com/k8up-io/k8up/blob/1088118931d95579d76d4b0b668214f21d63d01d/operator/restorecontroller/executor.go#L90, maybe in feature we add more options with a use case and in this use case we check something generate a error.

In this case no eed change function definition and just add our condition.

cmd/restic/integration_test.go
poyaz: I'm sorry, I fix it and change my idea formatter

poyaz: Fixed

e2e/lib/k8up.bash
poyaz: Fixed

e2e/lib/k8up.bash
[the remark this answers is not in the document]
poyaz: Fixed

cmd/operator/main.go
```diff
@@ -83,6 +83,8 @@ var (
 			&cli.BoolFlag{Destination: &cfg.Config.SkipWithoutAnnotation, Name: "skip-pvcs-without-annotation", EnvVars: []string{"BACKUP_SKIP_WITHOUT_ANNOTATION"}, Value: false, DefaultText: "disabled", Usage: "skip selecting PVCs that don't have the BACKUP_ANNOTATION"},
 			&cli.StringFlag{Destination: &cfg.Config.BackupCheckSchedule, Name: "checkschedule", EnvVars: []string{"BACKUP_CHECKSCHEDULE"}, Value: "0 0 * * 0", Usage: "the default check schedule"},
 			&cli.StringFlag{Destination: &cfg.Config.OperatorNamespace, Name: "operator-namespace", EnvVars: []string{"BACKUP_OPERATOR_NAMESPACE"}, Required: true, Usage: "set the namespace in which the K8up operator itself runs"},
+
+			&cli.StringFlag{Destination: &cfg.Config.PodVarDir, Name: "vardir", EnvVars: []string{"VAR_DIR"}, Value: "/k8up", Usage: "the var data dir for read/write k8up data or temp file in pod"},
```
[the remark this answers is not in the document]
poyaz: Fixed

docs/modules/ROOT/pages/how-tos/backup.adoc
```diff
@@ -12,3 +12,151 @@ Save the YAML above in a file named `backup.yaml` and use the `kubectl apply -f
 TIP: To have backups run automatically at a regular interval look at xref:how-tos/schedules.adoc[schedules].
 
 TIP: By default, all PVCs are backed up automatically. Adding the annotation `k8up.io/backup=false` to a PVC object will exclude it from all following backups. Alternatively, you can set the environment variable `BACKUP_SKIP_WITHOUT_ANNOTATION=true` if you want K8up to ignore objects without the annotation.
+
+== Self-signed issuer and Mutual TLS
+
+If you are using self-signed issuer or using mutual tls for authenticate client, you be able to using volume for mounting cert files into backup object.
```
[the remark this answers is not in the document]
poyaz: Fixed

api/v1/backend.go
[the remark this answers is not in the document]
Kidswiss: I'd say generic restic options should belong in a dedicated property outside the `Backend` object. As that would mix up the responsibilities of the `Backend` object.

operator/archivecontroller/executor.go
Kidswiss: If we'd ever add something more in the future that might return an error, we can always adjust the function signature then.

I'd like to avoid having non-functional code lines, as it will only confuse people.

cmd/restic/integration_test.go
Kidswiss: Thanks!

I'll try to have another in-depth look at the code this week.

api/v1/backend.go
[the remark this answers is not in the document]
poyaz: How would you like to desire this property? Which is better, renaming to `TLSOptions` or moving outside the `Backend` object?

operator/archivecontroller/executor.go
poyaz: Okay, You're right
I will change it

api/v1/backend.go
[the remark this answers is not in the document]
Kidswiss: Just rename it for now.

api/v1/restore_types.go
```diff
@@ -145,3 +147,9 @@ func init() {
 var (
 	RestoreKind = reflect.TypeOf(Restore{}).Name()
 )
+
+type RestoreOpts struct {
```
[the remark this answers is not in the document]
poyaz: fixed

api/v1/backend.go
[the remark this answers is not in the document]
poyaz: fixed

operator/archivecontroller/executor.go
poyaz: fixed

cmd/restic/integration_test.go
poyaz: fixed

e2e/test-10-self-signed-tls.bats
```diff
@@ -0,0 +1,580 @@
...
+# shellcheck disable=SC2034
+DETIK_CLIENT_NAME="kubectl"
+# shellcheck disable=SC2034
+DETIK_CLIENT_NAMESPACE="k8up-e2e-subject"
+# shellcheck disable=SC2034
+DEBUG_DETIK="true"
+
+### Start backup section
+
+@test "Given a PVC, When creating a Backup (TLS) of an app, Then expect Restic repository - using self-signed issuer" {
```
Kidswiss, line 16: Can you reduce this to 1 TLS and 1 mTLS test? This file contains more end-to-end tests than the rest of k8up combined and takes almost 1h to run. This is pretty hard to maintain.

E2e tests are usually not used to test each permutation of every possible scenario. That's something for unit and integration tests. E2e tests should be few and pretty broad. (Yes I know that there aren't that much of those either in K8up...)

operator/checkcontroller/executor.go
```diff
@@ -93,3 +108,77 @@ func (c *CheckExecutor) setupEnvVars(ctx context.Context) []corev1.EnvVar {
 func (c *CheckExecutor) cleanupOldChecks(ctx context.Context, check *k8upv1.Check) {
 	c.CleanupOldResources(ctx, &k8upv1.CheckList{}, check.Namespace, check)
 }
+
+func (c *CheckExecutor) appendTLSOptionsArgs() []string {
```
Kidswiss, line 112: You can generalize this function a bit. Instead of making it a method of the *Executor objects you could have something like:
```go
func AppendTLSOptionsArgs(opts *k8upv1.TLSOptions) []string
```

And put it in the new `utils` package. That will reduce a lot of code duplication.

operator/checkcontroller/executor.go
```diff
@@ -93,3 +108,77 @@ func (c *CheckExecutor) setupEnvVars(ctx context.Context) []corev1.EnvVar {
...
+			"-clientKey",
+			c.check.Spec.Backend.TLSOptions.ClientKey,
+		}
+		args = append(args, addMoreArgs...)
+	}
+
+	return args
+}
+
+func (c *CheckExecutor) attachMoreVolumes() []corev1.Volume {
```
Kidswiss, line 134: Same here.

Kidswiss (changes requested):
Just a few more things, then it should be ready to merge.

e2e/test-10-self-signed-tls.bats
Kidswiss: I'd probably go with something like `test-04-restore.bats` does. A backup and a restore in one single e2e test. (IMHO test-04-restore.bats probably makes the test-03-backup.bats test more or less redundant anyway. But that's a problem for another issue :) )

operator/checkcontroller/executor.go
poyaz: I removed duplicate code. I put all of them in a utils file and wrote tests for them

operator/checkcontroller/executor.go
poyaz: Fiexd

e2e/test-10-self-signed-tls.bats
poyaz: I removed the multiply e2e tests and created an integration for them.

Kidswiss (approved):
LGTM now!

Thanks for the contribution!

Issue #978, opened 2024-06-04
Backup of RWO volumes used by Pods fails


johbo:
### Description

I have a stuck backup job and I think this is because it tries to backup a RWO volume in the wrong Node.

On the backup Pod I see the following Event:

```
Multi-Attach error for volume "pvc-0df4196b-7261-425d-bee0-0e74d466a216" Volume is already used by pod(s) gitlab-postgresql-0   
```

The backup Pod is scheduled on my node `test-v2-01`, the Pod which has the volume mounted is running on Node `test-v2-02`.

### Additional Context

The volume is based on Rook Ceph and a volume in RWO mode.

I started to see the issue after switching over from plain local volumes. So far I only saw this during a bootstrap from scratch, have to observe how it behaves over time in the running cluster.

### Expected Behavior

The backup Pod should be scheduled on the correct Pod so that it can mount the Volume and perform the backup.

I am not sure if there could be a race condition, e.g. k8up decided where to put the backup Pod, the other Pod which uses the Volume starting right after this and k8up then using the outdated data.

### Steps To Reproduce

Did not yet find out how to reproduce this reliably. So far I see this problem after a bootstrap of my setup from scratch, this means fresh machines, fresh Kubernetes and then a fresh deployment of everything based on Flux. 

### Version of K8up

v2.10.0

### Version of Kubernetes

v1.30.1+k0s

### Distribution of Kubernetes

k0s

Kidswiss:
Hi @johbo 

Thanks for opening the issue.

Is it possible that the pod was restarted right after the backup started? K8up determines what pod is running on what node before creating the jobs. If the pod gets restarted and scheduled to a new node, it can happen that this doesn't match anymore.

johbo:
Yes, I think this could be the cause of the issue. I've not yet had a chance to look into the k8up code. 

Based on your hint I think this could explain it though. During bootstrap I have a few things happening in a sequence which could lead to this problem if I understand you correctly:

1. Create the volume (Deploy PVC)
2. Restore data into the Volume via a Job
3. Wait for the restore Job to complete
4. Deploy the application

Now I think that the Schedule object is deployed already early into the namespace. K8up could then trigger a backup based on the restore Job's Pod and once the application is deployed it could end up in the situation that the application's Pod is on a different Node.

I think for bootstrapping my setup I can workaround this by delaying the deployment of the backup Schedule. 

[...]

PR #979, opened 2024-06-07
Ignore terminated Pods in mapping of PVCs to Pods

johbo:
## Summary

Think that terminated Pods should be ignored when building the mapping from PVC to Pod.

Issue: https://github.com/k8up-io/k8up/issues/978

### For Code changes

### For Helm Chart changes

- [ ] Commits are [signed off](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

### DIFF
--- a/operator/backupcontroller/executor.go
+++ b/operator/backupcontroller/executor.go
@@ -69,8 +69,8 @@ func (b *BackupExecutor) listAndFilterPVCs(ctx context.Context, annotation strin
 		return nil, fmt.Errorf("list pods: %w", err)
 	}
 	for _, pod := range pods.Items {
-		if pod.Status.Phase == corev1.PodSucceeded || pod.Status.Phase == corev1.PodFailed {
-			log.V(1).Info("Ignoring terminated Pod", "pod", pod.GetName())
+		if pod.Status.Phase != corev1.PodRunning {
+			log.V(1).Info("Ignoring Pod which is not running", "pod", pod.GetName())
 			continue
 		}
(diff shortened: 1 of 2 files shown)


operator/backupcontroller/executor.go
```diff
@@ -69,6 +69,10 @@ func (b *BackupExecutor) listAndFilterPVCs(ctx context.Context, annotation strin
 		return nil, fmt.Errorf("list pods: %w", err)
 	}
 	for _, pod := range pods.Items {
+		if pod.Status.Phase == corev1.PodSucceeded || pod.Status.Phase == corev1.PodFailed {
```
Kidswiss, line 72: Maybe instead of not considdering Succedded and Failed pods, we could switch the logic arround:
only consider pods that are running.

This way we also ignore pods in Pending or Terminating states, which both could also lead to lock ups of jobs.

Kidswiss (changes requested):
Hi @johbo 

Thanks for the fix! I also think it makes sense to ignore non-running pods.

Can you please fix the DCO signature in your commits? See here how to do that: https://github.com/k8up-io/k8up/pull/979/checks?check_run_id=25919302156

operator/backupcontroller/executor.go
johbo: Yes, that's a good point, was focused so far only on the issue right in front of me 😃 

Will do a tweak for this.

Kidswiss (approved):

Kidswiss:
LGTM!

Thanks for the contribution!

Issue #981, opened 2024-06-22
[controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.


jessebot:
### Description

I always get the following in the logs of any backup pod: `[controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.` The backup pods are always successful and the data is actually backed up, but they always print this trackback and I don't know why? Nothing is broken, just a bit confusing.

### Additional Context

I'm using s3 as my backend and Backblaze B2 as the remote s3 provider. All is working there fine. Again, none of this _breaks_ anything. It just causes weird log messages that I don't understand. *Update*: I switched to Cloudflare R2 and the issue persists.

Thanks for any help you can provide and thanks for continuing to maintain this project!

### Logs

```shell
2024-06-22T08:34:49Z    INFO    k8up.restic.restic.backup.progress      backup finished {"new files": 0, "changed files": 56, "errors": 0}
2024-06-22T08:34:49Z    INFO    k8up.restic.restic.backup.progress      stats   {"time": 248.825449061, "bytes added": 2842538507, "bytes processed": 14227999250}
2024-06-22T08:34:49Z    INFO    k8up.restic.statsHandler.promStats      sending prometheus stats        {"url": "push-gateway.prometheus.svc:9091"}
2024-06-22T08:34:49Z    INFO    k8up.restic.restic.backup       backup finished, sending snapshot list
2024-06-22T08:34:49Z    INFO    k8up.restic.restic.snapshots    getting list of snapshots
2024-06-22T08:34:49Z    INFO    k8up.restic.restic.snapshots.command    restic command  {"path": "/usr/local/bin/restic", "args": ["snapshots", "--option", "", "--json"]}
2024-06-22T08:34:49Z    INFO    k8up.restic.restic.snapshots.command    Defining RESTIC_PROGRESS_FPS    {"frequency": 0.016666666666666666}
[controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.
Detected at:
        >  goroutine 1 [running]:
        >  runtime/debug.Stack()
        >       /opt/hostedtoolcache/go/1.21.9/x64/src/runtime/debug/stack.go:24 +0x5e
        >  sigs.k8s.io/controller-runtime/pkg/log.eventuallyFulfillRoot()
        >       /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.2/pkg/log/log.go:60 +0xcd
[...]

PR #988, opened 2024-07-24
Properly set logger before creating a typed client

Kidswiss:
## Summary

During pre-backup operations it was possible that the typed client was initialized without setting the logger first. Which lead to a big, but harmless, stacktrace in the backup pods.

This commit makes sure that the logger will be set every time the client gets created.

Resolves: #981

### For Code changes


zugao (approved):
LGTM

Issue #910, opened 2023-11-19
Failed backup stated as succeeded


xoxys:
### Description

Backups are stated as `Succeeded` even if the backup commands failed. This is just an example, the reason _why_ the backup failed has already been fixed. However, it is problematic to describe a defective backup as successful.

### Logs

```shell
❯ kubectl get backups.k8up.io -n authelia-public 
NAME                    SCHEDULE REF   COMPLETION   PREBACKUP   AGE
postgres                               Succeeded    Finished    12h
postgres-backup-qbq6q   postgres       Succeeded    Finished    12h
```

```
2023-11-18T23:02:31Z	INFO	k8up	Starting k8up…	{"version": "2.7.2", "date": "2023-10-09T10:13:29Z", "commit": "45d99dd90dbb2a080e6832c34e96b371216a3e0b", "go_os": "linux", "go_arch": "amd64", "go_version": "go1.19.13", "uid": 65532, "gid": 0}
2023-11-18T23:02:31Z	INFO	k8up.restic	initializing
2023-11-18T23:02:31Z	INFO	k8up.restic	setting up a signal handler
2023-11-18T23:02:31Z	INFO	k8up.restic.restic	using the following restic options	{"options": [""]}
2023-11-18T23:02:31Z	INFO	k8up.restic.restic.RepoInit.command	restic command	{"path": "/usr/local/bin/restic", "args": ["init", "--option", ""]}
2023-11-18T23:02:31Z	INFO	k8up.restic.restic.RepoInit.command	Defining RESTIC_PROGRESS_FPS	{"frequency": 0.016666666666666666}
2023-11-18T23:02:32Z	INFO	k8up.restic.restic.unlock	unlocking repository	{"all": false}
2023-11-18T23:02:32Z	INFO	k8up.restic.restic.unlock.command	restic command	{"path": "/usr/local/bin/restic", "args": ["unlock", "--option", ""]}
2023-11-18T23:02:32Z	INFO	k8up.restic.restic.unlock.command	Defining RESTIC_PROGRESS_FPS	{"frequency": 0.016666666666666666}
2023-11-18T23:02:36Z	INFO	k8up.restic.restic.snapshots	getting list of snapshots
2023-11-18T23:02:36Z	INFO	k8up.restic.restic.snapshots.command	restic command	{"path": "/usr/local/bin/restic", "args": ["snapshots", "--option", "", "--json"]}
[...]

PR #1027, opened 2024-12-05
Fail if the backup command in the annotation exits != 0

Kidswiss:
## Summary

This will fix issues were problems with pre-backup go undected. K8up will now completely fail the whole whole prebackup pod if the commands exist with a code != 0.
    
Also a small issue with concurrent `restic init` was fixed.
Fixes: #910

### For Code changes


operator/backupcontroller/executor.go
```diff
@@ -262,16 +263,17 @@ func (b *BackupExecutor) startBackup(ctx context.Context) error {
 			}
 			// each job sleeps for index seconds to avoid concurrent restic repository creation. Not the prettiest way but it works and a repository
 			// is created only once usually.
-			if index > 0 {
+			if name == "prebackup" || index != 0 {
```
Kidswiss, line 266: This fixes another small but unrelated issue.

If there's a PVC backup AND prebackup commands, it can happen that both want to `restic init` at the same time, which will corrupt the repo.

restic/kubernetes/pod_exec.go
```diff
@@ -73,6 +74,8 @@ func PodExec(pod BackupPod, log logr.Logger) (*ExecData, error) {
 
 		if err != nil {
 			execLogger.Error(err, "streaming data failed", "namespace", pod.Namespace, "pod", pod.PodName)
+			// we just completely hard fail the whole backup pod
+			os.Exit(1)
```
wejdross, line 78: why don't You simply return err to some new resChannel and then return it in the end of a function? os.Ecit(1) sounds bit harsh

wejdross (approved):

PR #1080, opened 2025-08-26
feat: documentation structure attribution

janikvonrotz:
Maybe I am mistaken, but the documentation structure is not from Divio, but from [Diátaxis](https://diataxis.fr/).

Let's give credits where they are due.

I am a big fan of this framework and it helped me a lot.


bastjan:
https://github.com/evildmp/diataxis-documentation-framework/issues/14#issuecomment-1170838014

Divio seems to have been around first according to the creator of Diátaxis.

I can't really judge anymore where the original maintainers found the inspiration so I'm going to close this PR.

Issue #862, opened 2023-05-25
Pre-Backup pod labels are too long


SchoolGuy:
### Description

The label of the pre-backup pod is too long. I have no clever way how to solve this but this should definitely not happen.

```yml
apiVersion: k8up.io/v1
kind: Schedule
metadata:
  name: backup-schedule-schoolguy-photoprism
  namespace: photoprism
spec:
  podSecurityContext:
    runAsUser: 0
  backend:
    rest:
      url: "http://<IP>:8000/<repository_name>"
    repoPasswordSecretRef:
      name: backup-repo
      key: password
  backup:
    schedule: '0 3 * * *'
    failedJobsHistoryLimit: 2
    successfulJobsHistoryLimit: 2
    promURL: http://prometheus-pushgateway.monitoring.svc.cluster.local:9091
  check:
    schedule: '0 5 * * 6'
    promURL: http://prometheus-pushgateway.monitoring.svc.cluster.local:9091
  prune:
    schedule: '0 0 1 * *'
    retention:
      keepLast: 5
      keepDaily: 14
```

### Logs

```shell
2023-05-25T19:35:44Z ERROR k8up.operator Reconciler error {"controller": "backup.k8up.io", "controllerGroup": "k8up.io", "controllerKind": "Backup", "Backup": {"name":"backup-schedule-schoolguy-photoprism-backup-j7s5z","namespace":"photoprism"}, "namespace": "photoprism", "name": "backup-schedule-schoolguy-photoprism-backup-j7s5z", "reconcileID": "313f0f45-4162-4af3-a009-584173651199", "error": "unable to createOrUpdate(\"backup-backup-schedule-schoolguy-photoprism-backup-j7s5z-prebackup\"): Job.batch \"backup-backup-schedule-schoolguy-photoprism-backup-j7s5z-prebackup\" is invalid: spec.template.labels: Invalid value: \"backup-backup-schedule-schoolguy-photoprism-backup-j7s5z-prebackup\": must be no more than 63 characters"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
 /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.1/pkg/internal/controller/controller.go:326
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
[...]

PR #1092, opened 2025-10-27
bug: truncate volume names to 63 characters in BackupController executor

SyeKlu:
## Summary

This PR fixes volume name handling in the BackupController executor to comply with Kubernetes’ 63-character limit.
volumes[*].name is now automatically truncated to 63 characters, preventing errors caused by longer names.
Other related fields, such as persistentVolumeClaim.claimName and configMap.name, remain unaffected.

Related Issue:
[Fixes k8up-io/k8up#862](https://github.com/k8up-io/k8up/issues/862)

### DIFF
--- a/operator/backupcontroller/backup_utils_test.go
+++ b/operator/backupcontroller/backup_utils_test.go
@@ -2,6 +2,7 @@ package backupcontroller
 
 import (
 	"testing"
+	"strings"
 
 	v1 "github.com/k8up-io/k8up/v2/api/v1"
 	"github.com/k8up-io/k8up/v2/operator/cfg"
@@ -76,6 +77,31 @@ func TestBackupExecutor_setupEnvVars(t *testing.T) {
 			for _, expectedEnv := range tt.expectedEnvVars {
 				assert.Contains(t, result, expectedEnv)
 			}
+		})
+	}
+}
+
+func TestTruncateVolumeName(t *testing.T) {
+	tests := map[string]struct {
+		input    string
+		expected string
+	}{
+		"short name": {
+			input:    "my-pvc",
+			expected: "my-pvc",
+		},
+		"exactly 63 chars": {
+			input:    strings.Repeat("a", 63),
+			expected: strings.Repeat("a", 63),
+		},
+		"over 63 chars": {
+			input:    strings.Repeat("a", 100),
+			expected: strings.Repeat("a", 63),
+		},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			assert.Equal(t, tc.expected, truncateVolumeName(tc.input))
 		})
 	}
 }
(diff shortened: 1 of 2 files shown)


aw-talos:
Hello. I am also affected by this. Will it be fixed soon?

arska:
Thanks for this contribution! We appreciate you taking the time to submit this PR.

To be able to merge it, we need the commits to be signed off (DCO — Developer Certificate of Origin), which is a requirement for all CNCF projects. You can do this by rebasing with:

```
git rebase --signoff HEAD~N
```

(where N is the number of commits in this PR)

and then force-pushing to your branch. If you have any questions about this, feel free to ask!

SyeKlu:
The commit is now signed with DCO and ready.

arska (commented):
Thanks for the update and the sign-off!

The fix is correct but I'd suggest two improvements:

1. **Extract a helper function** instead of the inline anonymous function — it's more readable and testable:

```go
// truncateVolumeName ensures the volume name doesn't exceed the Kubernetes 63-character limit.
func truncateVolumeName(name string) string {
	if len(name) > 63 {
		return name[:63]
	}
	return name
}
```

Then use it as:
```go
Name: truncateVolumeName(pvc.Name),
```

2. **Add a unit test** to prevent regressions:

```go
func TestTruncateVolumeName(t *testing.T) {
	tests := map[string]struct {
		input    string
		expected string
	}{
		"short name": {
			input:    "my-pvc",
			expected: "my-pvc",
		},
		"exactly 63 chars": {
			input:    strings.Repeat("a", 63),
			expected: strings.Repeat("a", 63),
		},
		"over 63 chars": {
			input:    strings.Repeat("a", 100),
			expected: strings.Repeat("a", 63),
		},
	}
	for name, tc := range tests {
		t.Run(name, func(t *testing.T) {
			assert.Equal(t, tc.expected, truncateVolumeName(tc.input))
		})
	}
}
```

Would you be able to make these changes? Happy to help if needed.

SyeKlu:
I've adjusted it

arska (approved):
Looks great — clean helper function with good test coverage. Thank you for the update!

arska:
CI is failing — missing `strings` import in the test file:

```
vet: operator/backupcontroller/backup_utils_test.go:93:14: undefined: strings
```

Please add `"strings"` to the imports in `backup_utils_test.go` and push.

SyeKlu:
I’ve added the import.

Kidswiss:
Hi @SyeKlu 

Sorry for the late reply and thanks for the contribution!

There's a slight issue with possible collisions with your solution.

Given:
verylongpvcname-1
verylongpvcname-2

Let's assume the are both 65 chars long. So they would lose the appendix, resulting in the same string:

verylongpvcname

Which will also result in an error. So I'd suggest to first hash the whole name, shorten in to something lower than 64 (e.g. 58 or so), then add a `-` and the first 5 characters of the hash.

This will ensure that the truncation will not create a collision, well at least if there was no collision to begin with.

PR #1030, opened 2024-12-09
feat: add clusterName label for identify each kubernetes cluster where K8UP is deployed

lenglet-k:
## Summary

We use K8up on multiple clusters. These multiple clusters perform backups of namespaces with the same name. All the metrics are sent in the same pushgateway. 

As the metrics have the same labels, we cannot have a detailed overview by cluster. This is the way K8UP currently works, which does not allow metric labels to be overloaded. This PR makes it possible to add a clusterName label per Kubernetes cluster. 

So in pushgateway we have backup metrics per Kubernetes cluster and per namespace, which improves monitoring of the execution of backups per Kubernetes cluster.

I essentially added a `clusterName` variable equivalent to the `promUrl` variable.

See: https://github.com/k8up-io/k8up/issues/1031

### For Code changes

- [ ] Categorize the PR by setting a good title and adding one of the labels:
  `bug`, `enhancement`, `documentation`, `change`, `breaking`, `dependency`
  as they show up in the changelog
- [ ] PR contains the label `area:operator`
- [x] Commits are [signed off](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)
- [x] Link this PR to related issues
- [ ] I have not made _any_ changes in the `charts/` directory.

PR #1097, opened 2025-12-02
Fix cluster label in backup metrics

eshutov:
## Summary

This PR fixes an issue with the Prometheus metrics sent from the restic backup pod to the pushgateway, specifically concerning the "cluster" label.

The "cluster" label was introduced in #1030. However, this label doesn't currently works as intended. The clusterName values from the backup/schedule manifest are not being passed to the backup pod's environment. As a result, the "cluster" label consistently appears empty ("") in pushgateway/prometheus.

Additionally, there is another point to discuss.
These two PRs #1030 and #260 added ClusterName and PromURL to "restic check" (./api/v1/check_types.go)
 code, which are intended for the code that pushes metrics to Pushgateway. It appears the actual code implementing this push functionality is missing, At least, I was unable to find it. So the "check" job doesn't push any metrics.

A reasonable solution might be to completely remove these configurations and clean the CRDs accordingly. But since these would be breaking changes, I have left the existing code unchanged, assuming that the metrics-pushing code either exists elsewhere or will be implemented in the future.

### For Code changes

### DIFF
--- a/operator/backupcontroller/backup_utils.go
+++ b/operator/backupcontroller/backup_utils.go
@@ -189,6 +189,7 @@ func (b *BackupExecutor) setupEnvVars() ([]corev1.EnvVar, error) {
 
 	vars.SetStringOrDefault("STATS_URL", b.backup.Spec.StatsURL, cfg.Config.GlobalStatsURL)
 	vars.SetStringOrDefault("PROM_URL", b.backup.Spec.PromURL, cfg.Config.PromURL)
+	vars.SetStringOrDefault("CLUSTER_NAME", b.backup.Spec.ClusterName, cfg.Config.ClusterName)
 	vars.SetString("BACKUPCOMMAND_ANNOTATION", cfg.Config.BackupCommandAnnotation)
 	vars.SetString("FILEEXTENSION_ANNOTATION", cfg.Config.FileExtensionAnnotation)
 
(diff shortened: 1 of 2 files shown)


Kidswiss:
Hi @eshutov 

Thank you for the PR!

> A reasonable solution might be to completely remove these configurations and clean the CRDs accordingly.

I agree, it might make sense to remove these. Not sure how helpful stats for the `check` command really are. Especially since most of its functionality is already [covered by prune](https://forum.restic.net/t/should-i-run-restic-check/1227/5).

I'd suggest to open an issue to discuss this further.

Did you run the e2e tests and verify that the metrics are now sent properly? For some reason Github doesn't trigger the actions properly on this PR...

If they are okay, I'll merge the change.

eshutov:
@Kidswiss I haven't actually run the built-in e2e tests. But i've been running this patch in my production and development environments (across 7 clusters total) for about five weeks. I'm using K8up as a backup tool for Kubernetes' etcd, which can be deployed in a native Kubernetes way and is integrated into my basic Ansible deployment configuration for K8s.

As you can see in the screenshot:
- The `k8up_backup_restic_available_snapshots` metric initially appeared without the "cluster" label (the aggregation shows "{}").
- Then I applied this patch and tested it on a single cluster, and the cluster label appeared. (The old metric without "cluster" label, showing "{}", still existed for a while because I forgot to remove it from the Pushgateway during testing.)
- Finally, I rolled it out to the entire infrastructure.

Screenshot 2026-01-13 at 14-35-47 Prometheus Time Series Collection and Processing Server

If it's necessary to test this with the built-in tests, I can do that, though it will take some time, likely a few weeks.

Kidswiss:
@eshutov Looks good!

I'll run the e2e tests locally, since Github is somehow stuck with the actions on the PRs... I don't expect any big surprises though.

Once they pass I'll force merge your PR.

Kidswiss:
Looks good. The tests pass.

PR #1104, opened 2026-02-13
Introduce relaxed backup job scheduling feature

anothertobi:
## Summary

The Kubernetes scheduler ensures that pods are scheduled to nodes where
a volume can be mounted for RWX and RWO volumes as long as the volumes
expose relevant topology information. Under optimal circumstances running
on the same node is not required to back up RWX volumes or to run backup
commands. The relaxed scheduling feature allows for backups to happen even
when a node is at the pod limit, volume attachment limit or being removed
while the backup job is being created.

The feature is opt-in to not break any backups relying on the existing
scheduling. This could be the case if for example a volume advertised as
RWX is actually RWO on the infrastructure layer.

### For Code changes

- [ ] Link this PR to related issues


anothertobi:
I changed the approach of the initial draft PR and made this an opt-in feature after discovering through the e2e tests that some storage providers don't set up their PVs with the required node affinity for this to work.

Issue #1055, opened 2025-04-10
Initial rest backup fails cause race condition leads to damaged repository key


JcGKitten:
### Description

Hey, when creating a new backup schedule, the initialization of the repository in the rest-server fails. I guess that's caused by a race condition mentioned [here](https://github.com/k8up-io/k8up/issues/861#issuecomment-2185114278) and #970.

### Additional Context

When initializing the repository manually backup works as expected.
K8Up was installed via helm. (4.8.4)

### Logs

```shell
logs prebackup pod:

2025-04-10T13:23:21Z    INFO    k8up    Starting k8up…  {"version": "2.12.0", "date": "2025-03-04T09:30:29Z", "commit": "0e10a932db96ef91d6fb2b69bbb16439e2fcdb73", "go_os": "linux", "go_arch": "amd64", "go_version": "go1.23.6", "uid": 65532, "gid": 0}
2025-04-10T13:23:21Z    INFO    k8up.restic     initializing
2025-04-10T13:23:21Z    INFO    k8up.restic     setting up a signal handler
2025-04-10T13:23:21Z    INFO    k8up.restic     sleeping until init     {"duration": "3s"}
2025-04-10T13:23:24Z    INFO    k8up.restic.restic.RepoInit.command     restic command  {"path": "/usr/local/bin/restic", "args": ["init"]}
2025-04-10T13:23:24Z    INFO    k8up.restic.restic.RepoInit.command     Defining RESTIC_PROGRESS_FPS    {"frequency": 0.016666666666666666}
2025-04-10T13:23:24Z    INFO    k8up.restic.restic.unlock       unlocking repository    {"all": false}
2025-04-10T13:23:24Z    INFO    k8up.restic.restic.unlock.command       restic command  {"path": "/usr/local/bin/restic", "args": ["unlock"]}
2025-04-10T13:23:24Z    INFO    k8up.restic.restic.unlock.command       Defining RESTIC_PROGRESS_FPS    {"frequency": 0.016666666666666666}
2025-04-10T13:23:25Z    INFO    k8up.restic.restic.unlock.restic.stderr Fatal: config or key 714e2d6ef4573d7d498f8d0db2f4e34549b113196ec996ca742d0e1acebc0a6e is damaged: ciphertext verification failed
2025-04-10T13:23:25Z    ERROR   k8up.restic     failed to remove stale locks from the repository, continuing anyway     {"error": "cmd.Wait() err: 1"}
[...]

PR #1105, opened 2026-02-20
Improve sleep for reliable repository creation

anothertobi:
## Summary

The SLEEP_DURATION was always set to 3 seconds which defeated the
purpose of the option. By setting it to index * 10 seconds the gap
between can be noticed and should improve the reliability of the
restic repository creation.

### For Code changes

- n/a: Link this PR to related issues

### DIFF
--- a/operator/backupcontroller/executor.go
+++ b/operator/backupcontroller/executor.go
@@ -301,12 +301,13 @@ func (b *BackupExecutor) startBackup(ctx context.Context) error {
 					Value: "true",
 				})
 			}
-			// each job sleeps for index seconds to avoid concurrent restic repository creation. Not the prettiest way but it works and a repository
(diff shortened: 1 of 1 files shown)


Kidswiss (approved):
Hi @anothertobi 

Long time no see :)

Thanks for the contribution! This sleep has been a bit of a thorn in my side for some time... I'd like to get rid of it at some time.

But for the time being, this is a good fix for the bandaid 

EDIT: re-triggered the E2E test. Maybe there's a race condition which gets amplified with this change...

Kidswiss:
This might not fix #1055 entirely, but should mitigate it a bit until we come up with a better idea.

Kidswiss:
thanks @anothertobi !

PR #1181, opened 2026-03-23
Only run govulncheck when Go dependencies change

arska:
## Summary

Change govulncheck trigger from `paths-ignore` (runs on almost everything) to `paths` (only `go.mod`/`go.sum`). Vulnerability status only changes when:
- Dependencies change → triggered by `go.mod`/`go.sum` paths
- New CVEs are disclosed → caught by weekly schedule (unchanged)

Code-only changes don't affect vulnerability status.

## Test plan

- [ ] Verify govulncheck runs on dependency PRs
- [ ] Verify govulncheck does NOT run on code-only PRs
- [ ] Verify weekly schedule still works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -2,15 +2,15 @@ name: Go Vulnerability Check
 
 on:
   pull_request:
-    paths-ignore:
-      - 'docs/**'
-      - 'charts/**'
+    paths:
+      - '**/go.mod'
+      - '**/go.sum'
   push:
     branches:
       - master
-    paths-ignore:
-      - 'docs/**'
-      - 'charts/**'
+    paths:
+      - '**/go.mod'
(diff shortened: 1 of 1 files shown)


anothertobi:
Limiting `govulncheck` to changes in `go.mod` and `go.sum` seems obvious, but actually defeats the purpose of the tool. To quote from the docs (https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck):
> Govulncheck reports known vulnerabilities that affect Go code. It **uses static analysis of source code** or a binary's symbol table to narrow down reports to only those that could affect the application.

It should therefore always run when any part of the code or the dependencies change to accurately report if something unsafe is introduced. This could be the case when a code path is added that makes use of a vulnerable part of a dependency.

PR #1203, opened 2026-04-15
On schedule creation, create labeled metrics with value 0

### LINKED ISSUE #976: Gaps in K8up metrics
### Description

Currently, K8up only emits timeseries for schedules for which it's seen at least one Job with the matching completion state, e.g. `k8up_jobs_successful_counter` will only have timeseries for schedules which have at least one successful job since the last K8up restart.
 
For schedules with relatively low frequency (e.g. 1/day) this can lead to significant gaps in the metric in Prometheus which confuses Prometheus functions such as `rate()` which otherwise can compensate for counter resets due to pod restarts.

### Additional Context

_No response_

### Logs

_No response_

### Expected Behavior

K8up initializes the counter metrics (`k8up_jobs_failed_counter`, `k8up_jobs_successful_counter`, and `k8up_jobs_total`) with value 0 for all job types and all namespaces in which a Schedule exists immediately after startup.

### Steps To Reproduce

_No response_

### Version of K8up

v2.7.2

### Version of Kubernetes

v1.27.13

OpenShift 4

vshntea:
## Summary

This change will report metrics for schedules starting at 0. Previously metrics started with value 1 only after a schedule has triggered.

Fixes https://github.com/k8up-io/k8up/issues/976

### DIFF
--- a/operator/monitoring/prometheus.go
+++ b/operator/monitoring/prometheus.go
@@ -35,6 +35,12 @@ var (
 	})
 )
 
+func CreateLabel(namespace string, jobType v1.JobType) {
+	metricsFailureCounter.WithLabelValues(namespace, jobType.String())
+	metricsSuccessCounter.WithLabelValues(namespace, jobType.String())
+	metricsTotalCounter.WithLabelValues(namespace, jobType.String())
+}
+
 func IncFailureCounters(namespace string, jobType v1.JobType) {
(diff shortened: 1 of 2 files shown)


vshntea:
Tested on homelab:
image

Values jump back to zero after restart of operator

Kidswiss (approved):
@vshntea Thanks for the contribution!

Issue #1210, opened 2026-05-29
Improve experience for missing restore method


BacLuc:
### Summary

**As** user\
**I want** a descriptive error message when i forget to specify the restore method\
**So that** i can find my mistake quickly

Currently, the error message is:
```
2026-04-27T09:32:05Z    ERROR    k8up.operator    Reconciler error    {"controller": "restore.k8up.io", "controllerGroup": "k8up.io", "controllerKind": "Restore", "Restore": {"name":"restore-
test","namespace":"vshn-grafana"}, "namespace": "vshn-grafana", "name": "restore-test", "reconcileID": "ddccfbc2-8a88-4bb1-b1c7-6a38c3392f9e", "error": "panic: runtime error: invalid memory a
ddress or nil pointer dereference [recovered]"}
```

in the controller logs.

Ferrari solution: Admission hook does not permit to apply the invalid manifest.

### Context

 I had to ask a k8up expert to find the bug.

PR #1211, opened 2026-05-29
docs/restore.adoc: add restoreMethod to first restore example

BacLuc:
## Summary

* A doc change to increase the chances of the first copy paste for a restore to succeed
* I copied the first example and then failed miserably. See https://github.com/k8up-io/k8up/issues/1210


Kidswiss:
Thanks @BacLuc

Kidswiss (approved):

PR #1213, opened 2026-06-10
Restrict job history cleanup to siblings of the same owner

### LINKED ISSUE #1212: labelSelectors are not respected at cleanup time and cause job history to disappear
### Description

When using labelSelectors for backups, it's possible for history to get truncated and successful backups to not show up / remain hidden via `kubectl get backup`.

That can make a user think the backup did not happen (though it did and can be seen via `kubectl get snapshot`)

More dangerously, the same might apply to failed backups, where a failed one can be hidden from the user.

### Additional Context

When running multiple schedules with [labelSelectors](https://docs.k8up.io/k8up/2.15/references/api-reference.html#k8s-api-github-com-k8up-io-k8up-v2-api-v1-backupschedule) , the backups will be successfully taken.

However, when `successfulJobsHistoryLimit` is cleaned up via `CleanupOldResources` in `operator/executor/generic.go` , it doesn't consider labelSelectors and can cause some schedules to show no recent successful run.

For a user, this presents as if the backup never ran (though it can be verified it did via `kubectl get snapshots`
It needs fixing up, so that proper alerting on last successful backup time (eg. via KubeStateMetrics) can be done.

### Logs

```shell

```

### Expected Behavior

I expect backup resources to show up.

### Steps To Reproduce

_No response_

### Version of K8up

4.9.0

### Version of Kubernetes

v1.33.5

kubeadm

danielpodwysocki:
This adds a regression test and fixes #1212 

TLDR; Multiple Schedules creating jobs of the same kind in one namespace shared a single cleanup pool, so a busy Schedule could evict every recent job of a less busy one, hiding successful (or failed) backups from users.

Some more verbose notes (LLM-derived, so a bit wordy):

Scheduled jobs now get a controller reference (Controller=true) and history-limit cleanup only considers jobs sharing the same owner UID. Jobs created by earlier k8up versions carry the Schedule owner ref without the Controller flag; controllerUID falls back to it so they are still cleaned up after an upgrade. One-off jobs without any owner keep grouping together, as before.

Also dropped the redundant namespace parameter from CleanupOldResources: every caller passed the running job's own namespace, and since sibling matching via owner references is namespace-local anyway, any other  value would be a bug — so it's now derived from the job itself.

### DIFF
--- a/operator/archivecontroller/executor.go
+++ b/operator/archivecontroller/executor.go
@@ -128,7 +128,7 @@ func (a *ArchiveExecutor) setupEnvVars(ctx context.Context, archive *k8upv1.Arch
 }
 
 func (a *ArchiveExecutor) cleanupOldArchives(ctx context.Context, archive *k8upv1.Archive) {
-	a.CleanupOldResources(ctx, &k8upv1.ArchiveList{}, archive.Namespace, archive)
+	a.CleanupOldResources(ctx, &k8upv1.ArchiveList{}, archive)
 }
 
 func (a *ArchiveExecutor) attachTLSVolumeMounts() []corev1.VolumeMount {
--- a/operator/backupcontroller/executor.go
+++ b/operator/backupcontroller/executor.go
@@ -396,7 +396,7 @@ func (b *BackupExecutor) createJob(name, node string, tolerations []corev1.Toler
 }
 
 func (b *BackupExecutor) cleanupOldBackups(ctx context.Context) {
-	b.CleanupOldResources(ctx, &k8upv1.BackupList{}, b.backup.Namespace, b.backup)
+	b.CleanupOldResources(ctx, &k8upv1.BackupList{}, b.backup)
 }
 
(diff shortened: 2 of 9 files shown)


danielpodwysocki:
lint /govulncheck are failing - govulncheck flags no findings on this particular change, however lint doesn't start at all.

It too has unrelated findings when I hit it locally, so we'd probably best 1) merge, 2) fix that.

Let me know if it'd be helpful if I submit another PR after this to fix those up. Otherwise it's a bit confusing  when submitting a PR and trying to gauge if my change is sane :sweat_smile:

danielpodwysocki-gini:
@bastjan @lieneluksika Does the overall direction look good here? Should I fix up those unrelated lint/vulncheck findings or just leave it as-is?

Kidswiss (approved):
Hi @danielpodwysocki 

Thanks for the contribution! LGTM.

I agree the failing checks seem unrelated. They should get fixed as soon as I do a round of renovate PR merging.

PR #1224, opened 2026-06-28
allow annotations on PreBackupPod metadata

AruneshDwivedi:
The PreBackupPod CRD schema rejected metadata.annotations with a strict decoding error because the metadata field was defined as type: object without x-kubernetes-preserve-unknown-fields. Adding it allows users to set annotations (e.g. Vault agent sidecar injection) on prebackup pods.


AruneshDwivedi:
Closing this PR — the fix requires modifying the Go struct with a kubebuilder marker and regenerating CRDs via controller-gen, which needs a proper build environment. The chart-lint CI regenerates CRDs from Go source, so manual YAML edits alone are insufficient. Will revisit when able to run `make generate` locally.

PR #1234, opened 2026-07-16
Fix potentially duplicate volumes and mounts

Kidswiss:
## Summary

#1223 introduced a potential bug where duplicate volumes and volumeMounts can be defined.

This commit adds functions to deduplicate the volumes and volumeMounts.

The e2e tests caught this issue:
```
make e2e-test -e KIND_KUBECTL_ARGS=--validate=false -e bats_args="--report-formatter junit" -e BATS_FILES="test-03-backup.bats"
```

For some reason I did not notice the broken check on the PR.

CC: @beniwtv 

### For Code changes

### DIFF
--- a/operator/archivecontroller/executor.go
+++ b/operator/archivecontroller/executor.go
@@ -55,8 +55,8 @@ func (a *ArchiveExecutor) Execute(ctx context.Context) error {
 
 		batchJob.Spec.Template.Spec.Containers[0].Env = append(batchJob.Spec.Template.Spec.Containers[0].Env, a.setupEnvVars(ctx, a.archive)...)
 		a.archive.Spec.AppendEnvFromToContainer(&batchJob.Spec.Template.Spec.Containers[0])
-		batchJob.Spec.Template.Spec.Containers[0].VolumeMounts = append(batchJob.Spec.Template.Spec.Containers[0].VolumeMounts, a.attachTLSVolumeMounts()...)
-		batchJob.Spec.Template.Spec.Volumes = append(batchJob.Spec.Template.Spec.Volumes, utils.AttachEmptyDirVolumes(a.archive.Spec.Volumes)...)
+		batchJob.Spec.Template.Spec.Containers[0].VolumeMounts = utils.AppendUniqueVolumeMounts(batchJob.Spec.Template.Spec.Containers[0].VolumeMounts, a.attachTLSVolumeMounts()...)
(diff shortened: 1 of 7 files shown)


operator/archivecontroller/executor.go
```diff
@@ -55,8 +55,8 @@ func (a *ArchiveExecutor) Execute(ctx context.Context) error {
 
 		batchJob.Spec.Template.Spec.Containers[0].Env = append(batchJob.Spec.Template.Spec.Containers[0].Env, a.setupEnvVars(ctx, a.archive)...)
 		a.archive.Spec.AppendEnvFromToContainer(&batchJob.Spec.Template.Spec.Containers[0])
-		batchJob.Spec.Template.Spec.Containers[0].VolumeMounts = append(batchJob.Spec.Template.Spec.Containers[0].VolumeMounts, a.attachTLSVolumeMounts()...)
-		batchJob.Spec.Template.Spec.Volumes = append(batchJob.Spec.Template.Spec.Volumes, utils.AttachEmptyDirVolumes(a.archive.Spec.Volumes)...)
+		batchJob.Spec.Template.Spec.Containers[0].VolumeMounts = utils.AppendUniqueVolumeMounts(batchJob.Spec.Template.Spec.Containers[0].VolumeMounts, a.attachTLSVolumeMounts()...)
+		batchJob.Spec.Template.Spec.Volumes = utils.AppendUniqueVolumes(batchJob.Spec.Template.Spec.Volumes, utils.AttachEmptyDirVolumes(a.archive.Spec.Volumes)...)
```
vshntea, line 59: Why are the volumes being handled explicitly here and not inside `job.MutateBatchJob`

vshntea (approved):

operator/utils/utils.go
```diff
@@ -18,6 +18,46 @@ const (
...
+				break
+			}
+		}
+
+		if !taken {
+			existing = append(existing, volume)
+		}
+	}
+	return existing
+}
```
vshntea, line 59: I am not yet a go developer, but this looks like a lot of duplication. Could this be simplified by using some sort of comparator and a generic?

operator/archivecontroller/executor.go
Kidswiss: Because `mutateBatchJob` handles job setup that is common for every job type.

The volumes a job needs depends on the type. For example, a backup needs other volumes than a restore or a prune.

operator/utils/utils.go
Kidswiss: That was also my first thought, however, after looking at the final generic implementation I was not quite happy, since it was almost unreadable.

I rather opt for a bit of duplicated, easily maintainable code, than shoehorning it into a DRY pattern.

Kidswiss:
@vshntea thanks for the review! 🎉

PR #1242, opened 2026-07-31
Set github action digests

Kidswiss:
## Summary

This commit will make renovate use digests for github actions instead of version numbers.

This slightly increases the security, since github actions don't provide locks for their versions. So a malicious actor could simply switch out a known good version of an action with a compromised one and github would gladly run it.

### For Code changes

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -14,23 +14,40 @@
   "packageRules": [
     {
       "description": "Label chart dependency updates",
-      "matchFileNames": ["charts/**"],
-      "addLabels": ["area:chart", "chart:k8up"]
+      "matchFileNames": [
+        "charts/**"
+      ],
+      "addLabels": [
+        "area:chart",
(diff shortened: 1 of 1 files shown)


bastjan (commented):
Change is a great idea I think. I would add a cooldown period tho. See for example:

https://github.com/helmetica-framework/chrysopoeia/blob/306bf3a12969ac9ba5a85ff7a89522e8e7f590c0/renovate.json#L26-L32

I'd prefer to have the format change in a separate PR or at least commit, this is way harder to review than it should be.

Kidswiss:
> I'd prefer to have the format change in a separate PR or at least commit, this is way harder to review than it should be.

Curse of autoformat on save... I'll split it into another commit.

bastjan (approved):

Kidswiss:
Thanks for the review!

Issue #828, opened 2023-03-21
Allow the `spec.backend.s3.endpoint` and `spec.backend.s3.bucket` values to be taken from secrets


akosma:
### Summary

**As** an AppCat and K8up user
**I want** to be able to specify the S3 endpoint and bucket name from secrets
**So that** I can reuse the information provided by AppCat when provisioning S3 buckets.

### Context

Right now the `spec.backend.s3.endpoint` and `spec.backend.s3.bucket` values must be strings. This feature would allow users to specify them as objects, getting their values from pre-existing secrets, similarly to how one can specify the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` values.

The YAML below (invalid at the time of this writing) describes how this feature would allow grabbing the S3 endpoint and bucket name values from pre-existing secrets (in this case named `objectbucket-creds`):

```
apiVersion: k8up.io/v1
kind: Backup
metadata:
  name: backup-test
spec:
  backend:
    repoPasswordSecretRef:
      name: restic-repo
      key: password
    s3:
      endpoint:
        valueFrom:
          secretKeyRef:
            name: objectbucket-creds
            key: ENDPOINT_URL
      bucket:
        valueFrom:
          secretKeyRef:
            name: objectbucket-creds
            key: BUCKET_NAME
      accessKeyIDSecretRef:
        name: objectbucket-creds
        key: AWS_ACCESS_KEY_ID
      secretAccessKeySecretRef:
        name: objectbucket-creds
        key: AWS_SECRET_ACCESS_KEY
```

### Further links

AppCat: https://docs.appuio.cloud/appcat/index.html

gunishmatta:
@akosma  I would like to work on this

Kidswiss:
@gunishmatta Have you started working on this?

We have this in our backlog, so I'd like to avoid having two implementations for this :)

akosma:
Any news about this one?

M0NsTeRRR:
Hello, it would be cool if this feature is implemented for other backends too.

PR #1249, opened 2026-08-24
Introduce secretRefs for endpoint and bucket

vshntea:
Introduce `endpointSecretRef` and `bucketSecretRef` which allow setting the endpoint and bucket via secret. It would have been possible to change the `endpoint` and `bucket` into an object that takes in either a string or a secretRef, however that wouldn't be backwards compatible and introduce a breaking change.

Issue: GH-828


vshntea:
Hm, the String Operation is also used for locking mechanisms, to make sure there are no concurrent pods working on the same repository. With everything being moved to secrets, the endpoint can now effectively be the hash of "s3:$(S3_ENDPOINT_SECRET)/$(S3_BUCKET_SECRET)" which will prevent other pods from being created, even if on different repositories...

vshntea:
> Hm, the String Operation is also used for locking mechanisms, to make sure there are no concurrent pods working on the same repository. With everything being moved to secrets, the endpoint can now effectively be the hash of "s3:$(S3_ENDPOINT_SECRET)/$(S3_BUCKET_SECRET)" which will prevent other pods with different endpoints/buckets from being created, even if on different repositories...

I don't necessarily want the operator to read the secrets to resolve this, I think it would make things overly complicated. Using the secret keys/names also won't really help as it would just push the collision to a different domain. I am thinking of introducing an identity string, which can be configured, to distinguish between different repositories. When the identity string is given, it will be used for hashing instead, so this is still backwards compatible.

vshntea:
As discussed, we will configure the k8up operator to resolve the secrets. This will work using RBAC with resourcenames so the operator can target only a specific secret per namespace.

The secretname will be called `k8up-repo` by default. This will be configurable via the operator and changes would have to be done on the rbac accordingly. This change is backwards-compatible and ideally doesn't touch any logic we currently have but instead resolves the secret on object creation.


> I am thinking of introducing an identity string, which can be configured, to distinguish between different repositories. When the identity string is given, it will be used for hashing instead, so this is still backwards compatible.

This change could corrupt the repo if the identities were set wrong on tasks that resolve at the repository. While restic does use locking, apparently it is not perfect.

api/v1/backend.go
```diff
@@ -71,6 +73,14 @@ func (in *Backend) String() string {
...
 }
 
+// ResolveSecretRefs resolves Secret-backed repository fields in place.
+func (in *Backend) ResolveSecretRefs(ctx context.Context, reader client.Reader, namespace string) error {
+	if in == nil || in.S3 == nil {
+		return nil
+	}
+	return in.S3.ResolveSecretRefs(ctx, reader, namespace)
+}
+
```
vshntea, line 83: Fairly empty, but would allow for future expansion and keeps the usage of this object agnostic to what filesystem is used in the background

charts/k8up/templates/operator-clusterrole.yaml
```diff
@@ -24,6 +24,14 @@ rules:
...
       - list
       - watch
+  - apiGroups:
+      - ""
+    resourceNames:
+      - k8up-repo
+    resources:
+      - secrets
+    verbs:
+      - get
```
vshntea, line 34: Should be configurable in a future commit of this branch

operator/archivecontroller/controller.go
```diff
@@ -29,6 +32,17 @@ func (r *ArchiveReconciler) NewObjectList() *k8upv1.ArchiveList {
 func (r *ArchiveReconciler) Provision(ctx context.Context, obj *k8upv1.Archive) (controllerruntime.Result, error) {
 	log := controllerruntime.LoggerFrom(ctx)
 
+	if obj.Spec.Backend != nil {
+		if err := obj.Spec.Backend.ResolveSecretRefs(ctx, r.KubeReader, obj.GetNamespace()); err != nil {
+			return controllerruntime.Result{}, fmt.Errorf("resolve backend Secret references: %w", err)
+		}
+	}
+	if obj.Spec.RestoreMethod != nil {
+		if err := obj.Spec.RestoreMethod.S3.ResolveSecretRefs(ctx, r.KubeReader, obj.GetNamespace()); err != nil {
```
vshntea: I don't like that I have to expose S3 in this part of the code, however given that this is the only occurrence, I also wouldn't want to implent `ResolveSecretRefs` for the RestoreMethod, it mostly looks like an API. This is some ugly thing we probably have to live with.

api/v1/backend.go
Kidswiss: Maybe it makes sense to already add the other backends as well though? While we use exclusively S3, there are other users who might want to use other backends.

Kidswiss (changes requested):
Can you bump the go version in `go.mod`? That should clean the govuln error.

